PR #213, opened 2013-03-22
Added new python Redis client: RedisStore

timothycrosley:
Include RedisStore python client.


seamusabshere:
@timothycrosley is this library still maintained?

PR #241, opened 2013-07-15
add exredis client

artemeff:
First client for the Elixir language, wrapper for [eredis](https://github.com/wooga/eredis)


artemeff:
#496

PR #343, opened 2014-04-10
Update clients.json

chiradip:
Added - A no nonsense Redis Client using pure scala.


chiradip:
What is this merge conflict?

chiradip:
What is this merge conflict?

antirez:
No problem with the conflict, we verify a bit entries before adding, @soveran tagged this `client` so I guess it is already in the processing queue.

PR #344, opened 2014-04-10
Add clarity to "partial synchronization"

### LINKED ISSUE #341: Add clarity to "partial synchronization"
We now had several users in IRC asking why Partial resynchronization failed with "Partial resynchronization not possible (no cached master)" when the slave was restarted.

The section in the documentation could be improved: https://github.com/antirez/redis-doc/blob/master/topics/replication.md#partial-resynchronization

I will add one or two sentences later (I don't have the time right now)

badboy:
Fixes #341


badboy:
@antirez quick ack please?

PR #353, opened 2014-04-26
Propose adding Pypredis client

pepijndevos:


pepijndevos:
Note that I developed PypRedis for my job at Freenom, where I no longer work. While it's used in production there, I'm not really working on it anymore.

badboy:
Marked as inactive in 11bb96b1f4aba4c1f84b0d7f872b7572fb9eb2e4.

PR #440, opened 2014-10-21
Add Haskell client `redis-io`.

twittner:


soveran:
Can you resubmit? Don't add empty new lines (I will do that in one pass after merging all the other pull requests) and rebase with current master.

PR #472, opened 2014-12-24
Add OCaml client

aluuu:


soveran:
Please change the patch so that it doesn't add empty newlines. In the future I may change that, but for now I'd like to keep merging other clients with the current format.

aluuu:
Ok. There's commit without newline changes.

soveran:
Thanks!

PR #507, opened 2015-03-13
hearthbeats → heartbeats

badboy:


antirez:
Thanks @badboy, sorry I was told than when in doubt it was better to add a few `h`s.

badboy:
:D

PR #509, opened 2015-03-15
Fix distribution for allkeys-random policy in lru-cache.md

xkrt:
As far as I can see in the description of the allkeys-random policy we are talking about [Uniform distribution](http://en.wikipedia.org/wiki/Uniform_distribution_(continuous)), not about [Normal distribution](https://en.wikipedia.org/wiki/Normal_distribution)


soveran:
@xkrt Thanks!

PR #515, opened 2015-03-23
Fix parameter name in CONFIG SET

jbochi:


badboy:
Indeed, that makes it a bit more clear. Will merge in a second.

PR #530, opened 2015-04-09
Setslot fixes

badboy:
1. Just a typo, it should be the destination id as in the docu above
2. Isn't the correct and better way to use `SETSLOT id NODE node-id` on one node and let it propagate? `redis-trib.rb` does the same.


badboy:
@antirez ping

antirez:
Hello @badboy, you are correct in both the instances. Thanks.

PR #538, opened 2015-04-24
Typo fix

miloprice:
The intro to datatypes makes a reference to `BITOPS`, which should be `BITPOS`. Note that `BITOPS` could also plausibly be a misspelling of `BITOP`, except that `BITOP` can't operate on ranges.


badboy:
Thanks, indeed a typo.

PR #546, opened 2015-05-11
Fix typo in cmsgpack eval example

romuald:
There was a missing trailing `"` in one of the cmsgpack EVAL examples. Simply added it


charsyam:
@romuald Thanks.

PR #552, opened 2015-05-27
Clean up some english

mperham:
Some editing and corrections from a native speaker.  I didn't edit the entire page but just a few quirks I saw.


badboy:
Thanks!

PR #558, opened 2015-06-11
Correct redis-cli pipe mode : available at 2.6 or higher

HeartSaVioR:
Based on https://github.com/antirez/redis/commit/088c508abc89b66ce14870a2abe06879238a3b21, we don't need to checkout unstable branch and build by hand to use pipe mode.


badboy:
Pipe mode is available since 2.6

HeartSaVioR:
@badboy Oh. OK. I'll correct it.

badboy:
Merged as a single commit: https://github.com/antirez/redis-doc/commit/def67b170d2f43875e0da31b347358b5eb429078

PR #562, opened 2015-06-16
"Escaped" KEYS + formatted KEYS

itamarhaber:
Per @badboy_'s recommendation


badboy:
Should be \`!KEYS\`

seppo0010:
It should probably be \`!KEYS\` both times? :P

badboy:
It should. But I can squash + fix that.

itamarhaber:
Feeling too stupid right now to come up with a snappy response - I can't even fix a doc without borking  the compilation.

badboy:
Nope, "compilation" is fine. And no problem, it's now fixed, so all good.

PR #577, opened 2015-07-10
Use Make for spell checking

djanowski:


makefile
```diff
@@ -1,4 +1,33 @@
...
+spell: tmp/commands tmp/topics $(SPELL_FILES)
+	find tmp -name '*.spell' | xargs cat > tmp/all.spell
+	cat tmp/all.spell
+	test -s tmp/all.spell && exit 1
+
+$(TEXT_FILES): tmp/%.txt: %.md
+	./bin/text $< > $@
+
+$(SPELL_FILES): %.spell: %.txt tmp/dict
+	aspell -a --extra-dicts=./tmp/dict 2>/dev/null < $< | \
```
badboy: We should pass `--lang=en` here. Not everyone has english as the default, and this should work on user machines as well.

makefile
```diff
@@ -1,4 +1,33 @@
...
-	ruby -rjson -r./utils/clients -e 'Clients.check(JSON.parse(File.read("clients.json"), symbolize_names: true))'
+MD_FILES:=$(shell find {commands,topics} -name '*.md')
+TEXT_FILES:=$(patsubst %.md,tmp/%.txt,$(MD_FILES))
+SPELL_FILES:=$(patsubst %.txt,%.spell,$(TEXT_FILES))
 
-.PHONY:
+spell: tmp/commands tmp/topics $(SPELL_FILES)
+	find tmp -name '*.spell' | xargs cat > tmp/all.spell
+	cat tmp/all.spell
+	test -s tmp/all.spell && exit 1
```
badboy: Should be

```
test -s tmp/all.spell || exit 0
```

as `test -s` will fail with an empty file and this is what we want (empty file = no corrections)

makefile
badboy: Never mind, we're creating the extra dict as english, that should do the trick.

badboy:
I got an updated wordlist for this already. :)

makefile
badboy: Sorry, we need both to invert `test -s`'s return value.

badboy:
I think I got it now, plus a whole lot more words that needs to be added. https://travis-ci.org/badboy/redis-doc/builds/70417449

(Not sure why they weren't found before…)

PR #578, opened 2015-07-15
grammar fixes

breyed:


badboy:
Thanks, I added another commit fixing two smaller things as well.

PR #583, opened 2015-08-01
Further PHP implementation

malkusch:
The current link to the PHP implementation has some limitations, i.e. there's no release, it doesn't support both Redis-APIs. So therefore I [implemented the Redlock algorithm](https://github.com/malkusch/lock#phpredismutex) as well and released it. I would be happy to see that reference in the documentation.

Also I'd like to use this opportunity to ask why the algorithm proposes to release also locks which weren't acquired.


seamusabshere:
@malkusch i think you should add https://github.com/malkusch/lock to tools.json too

PR #584, opened 2015-08-02
Add note about multi/exec in brpoplpush

seppo0010:


seppo0010:
This [undocumented?] behavior surprised me. https://github.com/antirez/redis/blob/unstable/src/t_list.c#L940

badboy:
Indeed. This should be in there.

PR #586, opened 2015-08-04
fix lru-cache.md: allkeys-lru is not suitable for persistent situation

swordflychen:
**allkeys-lru** is not suitable for both caching and to have a set of persistent keys in one instance. But **volatile-lru** is very suitable for the situation. 
 
This is just my point of view, please check it. Very thanks.


swordflychen:
fix lru-cache.md: allkeys-lru is not suitable for persistent situation

PR #592, opened 2015-08-17
Documents the cluster's READONLY and READWRITE commands

itamarhaber:


itamarhaber:
Doesn't look like this PR broke it @djanowski

djanowski:
There's a syntax error in the JSON file (highlighted by GitHub in the diff).

itamarhaber:
Fixes #596

charsyam:
@itamarhaber Thanks. but I have a question.
is it right for summary of read write?
I think it is just turn off read only flags.
but it looks like disable read features.

itamarhaber:
@charsyam `READWRITE` prevents reads from a cluster slave node (the default), so yeah, you could say it kind of disables reads

itamarhaber:
:scream: @badboy is needed in aisle one!

commands/client-list.md
```diff
@@ -40,7 +40,8 @@ b: the client is waiting in a blocking operation
 i: the client is waiting for a VM I/O (deprecated)
 d: a watched keys has been modified - EXEC will fail
 c: connection to be closed after writing entire reply
-u: the client is unblocked
+U: the client is unblocked
```
badboy: `u` = `This client was unblocked`

`U` = `Client connected via Unix domain socket`

itamarhaber: Redis haz socketz? WOW! :ok_hand:

PR #597, opened 2015-08-27
Make the number consistent with command output

randvis:
The output of command shows  **114** lost but the description is **144**.


randvis:
fix #598

PR #603, opened 2015-08-28
Add GEO's six commands into "redis-client for C++"

zhengshuxin:
The "redis-client for C++" of acl has supported GEO's six commands including "geoadd, geohash, geopos, geodist, georadius, georadiusbymember" functions in class redis_geo.


badboy:
Why not just removing the list then if it includes everything anyway?

zhengshuxin:
Other than the hiredis, the "redis-client for C++" of acl provide one or more function in one redis client class for each redis command, which let the users using the redis client more easily and safely.  Even if the redis client provides functions to access redis-server directly by protocol level, but it's too difficult and easy to make wrong just same as using hiredis.

PR #605, opened 2015-08-28
PFMERGE includes target HLL in the merge

avram:
This surprised me and wasn't a documented behavior, but it is 100% reproducible. The docs state that the source sets are merged into a new HyperLogLog, which set to the destination key; they do not state that the destination key is party to the merge.

```
127.0.0.1:6379> PFADD test 1
(integer) 1
127.0.0.1:6379> PFADD test 2
(integer) 1
127.0.0.1:6379> PFADD test 3
(integer) 1
127.0.0.1:6379> PFCOUNT test
(integer) 3
127.0.0.1:6379> PFADD foo 1
(integer) 1
127.0.0.1:6379> PFADD foo 2
(integer) 1
127.0.0.1:6379> PFCOUNT foo
(integer) 2
127.0.0.1:6379> PFMERGE targ foo test
OK
127.0.0.1:6379> PFCOUNT targ
(integer) 3
127.0.0.1:6379> PFADD bar 6
(integer) 1
127.0.0.1:6379> PFMERGE targ foo bar
OK
127.0.0.1:6379> PFCOUNT targ
(integer) 4
```

### DIFF
--- a/commands/pfmerge.md
+++ b/commands/pfmerge.md
@@ -5,6 +5,10 @@ structures.
 The computed merged HyperLogLog is set to the destination variable, which is
 created if does not exist (defaulting to an empty HyperLogLog).
 
+If the destination variable exists, it is treated as one of the source sets 
+and its cardinality will be included in the cardinality of the computed
+HyperLogLog.
+
 @return
 
 @simple-string-reply: The command just returns `OK`.


avram:
Or, to use the example embedded in the docs:

```
redis> PFADD hll1 foo bar zap a
(integer) 1
redis> PFADD hll2 a b c foo
(integer) 1
redis> PFMERGE hll3 hll1 hll2
OK
redis> PFCOUNT hll3
(integer) 6
redis> PFADD hll1 boom
(integer) 1
redis> PFCOUNT hll1
(integer) 5
redis> PFMERGE hll3 hll1
OK
redis> PFCOUNT hll3
(integer) 7
redis>
```

badboy:
Sounds more like a bug in the code than missing documentation.
@antirez, any input on this?

avram:
I was more willing to believe that this was a bug in the docs, but I can only conclusively say that the documentation and behavior don't align. One probably should be made to match the other!

antirez:
Hello, I just checked the code, the `PFMERGE` implementation looks definitely implemented to have the current behavior. Currently I don't recall exactly why I did this choice, but I can re-reason in order to understand why I picked such a design.

One interesting thing is that the other behavior is always easy to obtain with `DEL dstkey` followed by PFMERGE, possibly inside a transaction if `dstkey` is itself target of `PFADD` operations: but then you would lose the `PFADD` targeting `dstkey`, so likely `dstkey` is not targeted by `PFADD` operations if you want this behavior.

On the other hand, the reverse is also true, if the new HLL is created from scratch, anyway to mimic the current behavior is as simple as repeating the `dstkey` two times in the argument list: `PFADD dstkey dstkey otherkey`.

So anyway the two forms are kinda equivalent apparently.

Now to be honest, the current form is a bit unexpected... I think it is surprising since no Redis command behaves like that, new keys are created from scratch when they are target of an operation that is going to store some value: think at Set operations for example.

So, if we don't break code, I would change the implementation and take the doc as it is, however like @avram discovered this behavior, other users may already have discovered it and used it, without reporting it... At this point changing the implementation looks a bit risky.

Ok that's the story, do you have any hint?

avram:
Given that this has been the case for years, it seems like the right thing to do is adjust the docs accordingly, merging this PR.

My team's code definitely just has the DEL destkey in it and has ever since we ran into this.

itamarhaber:
Alrighty, enough time has passed to establish that this is the behavior and it should be documented. If it will be changed in a future version we'll adjust the docs. Merging and thank you @avram!

PR #608, opened 2015-09-07
simplify dependency installation, esp. ruby gems

seamusabshere:
using a `.gems` file is antiquated, use industry-standard `bundler` instead

also clarify `apt-get` and `brew` usage

### DIFF
--- a/.gems
+++ b/.gems
@@ -1,2 +1,0 @@
-batch -v 1.0.4
-redcarpet -v 3.3.2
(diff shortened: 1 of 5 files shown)


badboy:
> using a .gems file is antiquated, use industry-standard bundler instead

No, it's not. People are happily using [`dep`](https://github.com/cyx/dep) just fine in production. And `redis.io` runs with it as well. The maintainers of redis.io and the docs try to keep it down to the necessary tools. No need for the heavy-weight bundler here.

We should update the readme and the `.gems` file though.

PR #649, opened 2015-12-22
Fix active/passive terminology in EXPIRE docs, fixes #491

### LINKED ISSUE #491: Incorrect usage of active/passive terminology
From http://redis.io/topics/latency: 

> The active expiring is designed to be adaptive. An expire cycle is started every 100 milliseconds (10 times per second), and will do the following:

From http://redis.io/commands/expire:

> A key is actively expired simply when some client tries to access it, and the key is found to be timed out.

rohitpaulk:
https://github.com/antirez/redis-doc/issues/491

### DIFF
--- a/commands/expire.md
+++ b/commands/expire.md
@@ -125,7 +125,7 @@ lasting for 1000 seconds.
 
 Redis keys are expired in two ways: a passive way, and an active way.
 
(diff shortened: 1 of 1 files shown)


rohitpaulk:
... ping @antirez

PR #661, opened 2016-01-12
Adds details about NOKEY reply

itamarhaber:


badboy:
Can you wrap NOKEY in backticks? Then it will be ignored by the spellchecker (and will be properly highlighted in the rendered documentation)

itamarhaber:
@badboy Sure, done.

PR #669, opened 2016-01-31
Update clients.json

Ke-:
Added Lasso Redis client.


badboy:

Ke-:
Thanks, have removed the Github handle.

badboy:
:+1:

PR #672, opened 2016-02-04
add a c++ redis client

shawn246:
A c++ redis client supports cluster/pipeline, and will support transaction soon.


badboy:
Right now we link the author to a Twitter profile, therefore it should be your valid Twitter handle. It looks like `@shawn` isn't yours. Would be good to correct it in this PR then. If you don't have one, don't worry, just remove the author for now.

shawn246:
sorry for that, and i have removed the author.

PR #677, opened 2016-02-07
Update client.json with go-redis/redis client

eatonphil:
I'm surprised this package isn't already in the list, but I didn't see it. It is fairly popular.

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -92,8 +92,8 @@
   {
     "name": "go-redis/redis",
     "language": "Go",
-    "repository": "https://github.com/go-redis/redis"
-    "description": "Redis client for Golang.",
+      "repository": "https://github.com/go-redis/redis",
+    "description": "Redis client for Golang supporting Redis Sentinel and Redis Cluster out of the box.",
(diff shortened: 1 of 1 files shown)


cheprasov:
I think, it will be better to ask authors of client to add client here by theyself, maybe, they have some reasons not to do it yet.

badboy:
/cc @vmihailenco

@eatonphil: JSON is broken

clients.json
```diff
@@ -90,6 +90,15 @@
   },
 
   {
+    "name": "go-redis/redis",
+    "language": "Go",
+    "repository": "https://github.com/go-redis/redis"
+    "description": "Redis client for Golang.",
```
vmihailenco: Probably mention that it supports Redis Sentinel and Redis Cluster out of the box.

vmihailenco:
Thanks for adding client to the list! I don't have any objections :)

eatonphil:
@badboy thanks, made those changes. @vmihailenco updated to include that info!

PR #679, opened 2016-02-12
Fix hash encoding in object encoding command

misterion:
Actually Redis do not use zipmap. Ziplist used as small hash encoding. This pull request fix this in `object ecnding` command description.


badboy:
We used it once, we don't use it anymore. Correct.

PR #682, opened 2016-02-14
Remove flag "active" to some inactive clients

olivierlemasle:
Definition of "active" is taken from http://redis.io/clients
(activity on the official repository within the latest 6 months)


cheprasov:
@badboy 
I think, it is a good idea. Because, for example, last release of Rediska client  is 12 Aug 2014

badboy:
Well, as per definition of our "active" flag it's properly correct. I don't like this flag anyway, some clients simply don't need to be modified because they are pretty much feature-complete (and bug-free to some extent).

I trust you on this that the data is correct for now.

PR #686, opened 2016-02-17
Remove incorrect sentinel symbol in "Example 3: Sentinel in the client boxes"

randvis:
I'm a bit confused about the graph in "Example 3: Sentinel in the client boxes" section in sentinel.md doc. 

In my opinion, in this example, there're three sentinels and they're all in client boxes. S1 and S2 aren't supposed to stay with either master or slave. 

Could you please take a look at this? Thank you very much.


jlgreen:
That was my thinking too @randvis. Any chance this could be merged to avoid further confusion (assuming it's not intended for the sentinels to be on the master/ slave boxes)?

antirez:
Was an error indeed. Thanks for fixing.

PR #691, opened 2016-03-01
Added sort param for GEO commands

cheprasov:


cheprasov:
@badboy Let's do it )

PR #694, opened 2016-03-08
Change Swift-Redis client name to ZRedis

rabc:
I had to change the library name due to Swift name conventions.

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -1239,9 +1239,9 @@
   },
 
  {
-    "name": "Swift-Redis",
+    "name": "ZRedis",
     "language": "Swift",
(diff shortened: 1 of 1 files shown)


badboy:
What a swift change.

rabc:
@badboy sorry for bothering again! We had a discussion in our Slack about the libraries name, and we think it's better to stick just with `Redis` name, because it makes more sense with all other libraries we're building in our ecosystem.

Can you just change it, so I don't have to bother with other PR, please?
Name: Redis
Repository: https://github.com/rabc/swift-redis

Thanks again!

PR #695, opened 2016-03-09
Fixed #693: Confusing use of term "pure function"

### LINKED ISSUE #693: Confusing use of the term "pure function"
The [`EVAL`](https://github.com/antirez/redis-doc/blob/dab803f3842a8b9cafde4a2fbccc7b79ebe67c17/commands/eval.md) scripting documentation refers to scripts as *pure functions* in a few places (e.g. "A very important part of scripting is writing scripts that are pure functions").

However, the term "pure function" has a specific meaning in programming (see, for example, [Wikipedia](https://en.wikipedia.org/wiki/Pure_function)): they are functions whose return value is entirely determined by their arguments, and which don't have side effects.

Redis scripts are not that. They do have side effects (modifications to the database), and the return value can be random as long as the database writes are deterministic.

The rest of the documentation is pretty clear on this point (the key line is "The script must always evaluates [sic] the same Redis write commands with the same arguments given the same input data set"), so I would just avoid using the term *pure function*. "Scripts with deterministic side effects" doesn't have quite the same ring to it as "Scripts as pure functions", perhaps, but it would be more accurate.

marfire:
Also cleaned up and clarified other text in the
replication section of the EVAL command.

### DIFF
--- a/commands/eval.md
+++ b/commands/eval.md
@@ -323,32 +323,37 @@ SCRIPT currently accepts three different commands:
     not violate the scripting engine's guaranteed atomicity).
     See the next sections for more information about long running scripts.
 
-## Scripts as pure functions
+## Scripts with deterministic writes
 
-A very important part of scripting is writing scripts that are pure functions.
+A very important part of scripting is writing scripts that only change the
+database in a deterministic way.
 Scripts executed in a Redis instance are, by default, replicated on slaves
 and into the AOF file by sending the script itself -- not the resulting
 commands.
+Since the script will be re-run on the remote host (or when reloading the AOF
+file), the changes it makes to the database must
+be reproducible.
 
-The reason is that sending a script to another Redis instance is often much
-faster than sending the multiple commands the script generates, so if the
-client is sending many scripts to the master, converting the scripts into
-individual commands for the slave / AOF would result in too much bandwidth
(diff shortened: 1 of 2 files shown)


commands/eval.md
```diff
@@ -323,40 +323,45 @@ SCRIPT currently accepts three different commands:
...
+scripts to the master, converting the scripts into individual commands for the
+slave / AOF would result in too much bandwidth for the replication link or the
+Append Only File (and also too much CPU since dispatching a command received
+via network is a lot more work for Redis compared to dispatching a command
+invoked by Lua scripts).
+
+In some cases, however, it makes more sense to replicate the commands
+themselves than the script. 
+So starting with Redis 3.2 (currently not stable),
+this is an option. See the next section for more information.
```
zuiderkwast, line 349: Redis 3.2 was stable long time ago, I think... so no need to mention this.

zuiderkwast (approved):
The change contains many grammar improvements, which are good AFAICT, but it makes reviewing harder.

There are some references to "Redis 3.2 (not yet stable)" but those are not new, so I don't think it should be blocking the merge of this PR.

zuiderkwast:
Replaced by #1622, just a rebase of this PR.

PR #700, opened 2016-03-31
added noncis description

reidwmulkey:


badboy:
This looks more like a tool than a simple client library. Move it to tools.json please.

PR #701, opened 2016-03-31
C client library add. Eredis.

guillaumef:


badboy:

PR #705, opened 2016-04-08
Edits and such

itamarhaber:
Great addition - sorry for all the formatting changes.

PR #707, opened 2016-04-09
Minor misprint correction

qrilka:


badboy:
Contained in #705, so I will close this and wait for the big merge to happen.

PR #712, opened 2016-04-18
Fix typo topics/benchmarks

quiver:
Before this PR is merged,  antirez/redis#3164 needs to be merged.


badboy:
Merging now, the tool only accepts it with a double-dash, so why wait?

PR #717, opened 2016-04-30
Add Redis Cluster Java Client, Jedipus.

jamespedwards42:
https://github.com/jamespedwards42/jedipus


badboy:

jamespedwards42:
I guess I should start shouting out to the twitterverse... Went ahead and added my unused until now twitter name.

PR #718, opened 2016-05-01
Update clients.json

Qti3e:
New PHP Client


badboy:
Please fix the JSON

Qti3e:
Fixed

cheprasov:
I believe, that your client must have more tests for to be added on the official site as a client.
Because, now client has not any tests.

badboy:
Yup, it's pretty bare bones and incomplete. I'd be happy to add it when it's fully working. :)

PR #719, opened 2016-05-04
Small error in example

h0x91b:
075 and 200
0            2
 7            0
   5            0
027050 not 027005

Also there:
001[00]1011 - 9 bit
011[00]1000 - 9 bit
should be four zeroes and 18 bit result, no? 0001110011[0000]1010


badboy:
The result for the second example should be `000111000011001010`. Can you update that?

h0x91b:
Yep, my bad, should be 000111[0000]11001010

badboy:
:+1:

PR #720, opened 2016-05-06
Fix a minor error in the BITFIELD doc

trekawek:
BITFIELD allows to treat a string as a "array of bits" rather than "bit of arrays".


antirez:
Thanks!

PR #722, opened 2016-05-06
adds redis-bash to clients.json

caquino:
This PR adds redis-bash to the clients.json list as documented at http://redis.io/clients.
Thanks

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -1168,6 +1168,15 @@
   },
 
   {
+    "name": "redis-bash",
+    "language": "Bash",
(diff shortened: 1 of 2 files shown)


caquino:
@antirez if necessary the second commit can be reverted, it was a fix to make travis-ci happy.

badboy:
Yep, please revert the second commit.

Right now we link the author to a Twitter profile, therefore it should be a valid Twitter handle. 
Is [@caquino](https://twitter.com/caquino) your handle?
If you don't have one, don't worry, just remove the author for now.

PR #723, opened 2016-05-07
Fixed STORE and STOREDIST params for GEORADIUS and GEORADIUSBYMEMBER

cheprasov:
@badboy 
Thank you.


cheprasov:
Ooops.. What did I do wrong? Maybe 2 params with the same name? But they is keys. Hm..

antirez:
Thanks merged. You did not anything wrong, it's just that there is a test to disable in the CI since it does not work.

PR #730, opened 2016-06-07
Add Java rdb-parser tool

jwhitbeck:
We've been successfully using this library in production for 6 months and would like to share with the community. I just added support for parsing RDB version 7.

### DIFF
--- a/tools.json
+++ b/tools.json
@@ -574,7 +574,6 @@
     "name": "rdb-parser",
     "language": "Java",
     "repository": "https://github.com/jwhitbeck/java-rdb-parser",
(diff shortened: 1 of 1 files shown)


badboy:
Right now we link the author to a Twitter profile, therefore it should be a valid Twitter handle.
Is [@jwhitbeck](https://twitter.com/jwhitbeck) your handle?
If you don't have one, don't worry, just remove the author for now.

jwhitbeck:
Thanks for looking into this. I don't have a twitter profile, so I removed the author.

PR #728, opened 2016-05-29
Note on RENAME command changes in 3.2

popravich:
Note is missing in docs.
This is related to antirez/redis#2392
Also, as far as I see release notes missing info on this change.

PR #731, opened 2016-06-09
corrected description for case when key == newKey

rssh:
when source and destination are the same, no operations are performed, according to the comment in line 777 of db.c


popravich:
already waiting #728

PR #732, opened 2016-06-15
add r3c, a Redis Cluster C++ Client, based on hiredis and support standalone

eyjian:
a Redis Cluster C++ Client, based on hiredis and support standalone


eyjian:
@rsanheim

PR #735, opened 2016-06-20
Should be in Node.js section not in Javascript

h0x91b:


h0x91b:
Hi, this patch is ok or should I do something extra?

badboy:
All ok.

badboy:
Just triggering CI once more.

h0x91b:
Thanks

Issue #734, opened 2016-06-19
WATCH accounts for modifications by its own connection 


ProgerXP:
Documentation about Redis transactions (http://redis.io/topics/transactions) says this:

> ### Optimistic locking using check-and-set
> 
> WATCHed keys are monitored in order to detect changes against them. If at least one watched key is modified before the EXEC command, the whole transaction aborts [...]

The above suggests that it doesn't matter who changes the watched key(s) - the same client (that has watched them) or another client. 

But the docs have also this fragment:

> ### WATCH explained
> 
> [...] we are asking Redis to perform the transaction **only if no other client** modified any of the WATCHed keys [...]

This explicitly says that WATCH will ignore changes by the same client but doesn't say _who is that other client_.

I have tried this (single session - single "client"?):

```
127.0.0.1:6379> watch o
OK
127.0.0.1:6379> set o 1
OK
127.0.0.1:6379> multi
OK
127.0.0.1:6379> get o
QUEUED
127.0.0.1:6379> exec
(nil)
```

is this a bug in WATCH or the documentation is misleading? Should we remove this "only if no other client" reference from the docs?

badboy:
At this point I'd say *misleading documentation*, as the behaviour probably was always that every key modification counts.
cc @antirez

ProgerXP:
After some research I see why the docs mention "other clients":
- If the same client does watch-set-multi-exec then it fails (as in my first post) 
- If the same client does watch-multi-**set**-exec - *it succeeds*
- If client A does `watch`, client B does `set`, client A does multi-exec then it fails
- If client A does `watch`, client B does multi-set-exec, client A does multi-exec - it also fails (compare with case 2)

In other words, the docs should say that EXEC will fail in all cases when the key is modified unless it was done by the same client _inside the watched multi/exec block_.

ProgerXP:
To summarize my previous post, I'm testing php-redis extension (https://github.com/phpredis/phpredis):
```PHP
  function testWatchChangeBySelfBeforeMulti() {
    $redis = new Redis;
    $redis->connect('127.0.0.1');
    $redis->delete('o');
    $redis->watch('o');
    $redis->set('o', 1);
    $redis->multi();
    $redis->get('o');
    // This fails.
    $this->assertTrue(is_array( $redis->exec() ));
  }

  function testWatchChangeBySelfAfterMulti() {
    $redis = new Redis;
    $redis->connect('127.0.0.1');
    $redis->delete('o');
[...]

PR #737, opened 2016-06-24
Clarified when WATCH doesn't abort EXEC

ProgerXP:
As per #734 WATCH accounts for modifications by its own connection

### DIFF
--- a/topics/transactions.md
+++ b/topics/transactions.md
@@ -195,8 +195,10 @@ there's no need to repeat the operation.
 
 So what is `WATCH` really about? It is a command that will
 make the `EXEC` conditional: we are asking Redis to perform
-the transaction only if no other client modified any of the
-`WATCH`ed keys. Otherwise the transaction is not entered at
+the transaction only if none of the `WATCH`ed keys were modified. 
+(The only exception is that the same client may change these keys
+after `MULTI`. [More on this](https://github.com/antirez/redis-doc/issues/734).) 
+Otherwise the transaction is not entered at
 all. (Note that if you `WATCH` a volatile key and Redis expires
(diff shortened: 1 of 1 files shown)


badboy:
What about changing the sub-sentence into something similar to:

> WATCHed keys might be changed inside the transaction without aborting it.

ProgerXP:
Like this?

```
[we are asking Redis to perform]
the transaction only if none of the `WATCH`ed keys were modified. 
(But they might be changed by the same client inside the transaction without aborting it.
[More on this](https://github.com/antirez/redis-doc/issues/734).) 
[Otherwise the transaction is not entered at all.]
```

I think it's critical to explicitly say they can be changed by the same client only, because if they are changed in a transaction made by another client the WATCH will fail anyway.

badboy:
Yes, do we then really need that link? IMO that issue is too large to give users helpful information in one go.

ProgerXP:
It's definitely helpful because the docs already have a similar link later:
```
Otherwise the transaction is not entered at all. (Note that if you WATCH a volatile key and Redis expires the key after you WATCHed it, EXEC will still work. More on this.)
```

"More on this" here links to http://code.google.com/p/redis/issues/detail?id=270 which was very useful for me to understand the exact case about volatile keys. I believe it's the same with keys changed outside the transaction. The docs don't list examples and tests which the issue does.

It doesn't take up much space, does it?

badboy:
Hm, I see. Alright then.

ProgerXP:
Corrected the text, please check.

badboy:
:shipit:

ProgerXP:
Thank you!

PR #747, opened 2016-07-29
Perl client AnyEvent::Redis::RipeRedis renamed to AnyEvent::RipeRedis. Added Perl client AnyEvent::RipeRedis::Cluster

iph0:
AnyEvent::RipeRedis is the new incarnation of AnyEvent::Redis::RipeRedis.
AnyEvent::Redis::RipeRedis will no longer be supported and must not
be used in new code.

Also added Perl client AnyEvent::RipeRedis::Cluster

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -343,6 +343,17 @@
   },
 
   {
+    "name": "AnyEvent::RipeRedis::Cluster",
+    "language": "Perl",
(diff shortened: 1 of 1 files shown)


badboy:
Merged in 2f0cf34d5adad0d145e951b604a0d0c055614209

PR #752, opened 2016-09-07
Add tornadis client (python) to the client.json

asduj:


badboy:

asduj:
@badboy Ok, "authors" field removed.

PR #762, opened 2016-10-13
add hiredis-vip client for redis cluster

deep011:
as the title


badboy:
We currently link the author name to Twitter. Is [@deep011](https://twitter.com/deep011) your account? If so, fine. If not, please remove that line.

Plus I can't vouch for your implementation, so I can't really give it as a recommendation right now (I don't want to devalue your work). 
How long does it exist already? Are you using it in production? (We don't really have clear rules on what we mark as recommended, so I just try to gather more data)

deep011:
Yes,we used it in our production,and it works very well.
deep011 is my account,i finished the code of the hiredis-vip about a years ago.

PR #770, opened 2016-11-07
Add Pottery to distlock implementations

brainix:
Hi, @antirez!

Mentioning you so that you get a notification.  Just adding my own implementation of Redlock to the list.

Thanks for all of your hard work,
Raj


brainix:
@djanowski - Mentioning you so that you get a ping.  Does this PR look ok?  Thanks in advance!

PR #771, opened 2016-11-08
Redis plugin for IntellJ IDEA

seventh7me:
Redis plugin for IntellJ IDEA, it's compatible with all IntelliJ's products except MPS

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -1,4 +1,14 @@
 [
+  
+  {
+    "name": "Iedis",
+    "language": "Java",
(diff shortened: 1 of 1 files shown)


badboy:
What does it do? There's no documentation at all on the linked page.
This certainly doesn't fit any criteria for `recommended` (though we don't have clear rules here at all)

seventh7me:
Documentation will be available as soon as possible and I'll make another pull request when it's done     👍

PR #772, opened 2016-11-13
Replace gender-specific "he" with "they" when referring to users

igorwwwwwwwwwwwwwwwwwwww:


antirez:
Hello, please could you modify the patch to use "she or he" instead? This
is what we do in other parts of the doc. Thanks for spotting this issue.

igorwwwwwwwwwwwwwwwwwwww:
@antirez I've updated the patch.

antirez:
Thank you, merged

PR #774, opened 2016-11-17
Mention EXPIRE behavior with non-positive timeouts

marcverney:
The current documentation for EXPIRE does not mention that specifying a timeout less than or equal to zero does actually perform a DEL. This PR fixes this issue.

Also see http://stackoverflow.com/q/40636291/2516943

English is not my native language, so feel free to rephrase.

### DIFF
--- a/commands/expire.md
+++ b/commands/expire.md
@@ -23,6 +23,14 @@ that is overwritten by a call like `RENAME Key_B Key_A`, it does not matter if
 the original `Key_A` had a timeout associated or not, the new key `Key_A` will
 inherit all the characteristics of `Key_B`.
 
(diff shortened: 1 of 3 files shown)


badboy:
Phrasing sounds good!
Could you add a similar note to `EXPIREAT`?

marcverney:
Well, I considered the EXPIREAT doc only covered differences from the EXPIRE doc, the latter applying to both command (for the most part). I'm not against this repetition, but do you really think it's necessary?

badboy:
Make it a short note like "A time in the past will delete the key immediately".

badboy:
Perfect. Thank you! :shipit:

itamarhaber:
@marcverney :+1: 
@badboy :beers:

PR #775, opened 2016-11-23
Fix grammar

ajlanghorn:
This sounds much better with an 'a' between "used as" and "database". Fixes
the grammar, which I noticed whilst reading redis.io.


djanowski:
Thanks.

PR #779, opened 2016-12-08
Add GEO's six commands into "redis-client for C++"

zhengshuxin:
The "redis-client for C++" of acl has supported GEO's six commands including "geoadd, geohash, geopos, geodist, georadius, georadiusbymember" functions in class redis_geo.


badboy:
Why not say that it supports all command types?

PR #781, opened 2016-12-13
Added pronunciation

garrows:
At least this is how I've read its pronounced.


garrows:
😢  giving up.

badboy:
@garrows You're closing rebumped this. I re-opened and merged, because why not? :) Thanks for that contribution

PR #783, opened 2016-12-19
reword slavery note for grammar

wkalt:
Changes: "that would" -> "which would"; "remember you about" -> "remind you that"; "but something" -> "and something".


djanowski:
Thank you.

PR #787, opened 2017-01-04
Redis Plugin for IntelliJ platform.

seventh7me:
Iedis is a beautiful, easy-to-use Redis management plugin for IntelliJ platform, working with millions keys without blocking the Redis server.

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -1,5 +1,14 @@
 [
   {
+    "name": "Iedis",
+    "language": "Java",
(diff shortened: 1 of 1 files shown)


badboy:
Right now we link the author to a Twitter profile, therefore it should be a valid Twitter handle. Is the named one yours? If you don't have one, don't worry, just remove the author for now.

badboy:
Oh, this is also better suited for `tools.json`. Open a new PR with those changes please.

PR #790, opened 2017-01-17
Fix cluster spec doc: PFAIL -> FAIL

gianlucaborello:


tevino:
After reading this for several times:
> The majority of masters signaled the PFAIL or PFAIL condition

And compared `PFAIL` with `PFAIL` letter by letter several times:
> PFAIL or PFAIL

badboy:
That sounds about right. Thanks!

PR #768, opened 2016-11-02
Update php-redis-client

cheprasov:
I believe that the client was tested well, I use the client in personal projects, and everyone can use it in they own projects too.
The client has regular tests on latest versions of Redis, like 2.6.17, 2.8.24, 3.0.7, 3.2.4 (and older). Also, it has 644 tests for each version of PHP. And it passed well on the next versions of PHP: 5.5, 5.6, 7.0, 7.1, hhvm. ( https://travis-ci.org/cheprasov/php-redis-client )
And the client has good performance ( https://github.com/cheprasov/php-redis-client-vs-predis-test ).
Also, I am watching for Redis and trying to add new features to the client quickly and efficiently.
And I recommend to use this client.
Thanks.

PR #795, opened 2017-02-05
Removed recommendation flag for clients

cheprasov:
Let's remove the recommendation flag from clients, because you currently don't have clear rules on what you mark as recommended.
In addition, the recommendation flag makes uncompetitive of other clients.
I think it would be great if people choose the client is not based on you tips, because you do not know all the programming languages for to give right client recommendation.
Thanks,
The PR history https://github.com/antirez/redis-doc/pull/768

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -5,7 +5,6 @@
     "repository": "https://github.com/redis/redis-rb",
     "description": "Very stable and mature client. Install and require the hiredis gem before redis-rb for maximum performances.",
     "authors": ["ezmobius", "soveran", "djanowski", "pnoordhuis"],
-    "recommended": true,
     "active": true
   },
 
@@ -15,7 +14,6 @@
     "repository": "https://github.com/ptaoussanis/carmine",
     "description": "Simple, high-performance Redis (2.0+) client for Clojure.",
     "authors": ["ptaoussanis"],
-    "recommended": true,
     "active": true
   },
   {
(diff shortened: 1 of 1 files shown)


badboy:
I'm completely in favor of this change. At the same time we should probably remove that column in the HTML output as well.

cc @antirez @djanowski

charsyam:
@cheprasov @badboy I also agree with you :)

cheprasov:
Added PR https://github.com/antirez/redis-io/pull/115 for removing that column in the HTML

cheprasov:
up

antirez:
I don't like this idea *if not* immediately replaced (at the same time we remove the recommendation) by an automated system that gives an automated score based on latest commits activity and number of stars in the latest N months + number of star in total. There is to come up with a good enough formula, but it's important that immediately an user can follow what looks like the "to go" solution instead of wasting time exploring abandoned clients.

antirez:
Basically... something like that:

1. The activity *smailing face* remains the same, but becomes a count of number of commits in latest 3 months. If over a given threshold the client is marked as active.
2. The star instead will be removed completely, we'll just write near the client name (and sort by) the number of stars on Github in the latest 6 months.

Moreover all the clients that are not released under an OSS license should be removed.

l4u:
@antirez Should we reopen this or create a new issue for removing the stars?
Some of the stars can be misleading to new users who want to pick a recommended client.

PR #802, opened 2017-03-02
Update clients.json to include link to D's vibe.d framework.

stonemaster:


djanowski:
Not really a client, but thank you for using Redis!

PR #806, opened 2017-03-18
corrent cluster-spec.md about 'Practical example of configuration epo…

spccold:
corrent cluster-spec.md about 'Practical example of configuration epoch usefulness during partitions'


djanowski:
@antirez +1?

PR #807, opened 2017-03-18
Add QRedis

tiagocoutinho:
I have started working on a python, qt based GUI for redis.

In development state but already pretty much usable

### DIFF
--- a/tools.json
+++ b/tools.json
@@ -305,6 +305,13 @@
     "authors": ["u_glide"]
   },
   {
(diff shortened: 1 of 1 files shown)


badboy:
It would fit better in [tools.json](https://github.com/antirez/redis-doc/blob/master/tools.json)

tiagocoutinho:
Thanks @badboy. I've done like you suggested

PR #809, opened 2017-03-22
Grammar edits in cluster-failover and data-types-intro

kraftman:


djanowski:
I grew fond of some of these grammar errors :) But thank you very much for fixing.

PR #810, opened 2017-03-22
Added commas in ops/sec

stockholmux:
The added commas make it more clear the performance difference between Pi 3 and Pi 1/B. I had to look a few times to see that Test 1 was x10+ on the pi3 because it was just zeros.


djanowski:
Thank you.

PR #822, opened 2017-04-23
Adds @badboy to the hiredis team

itamarhaber:


charsyam:
@itamarhaber looks good, but why do you add '_' after of badboy?

badboy:
@charsyam because that's my nick ;)

charsyam:
@badboy Ah, I just think your nickname is badboy. It was a chance to merge it by me. :)

badboy:
It's with an underscore on Twitter, IRC and elsewhere

itamarhaber:
Underscore all over, except here :)

Issue #736, opened 2016-06-22
Bad math in data-types-intro.md 


cybersam:
The "Lexicographical scores" section of data-types-intro.md makes the astonishing claim that you can store a 128-bit number in 8 bytes.

In reality, since a byte is only 8 bits wide, you can only store a 64-bit number in 8 bytes.

itamarhaber:
In reality, you can simply submit a PR to replace 8 with 16 and fix this issue with it.

PR #825, opened 2017-04-30
Fix issue #736: 128-bits requires 16 bytes of storage

charleswhchan:


charleswhchan:
Thanks @badboy

badboy:
thanks for the contribution

PR #829, opened 2017-05-13
Update clients.json

stefb965:


badboy:
Repo doesn't exist. No explanation why this change would be needed. Please don't modify other entries.

PR #830, opened 2017-05-14
Update modules.json to add new modules by @datokenizer

daTokenizer:


badboy:
Merged as eda06668bf1e28370f1d5200c3cf7e68dc849f3a

PR #836, opened 2017-06-25
Add an MIT licensed RESP parser and Redis client

boazsegev:
facil.io includes an asynchronous [Redis client](https://github.com/boazsegev/facil.io/tree/master/lib/redis#getting-started-with-the-redis_engine) as well as a [RESP parser/formatter](https://github.com/boazsegev/facil.io/tree/master/lib/redis#the-resp-parser--formatter) that could be used as an independent module.

The parser and client were written from the ground up and are licensed under the MIT license.

This places fail.io in a unique position, the could provide an MIT licensing option where required. This is in contrast to most of the other clients, that are based on hiredis and require the BSD-3-clause license.

This might not be the fastest client or parser (it's more concerned with protecting itself from bad code than with being optimized), but it is (to the best of my knowledge) the only MIT licensed option.

Thanks! 👍🏻

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -1432,7 +1432,7 @@
     "name": "r3c",
     "language": "C++",
     "repository": "https://github.com/eyjian/r3c",
-    "description": "A Redis Cluster C++ Client, based on hiredis and support standalone, it's easy to make and use, not depends on C++11 or later.",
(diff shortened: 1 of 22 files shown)


badboy:

boazsegev:
Oh :-( ... the whole rebasing got messed up... I'm sure I turned left when I should have turned right, I just don't know where...

I should probably just start a new PR (or two)...

PR #837, opened 2017-06-25
Add iodine (Ruby) Websocket server with Pub/Sub and integrated Redis client

boazsegev:
Iodine is a Ruby HTTP / Websocket server with native Pub/Sub (built using the `facil.io` C framework).

Iodine includes an integrated Redis client that allows Pub/Sub to be extended across machine boundaries (the native Pub/Sub is limited to the process cluster).

Thanks! 🙏🏻

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -1432,7 +1432,7 @@
     "name": "r3c",
     "language": "C++",
     "repository": "https://github.com/eyjian/r3c",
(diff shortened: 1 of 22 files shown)


badboy:

While you're at it, please rebase against master.

boazsegev:
Oh :-( ... the whole rebasing got messed up... I'm sure I turned left when I should have turned right, I just don't know where...

I should probably just start a new PR (or two)...

PR #850, opened 2017-07-28
Add redisio for language Python

cf020031308:
redisio is a tiny, fast and fully-documented redis client which treats data in consistence with the protocol.


itamarhaber:
Hello @cf020031308 

itamarhaber:
Resolved via 842a34b

PR #851, opened 2017-07-29
Documents the `MEMORY` command

itamarhaber:
A stab a documenting v4's `MEMORY` command /cc @antirez 

Additionally, as part of this effort it would be good to update `INFO`'s page - ~currently marked as a WIP commit~ requires an expert review for correctness.

TODO:
1. Use this opportunity to document active defragmentation as well 

Additional ponderings:

1. Is 'server' the right group or is a new one needed
2. Should 'memory help' be documented (ATM yes), but also applies to 'debug help' etc
3. Speaking of debug, there's quite a few subcommands missing from the docs. Should they be documented? They're temporary but temporary things tend to last forever...
4. I see that that redis/unstable has 'memory doctor' in the help, whereas 4.0 doesn't although it is supported by the server - is that intentional or an omission (i.e. PR to antirez/redis)?
5. Should `memory` be enabled in the website's interactive mode?
6. `MEMORY USAGE` and the modules API - should the connection to ModuleMemoryUsageFunc be mentioned, perhaps as a note or some templatized "@module-api" kind of thing across the corpus?
7. Some of the docs/commands are `jemalloc` specific - perhaps a way to templatize/organize this can be added

### DIFF
--- a/commands/info.md
+++ b/commands/info.md
@@ -88,24 +88,30 @@ Here is the meaning of all fields in the **memory** section:
 *   `used_memory_rss_human`: Human readable representation of previous value
 *   `used_memory_peak`: Peak memory consumed by Redis (in bytes)
 *   `used_memory_peak_human`: Human readable representation of previous value
-*   `used_memory_peak_perc`: TBD
-*   `used_memory_overhead`: TBD
-*   `used_memory_startup`: TBD
-*   `used_memory_dataset`: TBD
-*   `used_memory_dataset_perc`: TBD
-*   `total_system_memory`: TBD
+*   `used_memory_peak_perc`: The percentage of `used_memory_peak` out of
+     `used_memory`
+*   `used_memory_overhead`: The sum in bytes of all overheads that the server
+     allocated for managing its internal data structures
+*   `used_memory_startup`: Initial amount of memory consumed by Redis at startup
+     in bytes
+*   `used_memory_dataset`: The size in bytes of the dataset
+     (`used_memory_overhead` subtracted from `used_memory`)
+*   `used_memory_dataset_perc`: The percentage of `used_memory_dataset` out of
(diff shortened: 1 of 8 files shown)


itamarhaber:
ping @antirez

itamarhaber:
@antirez I just noticed that the implementation of `MEMORY USAGE` **doesn't** include the key itself (i.e. strlen + perhaps the global dict entry). Is that intentional? If so, the doc in the PR should be fixed.

commands/info.md
```diff
@@ -182,15 +193,15 @@ Here is the meaning of all fields in the **stats** section:
...
+*   `total_net_output_bytes`: The total number of bytes written to the network
+*   `instantaneous_input_kbps`: The network's read rate per second in KB/sec
+*   `instantaneous_output_kbps`: The network's write rate per second in KB/sec
 *   `rejected_connections`: Number of connections rejected because of
      `maxclients` limit
-*   `sync_full`: TBD
-*   `sync_partial_ok`: TBD
-*   `sync_partial_err`: TBD
+*   `sync_full`: The number of full resyncs with slaves
+*   `sync_partial_ok`: The number of accpepted partial resync requests
```
boughtonp: Type - "accpepted" should be "accepted"

itamarhaber: True - thanks!

itamarhaber: BTW, @boughtonp, it would be better if you can make future reviews at the origin branch so I can commit them directly to the PR.

boughtonp: Ok, I'll keep that in mind for future.

avnerbarr:
is this going to be merged to the main site documentation?

antirez:
Merged, sorry for the delay, thanks @itamarhaber!

PR #861, opened 2017-08-31
Add Redis 3.2 and 4.0 config links to config docs

NickCraver:
https://redis.io/topics/config needs some link love for v4, this adds that and normalizes the periods on the end of the lines.


NickCraver:
Also: 3.2 includes the bind bits where 3.0 does not, I guess we should toss a link in there for that version as well?

antirez:
Thanks! I merged this for now, but probably the best thing is to remove 3.0 at all and add 3.2, 3.0 at this point is *olad* :-)

PR #866, opened 2017-09-07
Update clients.json

gistao:
Hi:
a new redis client for golang,thanks!


itamarhaber:
Hello @gistao 

PR #867, opened 2017-09-07
Added Noderis client

wallneradam:
A new Node.js client


itamarhaber:
Hello @wallneradam 

PR #871, opened 2017-10-06
Added note about deprecation

stockholmux:


badboy:
Could you change the `hset` documentation as well?

itamarhaber:
Hello @stockholmux - would you like to complete this effort? If so, we'll be able to merge it and maybe close #1063.

stockholmux:
@itamarhaber After must consternation, i've updated `hset`

commands/hmset.md
```diff
@@ -3,6 +3,8 @@ Sets the specified fields to their respective values in the hash stored at
 This command overwrites any specified fields already existing in the hash.
 If `key` does not exist, a new key holding a hash is created.
 
+As per Redis 4.0.0, HMSET is considered deprecated. Please use HSET in new code.
```
itamarhaber: Please use backticks around `HSET` to make it render as a link.

commands/hset.md
```diff
@@ -2,12 +2,11 @@ Sets `field` in the hash stored at `key` to `value`.
...
 
-@return
+As of Redis 4.0.0, HSET is variadic and allows for multiple `field`/`value` pairs.
 
-@integer-reply, specifically:
+@return
 
-* `1` if `field` is a new field in the hash and `value` was set.
-* `0` if `field` already exists in the hash and the value was updated.
+@integer-reply: The number of fields that were added or updated.
```
itamarhaber: "or updated" is incorrect empirically and from a cursory code examination - is there a citable source?

commands.json
```diff
@@ -1217,19 +1217,16 @@
   },
   "HSET": {
     "summary": "Set the string value of a hash field",
-    "complexity": "O(1)",
+    "complexity": "O(N) where N is the number of fields being set",
```
itamarhaber: One final (?) request - let's go with `SADD`-like complexity in this (and other variadics that need updates). So something like: "O(1) to set a single field. When called with multiple arguments, O(N) where N is the number of fields to set."

stockholmux: changed. O(n) for field/value pairs. It's a little weird, but I think it's the best language now.

itamarhaber:
Thanks - merging.

PR #882, opened 2017-11-23
Added tmux as a terminal multiplexer suggestion to the Quickstart guide

JoshuaRLi:
Added `tmux` and links to `screen`, `tmux`. Explained that they are terminal multiplexers.

[`tmux`](https://github.com/tmux/tmux) is one of the most popular `screen` alternatives out there.

### DIFF
--- a/topics/quickstart.md
+++ b/topics/quickstart.md
@@ -149,7 +149,7 @@ Running Redis from the command line is fine just to hack a bit with it or for
 development. However at some point you'll have some actual application to run
 on a real server. For this kind of usage you have two different choices:
 
-* Run Redis using screen.
+* Run Redis using [`screen`](https://www.gnu.org/software/screen/), [`tmux`](https://github.com/tmux/tmux/wiki), or any other terminal multiplexer.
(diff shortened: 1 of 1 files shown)


badboy:
Yeah, .... we should remove that recommendation in total. People will do it and that's bad.

JoshuaRLi:
I don't think it should be removed. Using a tmuxer is definitely not what you should do in production, but for a lot of beginners and hobbyists it's convenient.

I think there's merit in that, but to address your concerns perhaps it should also be mentioned tmuxers are recommended for up to hobbyist purposes only. Does that sound fair?

boughtonp:
The section is "_Installing Redis more properly_" and the suggestion is preceeded by _"at some point you'll have some actual application to run on a real server. For this kind of usage you have two different choices"_ - so whilst I have no view either way on whether a section recommending screen/tmux/etc is useful elsewhere, I definitely think this section should be simplified to only mention setting up the service.

JoshuaRLi:
Hmm, okay. I agree. I'll likely submit another PR that removes `screen` and rewords things so that service setup is the only recommended method.

PR #887, opened 2017-12-01
Added Redis Cluster client Redis::ClusterRider for Perl

iph0:
Added Redis Cluster client Redis::ClusterRider for Perl


itamarhaber:
Hello @iph0 

Sorry for the belated reply - merging, thanks,

PR #888, opened 2017-12-11
Makes counter pattern no. 1 concurrent

### LINKED ISSUE #777: Command INCR - Pattern: Rate limiter
Hi,

The Pattern: Rate limiter in command INCR doc [[1](https://github.com/antirez/redis-doc/blob/master/commands/incr.md#pattern-rate-limiter)] has a case to be invalid.

When the request count haven't hit the quota, if fire huge of requests at the same time, most/many of those may cross the validation. The time window exists between after the checking and updating the counter. I verified the issue by created num of threads and execute the example code in Java.

It can be fixed by put in a Lua script like:
```
local current = tonumber(redis.call("get", "the key"))
if (current ~= nil and current >= 10) then
    error("too many requests per second")
end
redis.call("incr", "the key")
redis.call("expire", "the key", 10)
```

This is a common case for heavy traffic system/API or protected from attack. I think it's better to update this or mention in the doc.

Thanks!

itamarhaber:
... by adding `WATCH`. Also resolves #777 kind of.

In addition, fixes a couple of typos.

### DIFF
--- a/commands/incr.md
+++ b/commands/incr.md
@@ -69,6 +69,7 @@ The more simple and direct implementation of this pattern is the following:
 FUNCTION LIMIT_API_CALL(ip)
 ts = CURRENT_UNIX_TIME()
 keyname = ip+":"+ts
+WATCH(keyname)
 current = GET(keyname)
 IF current != NULL AND current > 10 THEN
     ERROR "too many requests per second"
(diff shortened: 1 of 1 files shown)


commands/incr.md
```diff
@@ -69,25 +69,32 @@ The more simple and direct implementation of this pattern is the following:
 FUNCTION LIMIT_API_CALL(ip)
 ts = CURRENT_UNIX_TIME()
 keyname = ip+":"+ts
+WATCH(keyname)
 current = GET(keyname)
 IF current != NULL AND current > 10 THEN
     ERROR "too many requests per second"
```
Pyrolistical: Need to `UNWATCH()`

Pyrolistical: Need to `UNWATCH()`. Every `WATCH` need a matching `UNWATCH`, `EXEC`, or `DISCARD`

commands/incr.md
```diff
@@ -69,25 +69,32 @@ The more simple and direct implementation of this pattern is the following:
...
     MULTI
         INCR(keyname,1)
         EXPIRE(keyname,10)
-    EXEC
-    PERFORM_API_CALL()
+    reply = EXEC
+    IF reply != NULL
+        PERFORM_API_CALL()
+    ELSE
+      RETURN LIMIT_API_CALL(ip)
```
Pyrolistical: I wrote this differently: https://github.com/antirez/redis-doc/pull/893/files#diff-70e66b9c0c88bb9f371d12bb2bb9e4d8R112

It is more clear what the intent is. Making the recursive call directly is an implementation detail

nermiller:
@itamarhaber, do we want to close and reopen?

Issue #892, opened 2017-12-19
Command INCR Pattern: Rate Limiter 2


Pyrolistical:
I think rate limiter 2 can be implemented without the memory leak, and thus the lua script is not needed:

```
FUNCTION LIMIT_API_CALL(ip):
current = GET(ip)
IF current != NULL AND current > 10 THEN
    ERROR "too many requests per second"
ELSE
    MULTI
        SET(ip,0,EX 1,NX)
        INCR(ip)
    EXEC
    PERFORM_API_CALL()
END
```

The counter is init to 0 only if it doesn't exist, then increment. if it already exists, just increment. since this is in a multi, its impossible for expire to happen between a noop set and incr.

PR #893, opened 2017-12-19
Simplified rate limiting 2

Pyrolistical:
Split list implementation into rate limiting 3

Fix for issue: https://github.com/antirez/redis-doc/issues/892


Pyrolistical:
I added `Rate limiter 3`, but it looks out of place. It doesn't use `INCR`. This belongs somewhere else

PR #894, opened 2017-12-21
Example incorrectly shows i5 rather than i8

alvinr:
The example does not match the text of the page, therefore to make the example consistent the following change is required

```
> BITFIELD mykey INCRBY i5 100 1 GET u4 0
```

```
> BITFIELD mykey INCRBY i8 100 1 GET u4 0
```


itamarhaber:
Hello @alvinr 

Thanks for this PR - I see the issue was already resolved via 39460eac1a4d079f7d938b8e2477e9f0599ecdd8 so closing this.

PR #900, opened 2018-01-25
Don't say "simply"

jameshfisher:
I believe that the word "simply" in documentation is harmful. At best, "simply" is unnecessary filler, and at worst, it's an inflammatory word which shames readers. I'm on a bit of a mission to remove "simply" from software documentation, and as part of that, I created [a script called `dont-say-simply`](https://github.com/jameshfisher/dont-say-simply) which removes all uses of the word "simply" from a GitHub repository. That script created this PR. You can read more about my reasoning [here](https://github.com/jameshfisher/dont-say-simply).

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -629,7 +629,7 @@
   },
   "DUMP": {
     "summary": "Return a serialized version of the value stored at the specified key.",
-    "complexity": "O(1) to access the key and additional O(N*M) to serialized it, where N is the number of Redis objects composing the value and M their average size. For small string values the time complexity is thus O(1)+O(1*M) where M is small, so simply O(1).",
+    "complexity": "O(1) to access the key and additional O(N*M) to serialized it, where N is the number of Redis objects composing the value and M their average size. For small string values the time complexity is thus O(1)+O(1*M) where M is small, so O(1).",
     "arguments": [
       {
         "name": "key",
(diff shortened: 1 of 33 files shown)


itamarhaber:
Simply put, I like it.

badboy:
There's definitely no loss in clarity by simply removing that word. Thanks!

antirez:
Please could you revert @badboy? I wrote most of this doc and I don't agree.

badboy:
[¯\_(ツ)_/¯](https://github.com/antirez/redis-doc/commit/38c2bd6f5331aad382494b3c7833b0cc2b92222a)

antirez:
@jameshfisher I understand that you have a battle against "simply", but I don't have the same battle, I believe that the word simply actually improves things. For instance if I say:

"Redis replication can be configured simply by sending the SLAVEOF command to a slave instance"

It implies that there are *no other operations* to perform, while the users may expect so from other systems. If you remove the word "simply" people may wonder if there is more to do, based on other systems.

But in general, while you are offended by the word "simply" in documentation, I'm offended by the fact that people say to other people how to write text based on *personal believes*. This is extremely problematic for me. But fortunately nor I've to say you want to do, nor the reverse, we can act doing what we believe is the right thing, and I believe that "simply" is not an offending word. However we have to both obey the law about what can be done and what now. Being the word "simply" not banned, and me not agreeing with your point, I'm deciding as a free individual that I'll keep it in the Redis documentation.

PR #903, opened 2018-02-08
Make the format of variable names more consistent

stoman:
I just read the article contained in the file `twitter-clone.md`. Thank you for writing this nice introduction. I found some inconsistencies while reading and propose to make some format changes.

While most variable names and commands are highlighted using backticks, few of them are also wrapped in asterisks or single quotes. I propose to adjust the format to be more consistent to make the article easier to read.


itamarhaber:
Thanks - merging.

PR #914, opened 2018-03-20
Document that SIGINT is handled like SIGTERM

jey:
Document that SIGINT is handled like SIGTERM, as indicated in https://github.com/antirez/redis/blob/8d92885bac17d8d82177a2f51d5b2a03c11ac8fe/src/server.c#L3488


jey:
Bump. This PR documents existing behavior that is not currently documented. Thanks

PR #921, opened 2018-04-24
Fix minor typos and line lengths in mass-insert.md

linbug:


itamarhaber:
Hello @linbug - sorry for the belated reply. The typo fixes are good, but no need for the reformatting of lines. Would you care to revert that?

PR #928, opened 2018-05-24
SET signature containing misleading "expiration"

herzrasen:
@antirez & @itamarhaber sorry, I accidentally used "command" instead of "name". Fixed it.


itamarhaber:
Thanks - merging.

PR #931, opened 2018-06-07
Documents the MODULE command(s)

itamarhaber:


itamarhaber:
@stockholmux @bmansheim please review this.

commands/module-list.md
```diff
@@ -0,0 +1,10 @@
+Returns information about the modules loaded to the server.
+
+@return
+
+@array-reply: list of loaded modules. Each element in the list represents a
```
stockholmux: Not a big deal, but I would probably avoid the term 'list' as it can be confused for the data structure.

commands/module-load.md
```diff
@@ -0,0 +1,14 @@
+Loads a module from a dynamic library at runtime.
+
+This command loads and initializes the Redis module from the dynamic library
+specified by the `path` argument. The `path` should be the absolute path,
+including the full filename, of the library. Any additional arguments are passed
```
stockholmux: I'd reword this. The comma separated aside is awkward and gets in the way of understanding. One solution: If you move the aside to the end, I think it makes more sense.

commands/module-unload.md
```diff
@@ -0,0 +1,13 @@
+Unloads a module.
+
+This command unloads the module specified by `name`. Note that the module's name
+is reported by the `MODULE LIST` command, and may differ from the dynamic
+library's filename.
+
+Known limitations:
+
+*   Modules that register custom data types can not be unloaded.
```
stockholmux: IIRC, If you load a module, write some keys with the module data type then restart the server, it will choke trying to restore from RDB because of the missing data types. Should this be in the limitations?

commands/module-list.md
itamarhaber: As it already in the documentation in other commands at the same place (see `HGETALL`). I'd rather stick with the "standard" than be "technically correct" (debest kind of!)

commands/module-load.md
itamarhaber: Agreed.

commands/module-unload.md
itamarhaber: I don't think so... you can't load an RDB after unloading a module (unless you use DEBUG). That said, I believe that limitation is buried somewhere in the modules' docs (need to find ref though). In any case, once https://github.com/antirez/redis/pull/5926 is released we'll need to update this.

stockholmux: @itamarhaber fair enough. I'm pretty sure that limitation exists if you stop the redis server, change the redis.conf then restart it again. But, I suppose that is a different situation than `MODULE UNLOAD` (which I rarely use)

PR #932, opened 2018-06-09
Swap ZPOP pair order

rufoa:
Update command description to reflect this change https://github.com/antirez/redis/commit/0a698e499a0b543b350544c63701ab944151c6ed


antirez:
Thanks!

PR #949, opened 2018-07-25
A couple of typo fixes in streams-intro

paulsnar:


paulsnar:
Please merge #950 instead.

Issue #953, opened 2018-08-07
Typos, spelling and grammar in replication.md


m3talstorm:
#954

m3talstorm:
#954 953

itamarhaber:
Resolved via #1341

PR #954, opened 2018-08-07
#953 Spelling, typos and grammar

m3talstorm:


adjenks:
Additional:

### /topics/persistence:
> It is possible that the server crashed while writing the AOF file, or that the volume where the AOF file is stored is store was full.

### /topics/memory-optimization:
> However allocators are smart and are able to reuse free chunks of memory, so after you freed 2GB of your 5GB data set, when you start adding more keys again, you'll see the RSS (Resident Set Size) to stay steady and don't grow more, as you add up to 2GB of additional keys.

itamarhaber:
Hello @m3talstorm - thanks for this and sorry for the late reply. Please rebase so it can be reviewed and merged.

itamarhaber:
Resolved via #1341

PR #967, opened 2018-09-03
Fix typos

Abioy:
Add `lenN` before `argN`, and use `argc` instead of `args` to represent number of arguments.


itamarhaber:
Hello @Abioy - thanks for the contribution.

That example illustrates the generic case, and the one right after it gives a concrete use with the `SET` command. Therefore, I don't think this edit is needed.

Abioy:
> Hello @Abioy - thanks for the contribution.
> 
> That example illustrates the generic case, and the one right after it gives a concrete use with the `SET` command. Therefore, I don't think this edit is needed - please let me know if it can be closed.

Disagree with the reason, but ok if you close the pr. Thx for you reply.

PR #973, opened 2018-09-16
Fixes SET's syntax

itamarhaber:
Changes outputted syntax from the current:

```
SET key value [expiration EX seconds|PX milliseconds] [NX|XX]
```

to:

```
SET key value [EX seconds|PX milliseconds] [NX|XX]
```

Signed-off-by: Itamar Haber <itamar@redislabs.com>

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -2237,7 +2237,7 @@
         "type": "string"
       },
       {
-        "command": "expiration",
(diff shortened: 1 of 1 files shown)


itamarhaber:
Fixed by https://github.com/antirez/redis-doc/pull/928

PR #975, opened 2018-09-17
RedisModule_HashSet must end with NULL

gkorland:
Extended the RedisModule_HashSet docs to mark that each call must end with NULL


itamarhaber:
Hello @gkorland 

This file is automatically generated from the comments in the source code - please apply the change there instead of here.

PR #979, opened 2018-09-24
Fix typo: remove unnessary a

hayashier:
Fix typo: removing unnesessary "a".
"that operates a on a database number" -> "that operates on a database number"


itamarhaber:
Fixed via #919

PR #980, opened 2018-09-24
Fix some mistake of modules intro

hayashier:
This PR includes following 2 type of fixes

1. Although documents says the example of "INCRBY", but example code uses "INCR"

- Document says following

```
For example if I want to call INCRBY
```
- Actual example code

```
RedisModuleCallReply *reply;
reply = RedisModule_Call(ctx,"INCR","sc",argv[1],"10");
```

2.  Now it includes the unnessary character "se", so this PR remove this.

Finally in order to set the length se use the following call:
-> Finally in order to set the length, use the following call:

### DIFF
--- a/topics/modules-intro.md
+++ b/topics/modules-intro.md
@@ -270,7 +270,7 @@ number "10" as second argument (the increment), I'll use the following
 function call:
 
     RedisModuleCallReply *reply;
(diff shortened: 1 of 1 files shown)


itamarhaber:
@hayashier - thanks, merging.

PR #984, opened 2018-09-24
Remove Commons Clause modules

vmbrasseur:
Hi, all. Vice President of the Open Source Initiative (OSI) here.

The Commons Clause, when applied to an OSI-approved license, creates a new license that violates the Open Source Definition (OSD). The OSI does not approve licenses that violate the OSD and does not approve of the Commons Clause in general. Therefore modules with licenses that include the Commons Clause are not OSI-approved.

As this page says it lists only OSI-approved licenses, I've removed the Commons Clause-using modules from the list so that the page is now accurate.


antirez:
Hello, thanks @vmbrasseur, the current situation of the module page stating that only OSS modules are listed but instead reports the Common Clause modules is wrong indeed, however I don't think I'll remove the Redis Labs modules, I would like to rather find a way to enlist them as well, for two main reasons:

1. They are the most notable modules in the Redis ecosystem.
2. They could be seriously useful to Redis users, and are free as beer.

However how to change the policy is currently not clear, if to allow any commercial module to be listed or just "source available and free as beer" modules. However I plan to highlight such modules as **non OSS modules**.

vmbrasseur:
Removing the "only OSI-approved" bit would accomplish the immediate goal, but I like your additional plan to flag the proprietary-licensed modules as **non OSS modules**.

Those two things would keep the popular modules on the list, make their licensing clear to potential users, and be accurate from an OSI point of view.

I'm happy to update my PR to do these two things instead, if you'd like?

SirCmpwn:
Hey @antirez, can you answer @vmbrasseur's questions? The current page is disengenous and should probably be quickly corrected.

natoscott:
@vmbrasseur @antirez some of the Redis Labs' modules have been forked and are being maintained as open source software, it would be great if these could be mentioned on the modules page as well.

https://goodformcode.com/

SirCmpwn:
Bump.

antirez:
Hello, I did two things:

1. Modified the wording in the web site to specify that non OSI approved licenses are specified in the list.
2. Mark all the non OSI modules clearly in the listing.

The changes should be online in a matter of minutes as the site fetches the new version from Git.

Cheers,
Salvatore

ddevault:
Thanks! Took your time...

ddevault:
Next up: https://github.com/RedisLabsModules/RediSearch/issues/518

PR #997, opened 2018-10-07
BZPOPMAX and BZPOPMIN: correcting the order of returned values

RoeyPrat:
it seems in BZPOPMAX and BZPOPMIN documentation the order of returned values is incorrect.
The first value is the name of the key where a member was popped (in both docs and redis).
However docs state that `the second element being the score of the popped member`, when it is really the third element.


itamarhaber:
Hello @RoeyPrat 

Thanks - you are correct, we changed the implementation and the docs are indeed lagging.

itamarhaber:
Thanks - merging.

PR #1004, opened 2018-10-15
XPENDING - missing word 'group'

RoeyPrat:
first line reads consumer name, when i think you mean consumer group name :)


antirez:
Thanks!

PR #1006, opened 2018-10-18
Cluster tutorial: require Redis 5.0 or higher

jaimecbernardo:
cluster tutorial: require redis 5.0 or higher

Changes the minimal required version of Redis to 5.0, in which the
'redis-cli --cluster' commands mentioned in the tutorial were made
available.

### DIFF
--- a/topics/cluster-tutorial.md
+++ b/topics/cluster-tutorial.md
@@ -12,7 +12,7 @@ However this tutorial tries to provide information about the availability
 and consistency characteristics of Redis Cluster from the point of view
 of the final user, stated in a simple to understand way.
 
-Note this tutorial requires Redis version 3.0 or higher.
+Note this tutorial requires Redis version 5.0 or higher.
 
 If you plan to run a serious Redis Cluster deployment, the
 more formal specification is a suggested reading, even if not


antirez:
Hello, thanks but it's better to just state that redis-cli --cluster requires 5.0 and that >= 3.0 users can continue to use "redis-trib". I think I already have a pull request for that from @artix75 that I will be merging tomorrow.

antirez:
@jaimecbernardo sorry I think you are right, I think I already merged some changes but this was indeed a missing part. What we should do probably is to show the two forms, and say that the rest of the documentation will use redis-cli instead of redis-trib, but that the same can be achieved with redis-trib as well. Or... alternatively, since Redis 5 is still new, we could show both forms in every example.

antirez:
@jaimecbernardo just tried to fix it somewhat... Please if you have 5 minutes tell me if it makes sense and what we could change.

jaimecbernardo:
Hi @antirez,

Thank you. This is much clearer.
There are still more zones with `redis-cli --cluster` commands but at least people who are following the tutorial will know of the existence of `redis-trib.rb help` from the "Creating the cluster" section to look into the alternative.

I consider this PR is no longer needed, so I'll close it.

antirez:
Thanks @jaimecbernardo, I'll try to improve things in the future because checking it today I see that this tutorial is really *old*, and far from being optimal for the new needs. A documentation effort is needed ASAP. Thanks.

PR #1014, opened 2018-10-21
Redis modules introduce: fix typo in "creating new keys" example

charpty:
In chapter "Arity and type checks", there is a mistake: 
```
To create a new key, open it for writing and then write to it using one of the key writing functions. Example:

RedisModuleKey *key;

// I think here should be REDISMODULE_WRITE for create new key
key = RedisModule_OpenKey(ctx,argv[1],REDISMODULE_READ);

if (RedisModule_KeyType(key) == REDISMODULE_KEYTYPE_EMPTY) {
    RedisModule_StringSet(key,argv[2]);
}
```

Looking forward to your reply. D#-#

Signed-off-by: charpty <charpty@gmail.com>

### DIFF
--- a/topics/modules-intro.md
+++ b/topics/modules-intro.md
@@ -573,7 +573,7 @@ To create a new key, open it for writing and then write to it using one
 of the key writing functions. Example:
 
     RedisModuleKey *key;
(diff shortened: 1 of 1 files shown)


antirez:
Thank you @charpty !

PR #1018, opened 2018-10-22
 XREAD time complexity should be constant

gkorland:


gkorland:
@antirez Assuming I'm getting the implementation right (based on Radix Tree) look up is not relative to the amount of words in the tree but to the length of the word (which is the length of the ID)

antirez:
Hello @gkorland! I already fixed it! Sorry the complexity was cut&pasted from sorted sets for error. Please check if the new descriptions fit.

PR #1019, opened 2018-10-23
Fix groupname argument of XGROUP

zetaben:
Hello,

While reading the awesome Redis documentation I stumbled on what I thought was a typo in an example (DELCONSUMER). Instead of just fixing I thought it would be better for understanding to always use the same groupname on that page. 

Checking the other examples I actually stumbled on a bigger issue; the command header seem to miss the groupname when using SETID. From the source it actually appears to be required here (cf first commit). 

Anyway feel free to close if you actually wanted different group name in the examples but I believe the SETID part is actually an error.

Thanks again for this amazing project & piece of software! :tada: 

PS: This issue is also present in https://github.com/antirez/redis/blob/ad78b50f62c88b6396c5ee86cda89fc2313f77af/src/help.h#L1007 which seems to be autogenerated.

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -3589,8 +3589,8 @@
       },
       {
         "command": "SETID",
-        "name": ["key", "id-or-$"],
-        "type": ["key", "string"],
+        "name": ["key", "groupname", "id-or-$"],
+        "type": ["key", "string", "string"],
         "optional": true
       },
       {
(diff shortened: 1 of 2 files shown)


antirez:
Good changes @zetaben! Thanks, merged.

PR #1029, opened 2018-11-07
XREADGROUP ignores BLOCK and NOACK

gkorland:
XREADGROUP ignores BLOCK and NOACK when specific ID is used


commands/xreadgroup.md
```diff
@@ -54,6 +54,7 @@ be one of the following two:
 
 * The special `>` ID, which means that the consumer want to receive only messages that were *never delivered to any other consumer*. It just means, give me new messages.
 * Any other ID, that is, 0 or any other valid ID or incomplete ID (just the millisecond time part), will have the effect of returning entries that are pending for the consumer sending the command. So basically if the ID is not `>`, then the command will just let the client access its pending entries: delivered to it, but not yet acknowledged.
+Notice: in this case both BLOCK and NOACK are ignored
```
itamarhaber: Please:
1. Attach the line to the bullet as it refers only to it.
2. The docs use the "Note that" instead of "Notice:" standard.

itamarhaber:
Replaced by #1185

PR #1031, opened 2018-11-13
Add example for SET command with expiration

romuald:
I had trouble finding the correct way of setting a key with an expiration date with the current documentation (maybe because of #928 / #973)

So I though adding an example would be nice for new users


kishore25kumar:
👍

itamarhaber:
Thanks @romuald - good idea, merging.

PR #1045, opened 2018-11-25
wiredis added as c++ client

winnieelte:


itamarhaber:
Hello @winnieelte - thanks, merging after a rebase. I also took the liberty of removing your GitHub handle from the "authors" prop as it accepts only Twitter. Please feel free to submit a new PR to correct this if you should so wish.

winnieelte:
@itamarhaber Thanks! No problem at all, I definitely don't have Twitter :)

PR #1058, opened 2018-12-22
Fix three typos in streams-intro.md

ZYunH:
attack -> attach
conusmer-name -> consumer-name
entires -> entries


itamarhaber:
Thanks you and my apologies for the belated reply. I believe all of these have been addressed by previous pull requests so closing this one.

PR #1060, opened 2018-12-25
Add Go's RedisPipe

funny-falcon:
Add link to RedisPipe - Go's client with implicit pipelining.
Note: I've copied `"recommended": true` mark since there no rules about it, and I personally will recommend it :-)


itamarhaber:
Thanks.

> I've copied "recommended": true mark since there no rules about it

I removed via in https://github.com/antirez/redis-doc/pull/1060/commits/dca3a7dadc1477f8f96814f3b6b6974e62e0ff5b

Thanks - merging.

PR #1067, opened 2019-01-09
Adds Streams to data types intro

itamarhaber:
Added a reference to the Streams Intro, and an exception to the rule of empty aggregate data types.


charsyam:
Thanks @itamarhaber

PR #1068, opened 2019-01-09
update msetnx example to make it obvious that no partial updates are made

alexanderbird:
In the example, both the first (successful) and second (unsuccessful) command attempt to set key `key2` to value `there`. It's not obvious from the example that it's the first "there" that's responsible for `key2` having a value of "there", and not the second "there".

This minor change makes it obvious where the value "there" comes from.


itamarhaber:
Thanks - merging.

PR #1085, opened 2019-02-18
Fix BZPOP[MIN/MAX] reply order

michael-grunder:
Elements will be delivered member, score, member, score


itamarhaber:
Thanks - sorry for the late reply, this is a duplicate of #997 - closing.

Issue #964, opened 2018-08-28
Broken C# link


CrazyHackGUT:
Link `C#` on [Clients page](https://redis.io/clients) goes to `C`, not `C#`.

itamarhaber:
Hello @CrazyHackGUT 

Thanks for reporting - closing this as it is being tracked at https://github.com/antirez/redis-io/issues/182

PR #1087, opened 2019-02-21
C# -> CSharp

imba-tjd:
Fixes https://github.com/antirez/redis-doc/issues/964 and https://github.com/antirez/redis-io/issues/134.
I choose this way because I don't know ruby. If you think this not appropriate, it's OK to just close.

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -613,7 +613,7 @@
 
   {
     "name": "ServiceStack.Redis",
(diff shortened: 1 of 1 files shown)


itamarhaber:
Thanks, but the problem is with the code that runs the website - closing and tracking at https://github.com/antirez/redis-io/issues/182

PR #1088, opened 2019-02-25
form -> from typo fix.

maptile:
I think it should be "popped from the list"


itamarhaber:
Thanks - this is a duplicate of #1076 - closing.

PR #1106, opened 2019-04-28
Remove “abc” in the example of transactions

boyang9602:
I think this "abc" is written by mistake?


itamarhaber:
After further consideration, I've made a small tweak to make it clear.

Issue #1107, opened 2019-05-05
Typo error "losign" chould be changed to "losing". 


jacobliuat:
https://redis.io/topics/replication

 However with WAIT the probability of losign a write after a failure event is greatly reduced to certain hard to trigger failure modes.

==>  Typo error "losign" chould be changed to "losing". 

 However with WAIT the probability of losing a write after a failure event is greatly reduced to certain hard to trigger failure modes.

itamarhaber:
Hello @jacobliuat 

Thanks for spotting the typo. You can submit a PR that fixes this issue if you want.

PansonPanson:
fixed it

itamarhaber:
Fixed by #1158 (blame: @charsyam ;P)

PR #1108, opened 2019-05-09
fix issue #1107

PansonPanson:
fix ： https://github.com/antirez/redis-doc/issues/1107


itamarhaber:
It appears that #1158 is a duplicate of this that was merged before this was reviewed. Thanks for both the PR and the issue - please keep them coming :)

Closing.

PR #1109, opened 2019-05-15
Update example of bit level operations

cherrydev:
Existing example used sex of a user as being a binary male/female option. I believe this example has become outdated since it was written and can be better explained using a neutral example that is undeniably binary, such as a subscription to a mailing list.


itamarhaber:
Thanks @cherrydev - merging.

PR #1110, opened 2019-05-17
Fix XPENDING final argument descriptions and typo

vcing:
Fix XPENDING final argument descriptions in topics/streams-intro.md
The optional final argument in XPENDING is consumer-name


itamarhaber:
Thanks @vcing - both fixes have already been addressed apparently. Closing.

PR #1115, opened 2019-05-23
Add FlyRedis

icerlion:
FlyRedis is a C++ RedisClient, based on boost.asio, you can compile it on linux and windows. Easy to use.


itamarhaber:
Thanks - merging.

PR #1117, opened 2019-05-24
Fix typo for the size of an integer in the Bitfield documentation

sgama:
This typo had me very confused until I realized it was a typo and not the actual command.


itamarhaber:
Thanks for the resolution @sgama - a prior PR had already fixed that. Closing.

PR #1125, opened 2019-06-16
replace RedisModule_ThreadSafeCallStop with RedisModule_FreeThreadSaf…

gkorland:
…eContext


itamarhaber:
@gkorland - this needs to be made against the source repo, this file is autogenerated from the inline comments.

PR #1126, opened 2019-06-16
Document optionality of PFADD, EVAL, EVALSHA args

nbouscal:


zuiderkwast:
From what I can tell, skimming through the rendering of this in https://github.com/redis/redis-io/blob/master/lib/reference.rb, this would result in changing

    EVAL script numkeys key [key ...] arg [arg ...]

into

    EVAL script numkeys [key [key ...]] [arg [arg ...]]

If this is the case, then I totally approve. It's possible to call EVAL with 0 keys.

zuiderkwast:
Thanks @nbouscal and sorry for the delay!

PR #1134, opened 2019-07-03
Fix bzpopmin example

bontequero:


itamarhaber:
Thanks @bontequero - this was already fixed in #997. Closing.

PR #1136, opened 2019-07-07
Formatted ruby examples

dukaev:
👋 Hello
With this version of code it is easier to read
I used rubocop for analyze.


zuiderkwast (approved):
It does align the code with https://rubystyle.guide/ so I think we should merge this.

PR #1140, opened 2019-07-14
add soce-redis

wusongwei:
add soce-redis


itamarhaber:
Thanks @wusongwei - I had removed the 'author' as it should refer to Twitter (not GitHub). Merging.

PR #1143, opened 2019-07-22
Fix typo in zcount command

guyoun:


itamarhaber:
Hello @guyoun 

I don't believe this is a typo, but rather an example of an exclusive range.

guyoun:
@itamarhaber You are right. I closed this pull request.

Issue #981, opened 2018-09-24
Broken "Understanding Virtual Memory" Link in FAQ


hayashier:
In the FAQ( https://redis.io/topics/faq ), I clicked "Understanding Virtual Memory" and moved to the link( https://www.redhat.com/magazine/001nov04/features/vm/ ).
But transitioned page is 404 error now.

itamarhaber:
@hayashier - thanks :)

PR #1149, opened 2019-07-29
Fix a broken link

robin850:
Hello,

This pull request simply fixes a broken link in the FAQ.

The article is no longer available so let's fix the link to a reworked version of it.

Also, the 1 and 2 values are no longer swapped in this article but let's keep a link to the `proc` man page as it can be useful and simpler than the whole article (here's the [original article](http://web.archive.org/web/20120608085846/http://www.redhat.com/magazine/001nov04/features/vm/) that wrongly explains the 1 and 2 values).

Have a nice day.


itamarhaber:
Fixes #981

robin850:
(I guess you need to close it manually or use the "Fixes #n" syntax in a commit to close the mentionned issue ; but thanks ! :-))

Issue #1154, opened 2019-08-07
update lpushx rpushx doc for variable elements support 


patpatbear:
`LPUSHX` and `RPUSHX` accepts varaible number of elements since Redis 4.0.0-RC1,
but it seems like that this feature was never documented, hence the redis-cli hint
also does not imply `LPUSHX` and `RPUSHX`  support varaible elements.

release note for 4.0.0:

```
================================================================================
Redis 4.0.0-RC1   Released Fri Dec 2 10:40:01 CEST 2016
================================================================================
...
* RPUSHX and LPUSHX now accept a variable number of elements.
...
```

itamarhaber:
Thanks @patpatbear - #1155 has been merged so closing.

PR #1155, opened 2019-08-07
update LPUSHX and RPUSHX for variable elements support.

patpatbear:
This PR corresponds to issue #1154.


commands.json
```diff
@@ -1517,7 +1517,7 @@
     "group": "list"
   },
   "LPUSHX": {
-    "summary": "Prepend a value to a list, only if the list exists",
+    "summary": "Prepend one or multiple values to a list, only if the list exists",
     "complexity": "O(1)",
```
itamarhaber: Should be "O(N) where N is the number of elements to be added by the operation"

commands.json
```diff
@@ -2133,7 +2134,7 @@
     "group": "list"
   },
   "RPUSHX": {
-    "summary": "Append a value to a list, only if the list exists",
+    "summary": "Append one or multiple values to a list, only if the list exists",
     "complexity": "O(1)",
```
itamarhaber: Should be "O(N) where N is the number of elements to be added by the operation"

commands.json
patpatbear: updated complexity and history in commit 9a6afdd6f9

commands.json
patpatbear: updated complexity and history in commit 9a6afdd

itamarhaber:
Thanks - merging after rebase and a little tweak.

itamarhaber:
Fixes #1154

PR #1160, opened 2019-08-29
Fix typo `123` -> `1234` in streams-intro

BrusiRoy:


charsyam:
Good Catch!!! Thanks.

PR #1161, opened 2019-09-04
Update SPOP: the return value should be exactly same with SRANDMEMBER

laixintao:


laixintao:
Thanks for letting me know.

BTW I'm working on a Redis command line client, I made redis-doc as my git submodule so that user can check docs on command line like this

image

PR #1165, opened 2019-10-04
Fix XADD field value name type

d024441:


d024441:
Hello @itamarhaber 

exactly, same syntax is used in HMSET:
```
{
  "name": ["field", "value"],
  "type": ["string", "string"],
   "multiple": true
} 
```

but in XADD value and string are in the 'wrong' array:

```
{
  "name": ["field", "string"],
  "type": ["value", "string"],
  "multiple": true
}
```

which leads to the following rendered help text in [XADD](https://redis.io/commands/xadd):
XADD key ID field string [field string ...] 
which should read
XADD key ID field value [field value ...] 

Please be also noticed that the 'Capped streams' option of XADD is completely missing in command.json for XADD.

Best regards

itamarhaber:
Ok, I see now. Good.

> Please be also noticed that the 'Capped streams' option of XADD is completely missing in command.json for XADD.

Thanks for noticing, tracking this at https://github.com/antirez/redis-io/issues/167. The problem is that the syntax makes it impossible to express it (i.e. nested optionals).

PR #1167, opened 2019-10-10
Fix typo in BZPOPMIN example

pocams:
The BZPOPMIN example creates a set named "zset1" but shows the returned value from BZPOPMIN as "zet1".


itamarhaber:
Thanks @pocams - merging.

PR #1188, opened 2019-10-19
Adding Hashes to ACL Topic.

IAmATeaPot418:
Adding documentation for adding hashes to the ACL topic.


itamarhaber:
@IAmATeaPot418 - thanks for this, looks good and merging.

PR #1206, opened 2019-11-06
update commands.json: unique index parameters SWAPDB

d024441:
unique index parameters index1, index2 SWAPDB like member1, member2 GEODIST


d024441:
Hello @itamarhaber 
thanks for the swift response and merging the request.

PR #1211, opened 2019-11-18
Update docs to include MKSTREAM usage

lowewenzel:
`MKSTREAM` was added to Streams, but there is no documentation on it yet.

See #1007

### DIFF
--- a/commands/xgroup.md
+++ b/commands/xgroup.md
@@ -25,6 +25,14 @@ consumer group already exists, the command returns a `-BUSYGROUP` error.
 Otherwise the operation is performed and OK is returned. There are no hard
 limits to the number of consumer groups you can associate to a given stream.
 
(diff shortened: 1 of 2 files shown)


lowewenzel:
Thank you @itamarhaber. I opened another PR in the main repo: https://github.com/antirez/redis/pull/6602

Thank you again for approving!

itamarhaber:
WRT to the other PR - it isn't needed as that file is generated from this repo. That said, the issue is that `MKSTREAM` is an optional subcommand of only one of the options. The current syntax rules do not allow expressing this so I don't see how it can be accurately expressed (see https://github.com/antirez/redis-io/issues/159).

lowewenzel:
Ah, I see. Thank you for letting me know! I closed the other PR since it's the autogenerated file.

PR #1213, opened 2019-11-24
Update clients.json

d024441:
Go client based on RESP3.


d024441:
@itamarhaber Thanks for accepting!

PR #1218, opened 2019-12-08
update commands.json ZINCRBY: change increment type to double

d024441:
ZINCRBY accepts double for increment parameter - current type is integer


d024441:
Hello @itamarhaber,
yes, sure, you are right - my bad only taking RSP3 into consideration. 
Thanks and best regards

PR #782, opened 2016-12-18
Add concurrent note for Rate limiter Pattern

atealxt:
PR for [issue 777](https://github.com/antirez/redis-doc/issues/777).

PR #1219, opened 2019-12-08
Rate Limiter Pattern Improvements (new)

raginpirate:
# Why
The redis-docs are supposed to be designed for beginners learning about how to use redis. However, the incr docs can lead these new developers into implementing non-thread-safe rate limiters in their modern web applications.

Here is an explanation of why the current code fails, and how the new code improves in concurrency.
concurrency_explanation

# What
Made very minor changes to two of the three provided examples to support concurrent rate limiting.
Fixed minor grammar/spelling mistakes throughout the section.
Elaborated on **why** the examples are written the way they are (concurrency, exceptional behavior outcomes, etc.) and what the short-comings of each example are.

I believe these changes are simpler, easier to follow, and more effective than the previous PRs opened to address this problem. (#893 and #782).
Please consider my changes, and let me know if anything I wrote needs elaboration.

### DIFF
--- a/commands/incr.md
+++ b/commands/incr.md
@@ -120,9 +120,8 @@ value greater than 10, otherwise it will expire and start again from 0.
 If for some reason the client performs the `INCR` command but does not perform
 the `EXPIRE` the key will be leaked until we'll see the same IP address again.
 
-One solution to this is using a default TTL for your redis keys. Another 
-solution is converting the `INCR` with optional `EXPIRE` into a Lua script that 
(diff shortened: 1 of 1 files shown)


commands/incr.md
```diff
@@ -154,6 +154,8 @@ ELSE
 END
 ```
 
+**The above code is not safe for concurrent usage**
```
ZhangChaoWN, line 157: this problem can be fixed too, just add length checking after calling the RPUSH and before PERFORM_API_CALL()

PR #1222, opened 2019-12-14
update commands.json SETBIT: change value type to integer

d024441:


d024441:
@itamarhaber thanks for accepting!

PR #1223, opened 2019-12-14
Moves some commands to a new 'bit' category

itamarhaber:
This moves the 6 bit-related commands from the 'string' to a new
category. The motivation is giving more exposure for these lesser-
known capabilities.

@antirez - your thoughts?

TODO:
  [ ] Add Bit group in redis-io/lib/reference.rb

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -57,7 +57,7 @@
       }
     ],
     "since": "2.6.0",
-    "group": "bit"
+    "group": "bitmap"
   },
   "BITFIELD": {
     "summary": "Perform arbitrary bitfield integer operations on strings",
@@ -119,7 +119,7 @@
       }
     ],
     "since": "3.2.0",
-    "group": "bit"
+    "group": "bitmap"
   },
   "BITOP": {
     "summary": "Perform bitwise operations between strings",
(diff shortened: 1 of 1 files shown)


itamarhaber:
@antirez ping

commands.json
```diff
@@ -57,7 +57,7 @@
       }
     ],
     "since": "2.6.0",
-    "group": "string"
+    "group": "bitmap"
```
zuiderkwast, line 60: I suggest changing to "bitstring" because the word bitmap often refers to images. But I guess that's not even visible externally.

zuiderkwast (approved):
I think this is a good idea. Merge it! :+1:

Rose2161 (approved):

commands.json
madolson: They're called bit arrays or bitmaps everywhere else in our documentation, https://redis.io/topics/data-types-intro, seems like we should stick with bitmaps.

madolson (approved):

zuiderkwast:
@madolson Do you want to merge this one and at the same time this one supporting it: https://github.com/redis/redis-io/pull/192? I don't have commit rights in redis-io.

madolson:
@zuiderkwast Sure, seems.

PR #1232, opened 2019-12-29
feat: json-schema commands format

mmkal:
@itamarhaber this is an idea following the discussion over the DSL in https://github.com/antirez/redis-doc/pull/1231

This PR consists of a small js script which parses `commands.json` to produce a json file in a slightly different format. Each command has the same root-level properties, except for `arguments`, which is now in the format:

```typescript
{
  name: string;
  optional: boolean;
  schema: JSONSchema7;
}
```

This allows redis to avoid re-inventing the wheel for any of the undifferentiated schema-description parts of the documentation, instead relying on [json-schema](https://json-schema.org/), which is well established, widely used, unambiguous, flexible enough to cover edge-cases, and a good balance of human- and machine-readable.

It's worth taking a look at the [generated json-schema file](https://github.com/mmkal/redis-doc/blob/1b9ed2343df263405ec69d53ffbeec0e8d355d7d/commands.json-schema.json), since it's large so github suppresses the diff. 

Note:

Each argument is represented by JSON-schema, but conceptually the schemas are "flattened" when passed as CLI arguments. For example, [`DEL`](https://github.com/mmkal/redis-doc/blob/1b9ed2343df263405ec69d53ffbeec0e8d355d7d/commands.json-schema.json#L1086-L1105). This has one argument, `key` which is of type `array`, meaning it would be valid for inputs like `['foo', 'bar']`, corresponding to `redis del foo bar`. So to go from arguments valid per the schema to a CLI call, clients would have to do something like `redis ${command} ${flatten(arguments).join(' ')}`.

I also wrote a simple function that parses the `.md` documentation for each command to extract the return type and convert to json-schema, so that type can be kept alongside the arguments. Going forward, this json file could become the source of truth for return types, so that this repo could document not just return types like `array-reply`, but specify the type of the items in the array too (e.g. `{ "type": "array", "items": { "type": "string" } }`).

This will make client generation much easier, because existing json-schema tooling can be used rather than each client library having to write a custom parser which reverse-engineers the custom DSL currently in `command.json`. (The [library I maintain](https://github.com/mmkal/handy-redis) does exactly this, and has a lot of hacks and pieces of difficult-to-maintain code. I have held off adding it to client.json so far because of this).

My thought was that if this accepted, the generated `commands.json-schema.json` acts as a starting point, and would eventually take the place of `commands.json`, and be manually edited, as the source of truth, going forwards. The js file shouldn't need to hang around, it's just for generating the "starter" json-schema file. This will allow taking advantage of json-schema's API to go further with type specificity. For example, the problem highlighted in https://github.com/antirez/redis-doc/pull/1231 would go away, because `SET` could become (using yaml for conciseness):

```yaml
summary: Set the string value of a key
complexity: O(1)
arguments:
- name: key
  schema:
    type: string
- name: value
  schema:
    type: string
- name: expiration
  optional: true
  schema:
    type: array
    items:
    - type: enum
      enum: [EX, PX]
    - type: integer
- name: condition
  optional: true
  schema:
    type: string
    enum:
    - NX
    - XX
since: 1.0.0
group: string
return:
  anyOf:
  - type: string
  - type: 'null'
```

### DIFF
--- a/commands.json-schema.json
+++ b/commands.json-schema.json
@@ -1,0 +1,6172 @@
+{
+  "APPEND": {
+    "summary": "Append a value to a key",
+    "complexity": "O(1). The amortized time complexity is O(1) assuming the appended value is small and the already present value is of any size, since the dynamic string library used by Redis will double the free space available on every reallocation.",
+    "arguments": [
+      {
+        "name": "key",
+        "schema": {
+          "type": "string"
+        }
+      },
+      {
+        "name": "value",
+        "schema": {
+          "type": "string"
+        }
+      }
+    ],
+    "since": "2.0.0",
+    "group": "string",
+    "return": {
+      "type": "integer"
+    }
+  },
+  "AUTH": {
+    "summary": "Authenticate to the server",
+    "arguments": [
+      {
+        "name": "password",
+        "schema": {
+          "type": "string"
+        }
+      }
+    ],
+    "since": "1.0.0",
+    "group": "connection",
+    "return": {
+      "type": "string"
+    }
+  },
+  "BGREWRITEAOF": {
+    "summary": "Asynchronously rewrite the append-only file",
+    "since": "1.0.0",
+    "group": "server",
+    "arguments": [],
+    "return": {
+      "type": "string"
+    }
+  },
+  "BGSAVE": {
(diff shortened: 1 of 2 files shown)


mmkal:
@itamarhaber - any thoughts on this?

PR #1247, opened 2020-01-28
add reventis entry to modules.json file

starkdg:
New module to index events.

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -306,6 +306,6 @@
 		"authors": [
 			"starkdg"
 		],
-		"stars":1
+		"stars": 2
 	}
 ]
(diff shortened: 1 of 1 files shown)


itamarhaber:
Thanks @starkdg - merging and looks promising :)

starkdg:
You're very welcome @itamarhaber . It's in a fairly stable state now.  I figured it was time to get it out there and see what else can be done with it.  

With some modifications, the code should be able to handle more than just three dimenions of data.  Also, a flag could be introduced to handle categories of events. 

Any comments and suggestions for improvement would be greatly appreciated.

PR #1255, opened 2020-02-18
Listing other not logged commands by MONITOR

atealxt:
Other than admin commands ([1][2]), there are several not logged commands. List them on the same section could be helpful I think.

[1] https://github.com/antirez/redis/commit/d6410ed19a4930895f2591eed224d5ec5449393a
[2] https://github.com/antirez/redis-doc/commit/80129ef3a0fc0e48aeda73f9c9b790cf7bdb2636

### DIFF
--- a/commands/monitor.md
+++ b/commands/monitor.md
@@ -45,6 +45,13 @@ Manually issue the `QUIT` command to stop a `MONITOR` stream running via
 For security concerns, certain special administration commands like `CONFIG`
 are not logged into the `MONITOR` output.
 
+Other not logged commands: 
+
+ * `AUTH` (start from Redis 6.0)
+ * `EXEC`
+ * `HELLO`
+ * `QUIT`
+
 ## Cost of running `MONITOR`
 
 Because `MONITOR` streams back **all** commands, its use comes at a cost.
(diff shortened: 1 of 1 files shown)


commands/monitor.md
```diff
@@ -42,17 +42,17 @@ Manually issue the `QUIT` command to stop a `MONITOR` stream running via
...
 
-For security concerns, certain special administration commands like `CONFIG`
-are not logged into the `MONITOR` output.
+Because of security concerns, all administrative commands are not logged
+by `MONITOR`'s output.
 
-Other not logged commands: 
+Furthermore, the following commands are also not logged:
 
- * `AUTH` (start from Redis 6.0)
```
atealxt: @itamarhaber 
I reviewed the code seems logging `AUTH` is being removed from Redis 6.0, which version is in beta currently.
Right now 5.0 (and before) are more common versions, not sure if it will make confusion without indicate version.

itamarhaber: Are you saying that people will not scroll down all the way and read the "History" section?

atealxt: Oh, there it is, that's OK.

Remember that `EXEC` and `QUIT` are not excluded from 6.0, but from previous versions already.

itamarhaber: Right - fixed via 0bc9e22

PR #1261, opened 2020-03-02
Fix typos in Sorted set section

samueldg:
* leader board --> leaderboard
* an user --> a user


samueldg:
Just noticed "leader board" is used elsewhere -- let me know which is preferred and I can revert this or update elsewhere.

PR #1263, opened 2020-03-07
latency reset's args cloud be multiple.

laixintao:
LATENCY HELP commands says:
```
127.0.0.1:6379> latency help
6) RESET   [event ...] -- Resets latency data of one or more event classes.
7)                        (default: reset all data for all event classes)
```


laixintao:
Could you please take a look at this? @itamarhaber

PR #1265, opened 2020-03-09
Update documentation:: smembers command  

Azhovan:
Indicate that set will be sorted automatically if and only if all members are integers


madolson:
Yeah, also just generally seems like a bad practice to say that this is sorted. Although it may be at points in time, the end users shouldn't rely on that.

zuiderkwast:
Agree with previous reviewers: The order is undocumented on purpose and should remain undocumented.

PR #1270, opened 2020-03-20
ACL Command Category from @READONLY to @READ

hwware:
there is no readonly command category but read


itamarhaber:
Thanks @hwware - merging.

PR #1271, opened 2020-03-22
add AnotherRedisDesktopManager to tools.json

qishibo:
repo: https://github.com/qishibo/AnotherRedisDesktopManager
desc: 🚀🚀🚀A faster, better and more stable redis desktop manager. What's more, it won't crash when loading a large number of keys.


qishibo:
Thx😀ོ

PR #1273, opened 2020-03-26
Update active status of predis in clients.json

ronkrl:
Removing "Active" status for predis, since it has no commits in almost 3 years

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -437,7 +437,6 @@
     "description": "Mature and supported",
     "authors": ["JoL1hAHN"],
     "recommended": true,
-    "active": true
   },
 
   {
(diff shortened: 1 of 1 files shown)


itamarhaber:
Hello @ronkrl 

Thanks for the update, I agree that the flag can be removed in this case. Please fix the PR so CI doesn't fail (there's an extra comma there I believe).

ronkrl:
> Hello @ronkrl
> 
> Thanks for the update, I agree that the flag can be removed in this case. Please fix the PR so CI doesn't fail (there's an extra comma there I believe).

Yep, thanks for catching; thought I had gotten that. Updated now.

PR #1284, opened 2020-04-30
XINFO STREAM FULL docs

guybe7:


guybe7:
@antirez i didn't modify `commands.json` because nested optionals are not (yet?) supported AFAIK...
i took a look at `XGROUP CREATE` and saw that the optional `MKSTREAM` is not documented in `commands.json` but only in `xgroup.md` so i followed the same path

commands/xinfo.md
```diff
@@ -36,6 +36,73 @@ not be the same as the last entry ID in case some entry was deleted. Finally
...
+                    2) (integer) 1588152520299
+                    3) (integer) 1
+```
+
+The reported information contains all of the fields reported by the simple
+form of `XINFO STREAM`, with some additional information:
+1. Stream entries are returned, including fields and values.
+2. Groups, consumers and PELs are returned.
+The `COUNT` option is used to limit the amount of stream/PEL entries that are
+returned (The first <count> entries are returned). The default `COUNT` is 10 and
```
itamarhaber, line 102:
```suggestion
returned (The first `<count>` entries are returned). The default `COUNT` is 10 and
```

itamarhaber:
@guybe7 

Can you please add a `@history` telling about this change to the xinfo.md page?

guybe7:
@itamarhaber done

itamarhaber:
Thanks @guybe7 - merging.

PR #1291, opened 2020-05-01
Adds ACL categories to each command

itamarhaber:


oranagra:
@itamarhaber i took a quick look at the changes, the diff is very hard to review, some commands seem to have moved to a different place in the file, makes it hard to review, but also introduces unnecessary conflicts for this and other PRs.

is it just about specifying the ACL categories for each command? or does this PR include other changes?
about the ACL categories, i'm not certain we wanna document them, same as command flags (ok-loading and such), they might change and we'll need to keep updating the docs.
instead there's a simple command that let's the people get the categories from a the actual server.

zuiderkwast:
Would we instead want to list the categories automatically by querying the local Redis instance? That would eliminate the need to update the docs when ACL categories change.

oranagra:
I don't think that's needed. I think what we have now is enough. If someone wants the exact details, depending on the version he uses, he should get that from Redis.. 
Like Salvatore said in the (now deleted) docs: listing them would be boring...

itamarhaber:
Closing as this was implemented elsewhere.

PR #1297, opened 2020-05-05
bugfix: <count> is a html tag, should be `<count>`

laixintao:
Hi @antirez 

For most markdown renders `<count>` is traded like a HTML tag, thus can't be rendered. What's worse is that some renders think it's not closed so they just yield errors.


laixintao:
That's what it looks like on GitHub (Note `<count>` is missing):

image

That's what it looks on https://redis.io/commands/xinfo :

image

laixintao:
cc @guybe7 @itamarhaber

guybe7:
@laixintao indeed. thanks!

PR #1308, opened 2020-05-18
Remove misleading statement about security

banker:
This sentence makes less sense with all of the new security features in Redis 6.0.


antirez:
Thanks

PR #1311, opened 2020-05-20
fix typo in CLIENT REDIR command

hwware:


itamarhaber:
Thanks @hwware - merging.

PR #1317, opened 2020-05-27
fix typo in module api ref doc

hwware:


itamarhaber:
Dear @hwware - approved. However.

This file is autogenerated from the inline comments in redis/src/module.c ... please move this upstream.

hwware:
@itamarhaber Hello Itamar, thank you! I will submit the changes in upstream then

PR #1321, opened 2020-06-01
Adding river to C++ and Python Redis clients

pbotros:
River is a library built on top of Redis Streams written in C++ and with
Python bindings. It's not a client per-say but rather a useful framework
that people can use for a more tailored interface to Redis Streams for
both streaming and persistence.


itamarhaber:
Helllo @pbotros - this looks quite useful, but perhaps river would feel more at home in the tools.json file?

pbotros:
@itamarhaber totally right - missed the tools at the bottom of the page. Updated now.

itamarhaber:
Merging - thanks for making it :)

PR #1337, opened 2020-06-24
update tracking doc to make it consistant with implementation

hwware:
This PR fix two things:
1. use the correct command for enabling client tracking for next keys in opt in mode
2. since noreply was not supported in current version of Redis(not sure whether we need to support it later or not), but in my point of view we should not include this in this documentation in order to avoid confusion for users..


hwware:
@itamarhaber thank you Itamar :)

PR #1349, opened 2020-07-15
Added Redis-COM-client to redis.io/clients

ricosuave73:
Added the _Redis COM client_ wrapper for COM environments like classic ASP.

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -1809,5 +1809,16 @@
     "authors": ["kristoff-it"],
     "recommended": true,
     "active": true
+  },
+
+  {
+    "name": "Redis COM client",
+    "language": "ActiveX/COM+",
+    "url": "https://gitlab.com/erik4/redis-com-client",
+    "repository": "https://gitlab.com/erik4/redis-com-client",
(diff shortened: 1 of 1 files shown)


clients.json
```diff
@@ -1809,5 +1809,16 @@
...
     "active": true
+  },
+
+  {
+    "name": "Redis COM client",
+    "language": "ActiveX/COM+",
+    "url": "https://gitlab.com/erik4/redis-com-client",
+    "repository": "https://gitlab.com/erik4/redis-com-client",
+    "description": "A COM wrapper for StackExchange.Redis that allows using Redis from a COM environment like Classic ASP (ASP 3.0) using vbscript, jscript or any other COM capable language.",
+    "authors": ["Erik Oosterwaal"],
```
itamarhaber, line 1820: This should be your Twitter handle... seeing as you're not using it, just leave it blank like so:
```suggestion
    "authors": [],
```

ricosuave73: All clear, I've removed the value.

PR #1352, opened 2020-07-18
Adds monix redis repo

paualarco:


paualarco:
> Hello @paualarco - can you provide a ref for the updated name?
`monix-redis` is a sub-module of the `monix-connect` project, but not a repo on its own. 
Sorry, I was unsure if the name had to strictly be the same as the repo name. I have just updated it accordingly.
https://github.com/monix/monix-connect

itamarhaber:
Thanks - makes sense. Merging.

PR #1354, opened 2020-07-21
added a new redis client for node

razaellahi:


clients.json
```diff
@@ -775,7 +775,17 @@
...
     "active": true
   },
-
+  
+  {
+    "name": "xredis",
+    "language": "Node.js",
+    "repository": "https://github.com/razaellahi/xredis",
+    "description": "Redis client with redis ACL features",
+    "authors": ["RazaEllai"],
```
itamarhaber, line 784: This is supposed to be the author's Twitter account - if you don't have one just leave it blank like so:
```suggestion
    "authors": [],
```

PR #1355, opened 2020-07-21
fix expire command in keyspace event doc

hwware:
The expire command in keyspace notification has two potential issues: 1. it missed the other related commands,like PEXPIRE. 2. The event generated when giving past time is wrong, should be del rather than expired.


itamarhaber:
Good addition - ty - please consider my edits below.

```suggestion
* `EXPIRE` and all its variants (`PEXPIRE`, `EXPIREAT`, `PEXPIREAT`) generate an `expire` event when called with a positive timeout (or a future timestamp). Note that when these commands are called with a negative timeout value or timestamp in the past, the key is deleted and only a `del` event is generated instead.
```

hwware:
@itamarhaber thanks Itamar, I modified this pr based on your comments, thanks :)

PR #1360, opened 2020-07-30
adds Rediska to clients.json

pepelev:
Adding a new Rediska client to clients.json list.

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -727,7 +727,7 @@
     "language": "C#",
     "url": "https://github.com/pepelev/Rediska",
     "description": "Rediska is a Redis client for .NET with a focus on flexibility and extensibility.",
-    "authors": ["pepelev"],
+    "authors": [],
     "recommended": false,
     "active": true
   },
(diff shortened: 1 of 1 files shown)


clients.json
```diff
@@ -722,6 +722,16 @@
     "authors": ["chakrit"]
   },
 
+  {
+    "name": "Rediska",
+    "language": "C#",
+    "url": "https://github.com/pepelev/Rediska",
+    "description": "Rediska is a Redis client for .NET with a focus on flexibility and extensibility.",
+    "authors": ["pepelev"],
```
itamarhaber, line 730: This should be your Twitter handle, or just empty if you don't use that network,.

```suggestion
    "authors": [],
```

pepelev: It was not clear for me. So far i don't use Twitter.

itamarhaber: Sure - I'm aware there something left to be desired with the current process :) Thinking about improving this.

PR #1361, opened 2020-07-31
Add documentation for zmscore command

TysonAndre:
Docs PR for https://github.com/redis/redis/pull/7593 , which is still in review

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -3879,6 +3879,23 @@
     "since": "1.2.0",
     "group": "sorted_set"
   },
+  "ZMSCORE": {
+    "summary": "Get the score associated with the given members in a sorted set",
+    "complexity": "O(N) where N is the number of fields being requested.",
+    "arguments": [
+      {
+        "name": "key",
+        "type": "key"
+      },
+      {
+        "name": "member",
+        "type": "string",
+        "multiple": true
+      }
+    ],
+    "since": "6.2.0",
+    "group": "sorted_set"
+  },
   "ZUNIONSTORE": {
     "summary": "Add multiple sorted sets and store the resulting sorted set in a new key",
     "complexity": "O(N)+O(M log(M)) with N being the sum of the sizes of the input sorted sets, and M being the number of elements in the resulting sorted set.",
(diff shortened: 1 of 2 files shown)


commands.json
```diff
@@ -3879,6 +3879,23 @@
     "since": "1.2.0",
     "group": "sorted_set"
   },
+  "ZMSCORE": {
+    "summary": "Get the score associated with the given members in a sorted set",
+    "complexity": "O(1)",
```
oranagra, line 3884:
```suggestion
    "complexity": "O(N) where N is the number of fields being requested.",
```

commands.json
```diff
@@ -3879,6 +3879,23 @@
...
+        "name": "key",
+        "type": "key"
+      },
+      {
+        "name": "member",
+        "type": "string",
+        "multiple": true
+      }
+    ],
+    "since": "1.2.0",
```
oranagra, line 3896:
```suggestion
    "since": "6.2.0",
```

commands/zmscore.md
```diff
@@ -0,0 +1,16 @@
+Returns the scores associated with the specified `members` in the sorted set stored at `key`.
+
+If `key` does not exist, `nil` is returned.
```
oranagra, line 3: For every `member` that does not exist in the hash, a `nil` value is returned.

commands.json
TysonAndre: Done

commands.json
TysonAndre: Done

commands/zmscore.md
```diff
@@ -0,0 +1,16 @@
+Returns the scores associated with the specified `members` in the sorted set stored at `key`.
+
+For every `member` that does not exist in the hash, a `nil` value is returned.
```
itamarhaber, line 3:
```suggestion
For every `member` that does not exist in the sorted set, a `nil` value is returned.
```

oranagra:
@itamarhaber since the command is already in unstable, i think we can merge this one (whenever you're ready).

i suppose that in theory it's good to have docs visible early (client library maintainers may look at them).

TysonAndre:
This should be ready.

itamarhaber:
@oranagra I'm waiting for the plate to stop spinning wrt the version this will be picked from unstable - currently this PR says 6.2, but it may end up in 6.0.7 so I want to get that right.

PR #1367, opened 2020-08-12
Adds a note about cluster keyspace notifications

itamarhaber:
And some minor edits


ushachar:
@itamarhaber shouldn't this close https://github.com/redis/redis/issues/2541 as well?

itamarhaber:
@ushachar definitely looks like it can be put to rest - thanks for the bump :)

PR #1380, opened 2020-08-31
@haadj explicit the OOM abbreviation

haadj:
The OOM (Out Of Memory) meaning is not explained elsewhere in the page


itamarhaber (approved):
Thanks @haadj - merging after edits.

PR #1393, opened 2020-09-18
Update clients.json for Swift language

Mordil:
Also marks any Swift client that hasn't received an update in over two years as inactive.

The justification for **RediStack** being the recommended client library is due to being accepted by the [Server-side Swift Working Group](https://github.com/swift-server/sswg/blob/main/proposals/0004-nio-redis.md)


itamarhaber (approved):

itamarhaber:
Thanks @Mordil - merging.

PR #1397, opened 2020-09-23
Documentation update for new ZADD GT and LT options

alexronke-channeladvisor:
Per the pull request in the main redis repository:
https://github.com/redis/redis/pull/7818

This is an update to the documentation describing the GT and LT options for the ZADD command.

### DIFF
--- a/commands/zadd.md
+++ b/commands/zadd.md
@@ -18,8 +18,8 @@ the first score argument. Options are:
 
 * **XX**: Only update elements that already exist. Never add elements.
 * **NX**: Don't update already existing elements. Always add new elements.
-* **LT**: Only update existing elements if the new score is **less than** the current score. Adds new elements unless used with **XX**.
-* **GT**: Only update existing elements if the new score is **greater than** the current score. Adds new elements unless used with **XX**.
+* **LT**: Only update existing elements if the new score is **less than** the current score. This flag doesn't prevent adding new elements.
+* **GT**: Only update existing elements if the new score is **greater than** the current score. This flag doesn't prevent adding new elements.
 * **CH**: Modify the return value from the number of new elements added, to the total number of elements changed (CH is an abbreviation of *changed*). Changed elements are **new elements added** and elements already existing for which **the score was updated**. So elements specified in the command line having the same score as they had in the past are not counted. Note: normally the return value of `ZADD` only counts the number of new elements added.
 * **INCR**: When this option is specified `ZADD` acts like `ZINCRBY`. Only one score-element pair can be specified in this mode.
 
--- a/commands/zadd.md
+++ b/commands/zadd.md
@@ -18,6 +18,8 @@ the first score argument. Options are:
 
 * **XX**: Only update elements that already exist. Never add elements.
 * **NX**: Don't update already existing elements. Always add new elements.
+* **LT**: Only update existing elements if the new score is *less than* the current score. Will add new elements unless **XX** is used in combination. **(Redis 6.2 or greater)**
+* **GT**: Only update existing elements if the new score is *greater than* the current score. Will add new elements unless **XX** is used in combination. **(Redis 6.2 or greater)**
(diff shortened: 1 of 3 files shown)


topics/modules-api-ref.md
```diff
@@ -1053,6 +1053,8 @@ The input flags are:
 
     REDISMODULE_ZADD_XX: Element must already exist. Do nothing otherwise.
     REDISMODULE_ZADD_NX: Element must not exist. Do nothing otherwise.
+    REDISMODULE_ZADD_LT: For existing element, only update if the new score is less than the current score.
+    REDISMODULE_ZADD_GT: For existing element, only update if the new score is greater than the current score.
```
oranagra, line 1057: this change isn't needed, auto generated when we make a release.
but i suppose we can merge it, it'll get overwritten later.

oranagra (approved):
@itamarhaber do we want for 6.2 to release it? or merge it as soon as the code is in redis's `unstable`?

itamarhaber:
Thanks for this @alexronke-channeladvisor - I made a couple of edits but looks good all in all.

@oranagra since this is an extension of an existing command, and because we're pretty sure it will be in the upcoming v6.2.0, I think we can merge once in unstable.

commands/zadd.md
```diff
@@ -10,17 +10,21 @@ members is created, like if the sorted set was empty. If the key exists but does
...
+ZADD options
 ---
 
 ZADD supports a list of options, specified after the name of the key and before
 the first score argument. Options are:
 
 * **XX**: Only update elements that already exist. Never add elements.
 * **NX**: Don't update already existing elements. Always add new elements.
+* **LT**: Only update existing elements if the new score is **less than** the current score. Adds new elements unless used with **XX**.
+* **GT**: Only update existing elements if the new score is **greater than** the current score. Adds new elements unless used with **XX**.
```
oranagra, line 22: I think that mixing XX into that description is harder to understand.
i guess the intention was that it updates only if less/greater than, but can also add new one.
```suggestion
* **LT**: Only update existing elements if the new score is **less than** the current score. This flag doesn't prevent adding new elements if one doesn't exist.
* **GT**: Only update existing elements if the new score is **greater than** the current score. This flag doesn't prevent adding new elements if one doesn't exist.
```

alexronke-channeladvisor: You could drop the "if one doesn't exist" part, since that's the definition of a new element.

alexronke-channeladvisor:
> Thanks for this @alexronke-channeladvisor - I made a couple of edits but looks good all in all.
> 
> @oranagra since this is an extension of an existing command, and because we're pretty sure it will be in the upcoming v6.2.0, I think we can merge once in unstable.

I put "6.2" in the documentation description because the PR had been flagged with that milestone. Is there a better way to describe that?

topics/modules-api-ref.md
alexronke-channeladvisor: Interesting.  If it's auto-generated, then it seems to be missing some flags.

oranagra: we (manually) auto generate it before publishing a new release. so maybe what's missing are flags that were not released yet?
which flags are missing?

commands/zadd.md
oranagra: updated.. does that reflect the original intent better? (better than "unless used with XX")

oranagra:
@alexronke-channeladvisor the marking of 6.2 in your PR was correct. we where trying to decide if it should be merged right now or wait.. we concluded it can be merged right now.

commands/zadd.md
alexronke-channeladvisor: I think that's a good phrasing.  It was difficult to find a short way to say that.

As for how the options are arranged, it will probably show up like this:
**ZADD key [NX|XX] [GT|LT] [CH] [INCR] score member [score member ...]**

I figured that was simpler than trying to do the technically more accurate:
**ZADD key [NX|[[XX] [GT|LT]]] [CH] [INCR] score member [score member ...]**

oranagra: I agree

oranagra (approved):

PR #1398, opened 2020-09-25
Fixed an awk regexp escape sequence warning in the makefile.

wingunder:
The following warning was printed by awk, when running 'make spell':
awk: cmd. line:1: warning: regexp escape sequence `\&' is not a known
regexp operator

This patch removes the escape in front of the '&' in the awk regexp
and fixes this issue.


itamarhaber:
Thanks @wingunder - this is brilliant \& merging!

itamarhaber (approved):
Tested locally, works.

PR #1399, opened 2020-09-25
Added make-check-targets for duplication inside the ./wordlist file.

wingunder:
Spell checking by the 'spell' makefile target is done by using
white-list words originating from commands in the ./commands.json file
as well as words in the ./wordlist file. Two ways of word duplication
in these two files are possible:

   1. A new command gets added to the ./commands.json file, while the
   command already exists in the ./wordlist file.

   2. The same word simply exist twice inside the ./wordlist file.

This patch adds two makefile targets that check for these two cases,
when running 'make spell'.

### DIFF
--- a/makefile
+++ b/makefile
@@ -16,8 +16,31 @@ clients:
 tools:
 	ruby utils/clients.rb tools.json
 
+check_duplicate_wordlist: wordlist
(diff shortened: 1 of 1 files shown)


itamarhaber:
Thanks, LGTM, merging.

itamarhaber (approved):

PR #1405, opened 2020-09-29
Add GET parameter to SET command

nykolaslima:
Closes #2349

Add optional GET parameter to SET command in order to set a new value to
a key and retrieve the old key value. With this change we can deprecate
`GETSET` command and use only the SET command with the GET parameter.

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -2818,6 +2818,14 @@
           "XX"
         ],
         "optional": true
+      },
+      {
+        "name": "get",
+        "type": "enum",
+        "enum": [
+          "GET"
+        ],
+        "optional": true
       }
     ],
     "since": "1.0.0",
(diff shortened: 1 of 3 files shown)


commands.json
```diff
@@ -2815,7 +2815,8 @@
         "type": "enum",
         "enum": [
           "NX",
-          "XX"
+          "XX",
+          "GET"
         ],
```
oranagra, line 2820: doesn't that mean that they're all mutually exclusive?
but in fact we do allow GET together with XX, so i gues it needs a separate block like the KEEPTTL has (and possibly a comment in the command's page that mention  that GET is incompatible with NX)

oranagra (commented):
@itamarhaber please have a look too

commands.json
nykolaslima: Done!

itamarhaber (approved):

itamarhaber:
LGTM

PR #1408, opened 2020-10-03
[WIP] Added Threaded I/O topic

2014BDuck:
The idea of threaded I/O is mainly about to delay the reads/writes to next aeProcessEvent loop. Working on this topic.


2014BDuck:
@itamarhaber PTAL.

2014BDuck:
Hi. Any update or suggestions for this one?

itamarhaber:
Heya @2014BDuck - this is on @yossigo's and my plates, sorry for the delay and thanks for the contribution.

2014BDuck:
Hi. Is there any suggestion for this one ~? Thanks.

2014BDuck:
Hi. Any feedback for this topic proposal?

topics/threaded-io.md
```diff
@@ -0,0 +1,127 @@
...
+    while(1)
+        ...
+        pthread_mutex_lock(&io_threads_mutex[id]);
+        pthread_mutex_unlock(&io_threads_mutex[id]);
+        ...
+...
+```
+
+## Limitations
+The I/O threads won't work (won't postpone read/write) when:
```
filipecosta90, line 121: I/O threading is currently not supported with TLS.

topics/threaded-io.md
```diff
@@ -0,0 +1,127 @@
+# Threaded I/O
+## Introduction
```
filipecosta90, line 2: Should we mention why is threading avoided in Redis ( simplicity among others .... )

topics/threaded-io.md
```diff
@@ -0,0 +1,127 @@
...
+io-threads 4
+```
+The `io-threads` option control how many threads you want to use on reading/writing client connections.
+- 1 or not configured: Use single thread like what we have in the older version.
+- n (n >= 2): Use n threads(1 main thread & n-1 I/O threads) to handle network reads/writes.
+
+```
+io-threads-do-reads yes
+```
+By default, Redis uses I/O threads on the writing process only. If you want to use it on the reading process, you have to enable this option as well. Usually, threaded reads don't help much.
```
filipecosta90, line 19: should we avoid the double meaning of the word process?

topics/threaded-io.md
```diff
@@ -0,0 +1,127 @@
...
+
+### Blocking and Locking
+There will be a racing condition if the I/O thread keeps monitoring and editing the `io_threads_list[i]` list. So Redis prepared another global variable named `io_threads_pending`. It's a thread number length list and stores an integer value representing the number of the remaining task that thread i should handle. When the main thread finished dispatching task to `io_threads_list`, it set `io_threads_pending[i]` to the actual task count.
+
+The I/O threads will stay in a `for (int j = 0; j < 1000000; j++)` loop and checking if `io_threads_pending[id] != 0`. So when the value changed to an positive integer, thread i begin to read/write through the client in `io_threads_list[i]`. Once I/O threads finish their job, they will empty the list in `io_threads_list[i]`, and set `io_threads_pending[i]` to 0.
```
filipecosta90, line 105: Should we swap the code (`for....`) for it's purpose? example: The I/O threads will busy-wait waiting for work ( writes or reads )?

topics/threaded-io.md
```diff
@@ -0,0 +1,127 @@
...
+  +----------+   +----------+  +----------+  +----------+
+  | client 4 |   | client 5 |  | client 6 |
+  +----------+   +----------+  +----------+
+
+```
+
+Each thread can iterate over the clients they should read, do read action, and store the result to `client.querybuf` attribute. The main thread will also handle part of the pending tasks after dispatching. In this example, thread 0, which is the main thread, will read through client 0 and then client 4, thread 1 will handle client 1 and client 5, and so on.
+
+### Blocking and Locking
+There will be a racing condition if the I/O thread keeps monitoring and editing the `io_threads_list[i]` list. So Redis prepared another global variable named `io_threads_pending`. It's a thread number length list and stores an integer value representing the number of the remaining task that thread i should handle. When the main thread finished dispatching task to `io_threads_list`, it set `io_threads_pending[i]` to the actual task count.
```
filipecosta90, line 103: Should we use this section to describe briefly that the main thread signals the io-threads for pending work by setting an atomic variable `io_threads_pending`, then the threads move from busy-waiting to a stage in which they process the pending writes/reads and during that time the main thread is busy waiting for that work to be done and after that the io-threads are put into a busy-wait state again? I believe that the code analysis is too much for the docs? IMHO it's better to describe what the code does... WDYT?

filipecosta90:
@2014BDuck I believe this topic will be very useful but IMHO we can reduce a bit the amount of `code` analysis done in the page and move towards explaining the redis threading current implementation/approach for io as so well described by Salvatore in: https://youtu.be/l7e5ve-ffmI?t=1308

2014BDuck:
@filipecosta90 Thanks! I am going to modify it accordingly asap.

topics/threaded-io.md
2014BDuck: Added `- Server running with TLS.`

topics/threaded-io.md
2014BDuck: Added a few desc about this:
- simple and reliable
- entry barrier for every new feature

topics/threaded-io.md
2014BDuck: Replaced `process` with other words

topics/threaded-io.md
2014BDuck: Replaced those code description

topics/threaded-io.md
2014BDuck: Actually the codes might be helpful for me. But yes the document should be more concise and easy to understand for all people who read it.

2014BDuck:
@filipecosta90 Hi. I tried to replace those code blocks and explain the what main thread and io threads is doing. Not sure if it's clear enough. PTAL.

topics/threaded-io.md
```diff
@@ -0,0 +1,119 @@
+# Threaded I/O
+## Introduction
+In order to keep it simple and reliable, Redis is mostly single-threaded. The implementation of parallel query execution might set an entry barrier for every new feature. However, there are certain threaded operations such as UNLINK, slow I/O accesses and other things that are performed on side threads.
+
+Now it is also possible to handle Redis clients' socket reads and writes in different I/O threads. Since especially writing is so slow, normally Redis users use pipelining in order to speed up the Redis performances per core, and spawn multiple instances in order to scale more. Using I/O threads it is possible to easily speedup two times Redis without resorting to pipelining nor sharding of the instance.
+
+## Configuration
+We have two options to control how I/O threads perform.
+```
+io-threads 4
```
filipecosta90, line 10: should we change the 4 by a placeholder <n threads>?

topics/threaded-io.md
```diff
@@ -0,0 +1,119 @@
...
+```
+The `io-threads` option control how many threads you want to use on reading/writing client connections.
+- 1 or not configured: Use single thread like what we have in the older version.
+- n (n >= 2): Use n threads(1 main thread & n-1 I/O threads) to handle network reads/writes.
+
+```
+io-threads-do-reads yes
```
filipecosta90, line 17: same as above. replace yes with [yes|no] ?

filipecosta90 (changes requested):
@2014BDuck small comments I I believe it's looking very good. Pinging @itamarhaber and @yossigo also.

2014BDuck:
Thanks for pointing it out. I've fixed those placeholder stuffs. 
Thank you for your patience @filipecosta90 

ptal @itamarhaber @yossigo @soloestoy ^_^

filipecosta90 (approved):

2014BDuck:
@itamarhaber @yossigo Hello, would you mind sending me some feedback so that I could modify this topic accordingly :)

topics/threaded-io.md
```diff
@@ -0,0 +1,119 @@
+# Threaded I/O
+## Introduction
+In order to keep it simple and reliable, Redis is mostly single-threaded. The implementation of parallel query execution might set an entry barrier for every new feature. However, there are certain threaded operations such as UNLINK, slow I/O accesses and other things that are performed on side threads.
```
madolson, line 3:
```suggestion
Redis is implemented as mostly single-threaded application in order to keep it simple and reliable. There are certain threaded operations such as UNLINK, slow I/O accesses and other operations that are performed on side threads that are coordinated by the main thread.
```

topics/threaded-io.md
```diff
@@ -0,0 +1,119 @@
...
+
+Now it is also possible to handle Redis clients' socket reads and writes in different I/O threads. Since especially writing is so slow, normally Redis users use pipelining in order to speed up the Redis performances per core, and spawn multiple instances in order to scale more. Using I/O threads it is possible to easily speedup two times Redis without resorting to pipelining nor sharding of the instance.
```
madolson, line 5: Starting with Redis 6.0 and great, it is now possible*

topics/threaded-io.md
```diff
@@ -0,0 +1,119 @@
...
+- 1 or not configured: Use single thread like what we have in the older version.
+- n (n >= 2): Use n threads(1 main thread & n-1 I/O threads) to handle network reads/writes.
+
+```
+io-threads-do-reads [yes|no]
+```
+By default, Redis uses I/O threads on the write system call only. If you want to use it on the read  system call, you have to enable this option as well. Usually, threaded reads don't help much.
+
+## High-Level Architecture
+The main idea of this feature is to postpone the read and write action and accumulate those events to the next event loop cycle. So that we have multiple events waiting to be handled and use I/O threads to deal with them.
```
madolson, line 22: I didn't follow this sentence. The high level idea is that IO operations can be done in parallel by making the networking primitives for reading and writing to a client thread safe. So we split the work into identifying the clients with pending operations and then we fan out threads to do it.

topics/threaded-io.md
```diff
@@ -0,0 +1,119 @@
...
+
+1. main thread prepare the pending list for read
+2. I/O threads read from the connections parallelly
+3. main thread executes command one by one
+4. main thread prepare the pending list for write
+5. I/O threads write reply content to the connections parallelly
+
+
+## Low-Level Design
+### Where is the pending list
```
madolson, line 38: My take is that this section is way too detailed. Someone will probably just go read the code if they want to understand this low level of detail. I think if you remove all of the specific implementation details (names of the arguments, functions, etc), and just leave what the algorithm looks like, it will probably be 50% shorter and just as readable.

I would also say you maybe should fold the high-level/low-level design together and just describe the write flow and then the read flow. Although they have some similarities, and you were trying to generalize, they end up being pretty different.

topics/threaded-io.md
```diff
@@ -0,0 +1,119 @@
...
+We have two options to control how I/O threads perform.
+```
+io-threads [thread-num]
+```
+The `io-threads` option control how many threads you want to use on reading/writing client connections.
+- 1 or not configured: Use single thread like what we have in the older version.
+- n (n >= 2): Use n threads(1 main thread & n-1 I/O threads) to handle network reads/writes.
+
+```
+io-threads-do-reads [yes|no]
```
madolson, line 17: I would punt the configuration to the end or interleave the configuration with the introduction of the topics. io-threads-do-reads is hard to understand without context.

madolson (commented):
Thanks for taking the time to help improve our corpus of documentation! I left some high level comments to try to make it clearer.

jiekun:
Close as stale.

PR #1411, opened 2020-10-08
Uses double timeout value in blocking commands

itamarhaber:
WRT https://github.com/redis/redis-doc/pull/1347#discussion_r500768125

### DIFF
--- a/commands/bzpopmin.md
+++ b/commands/bzpopmin.md
@@ -5,7 +5,7 @@ members to pop from any of the given sorted sets.
 A member with the lowest score is popped from first sorted set that is
 non-empty, with the given keys being checked in the order that they are given.
 
-The `timeout` argument is interpreted as an integer value specifying the maximum
+The `timeout` argument is interpreted as an double value specifying the maximum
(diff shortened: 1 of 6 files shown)


commands/bzpopmin.md
```diff
@@ -23,6 +23,10 @@ popped from.
   where a member was popped, the second element is the popped member itself,
   and the third element is the score of the popped element.
 
+@history
+
+* `>= 6.0`: `timeout` is interpreted as a double instead of an integer.
+
```
oranagra, line 29: don't we need a comment like that in BZPOPMAX too?
and also in LMOVE and BRPOPLPUSH

itamarhaber: WRT BZPOPMAX, as it refers to BZPOPMIN for syntax I thought it would suffice, but sure, it can be added.

WRT LMOVE, it doesn't have a timeout and BLMOVE is a new command anyway.

WRT BRPOPLPUSH, same logic as BZs but added.

oranagra (approved):

PR #1412, opened 2020-10-08
Rewrite all links to redis.io with https scheme.

cristiangreco:
On some pages, Firefox shows a warning in the url bar when visiting the website over https, caused by images being loaded over http.

For the sake of completeness, here I've moved all internal links to use https scheme.

redis docs


itamarhaber (approved):
Hello @cristiangreco - thank you, that's an important PR, merging.

PR #1414, opened 2020-10-11
Switch GEORADIUS example arg ordering to match commands.json

mmkal:
commands.json specifies `WITHCOORD` comes before `WITHDIST`: https://github.com/redis/redis-doc/blob/8fa5ba577f3f9a90f55027c5a0d83bec8271ad27/commands.json#L1319-L1334

### DIFF
--- a/commands/georadius.md
+++ b/commands/georadius.md
@@ -60,5 +60,5 @@ Both commands were introduced in Redis 3.2.10 and Redis 4.0.0 respectively.
 GEOADD Sicily 13.361389 38.115556 "Palermo" 15.087269 37.502669 "Catania"
 GEORADIUS Sicily 15 37 200 km WITHDIST
 GEORADIUS Sicily 15 37 200 km WITHCOORD
-GEORADIUS Sicily 15 37 200 km WITHDIST WITHCOORD
+GEORADIUS Sicily 15 37 200 km WITHCOORD WITHDIST
 ```


itamarhaber (commented):
Hello @mmkal - the order of arguments, in this case, is irrelevant as both `WITHDIST` and `WITHCOORD` are considered optional flags rather than positional arguments. Therefore, I'm not sure this fix is needed.

mmkal:
@itamarhaber you are right it's not strictly necessary - but my use case is generating test cases for a strongly-typed client from docs examples, so it's a nice-to-have. My hope was that it at least does no harm, and it encourages consistency in a minor way.

commands/georadius.md
```diff
@@ -60,5 +60,5 @@ Both commands were introduced in Redis 3.2.10 and Redis 4.0.0 respectively.
 GEOADD Sicily 13.361389 38.115556 "Palermo" 15.087269 37.502669 "Catania"
 GEORADIUS Sicily 15 37 200 km WITHDIST
 GEORADIUS Sicily 15 37 200 km WITHCOORD
-GEORADIUS Sicily 15 37 200 km WITHDIST WITHCOORD
+GEORADIUS Sicily 15 37 200 km WITHCOORD WITHDIST
```
zuiderkwast, line 63: I think it's better to change commands.json to list them in the order `[WITHDIST] [WITHHASH] [WITHCOORD]` since that's the order the information is returned when all (or two of them) are specified.

nermiller:
@zuiderkwast ok to close this PR?

zuiderkwast:
@nermiller thanks for going over these old PRs! i'm ok with closing this one.

PR #1416, opened 2020-10-12
Add handy-redis to list of clients

mmkal:
Hello - this is to add a [client I maintain](https://npmjs.com/package/handy-redis) to clients.json

I also re-ordered some of the clients and updated some "active" flags (rough method I used: no activity for >1 year and very few downloads => not active).

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -787,21 +787,21 @@
   },
 
   {
+    "name": "xredis",
+    "language": "Node.js",
+    "repository": "https://github.com/razaellahi/xredis",
+    "description": "Redis client with redis ACL features",
+    "authors": ["razaellahi531"],
+    "recommended": false,
+    "active": true
+  },
+  
+  {
     "name": "node_redis",
     "language": "Node.js",
     "repository": "https://github.com/NodeRedis/node_redis",
     "description": "Recommended client for node.",
     "authors": ["mranney"],
-    "recommended": true,
-    "active": true
-  },
-
-  {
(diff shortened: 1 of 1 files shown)


itamarhaber (commented):
Hello @mmkal - thanks for the PR.

I'd be happy to merge the updates to the `active` property as well as your client's entry, but please revert the ordering changes (in any case, the current rendering is alphabetical at redis.io)

clients.json
```diff
@@ -806,13 +806,23 @@
     "active": true
   },
 
+  {
+    "name": "handy-redis",
+    "language": "Node.js",
+    "repository": "https://github.com/mmkal/handy-redis",
+    "description": "A wrapper around node_redis with Promise and TypeScript support.",
+    "authors": ["mmkal"],
```
itamarhaber, line 814: This is supposed to be your (the author's) Twitter handle. I can see one exists, but it appears inactive so if you're not using it you can just leave it empty, i.e.:
```suggestion
    "authors": [],
```

itamarhaber (commented):
Thanks for reverting - please lmk about the Twitter handle.

clients.json
mmkal: Thanks for pointing out! I don't have that twitter handle sadly. I don't have a GitHub-equivalent twitter right now. Will open a new PR someday when I do 🙃

itamarhaber (approved):

PR #1417, opened 2020-10-13
update sentinel doc for adding ACL authentication

hwware:
Note: This PR is for review only since In line 790 I left a placeholder for the redis version for releasing ACL on Sentinel. We can replace it with correct version once this is released.

### DIFF
--- a/topics/sentinel.md
+++ b/topics/sentinel.md
@@ -785,14 +785,10 @@ configured with `requirepass`, the Sentinel configuration must include the
 Configuring Sentinel instances with authentication
 ---
 
-You can also configure the Sentinel instance itself in order to require
-client authentication via the `AUTH` command, however this feature is
-only available starting with Redis 5.0.1. Start from Redis 6.2.0, ACL 
-is also supported for Sentinel mode. please visit (https://redis.io/topics/acl)
-for more details.
+You can also configure the Sentinel instance itself to require client authentication via the `AUTH` command.
+This feature is only available starting with Redis 5.0.1 by using requirepass config, and starting from Redis 6.2 by using ACL (details below).
 
-In order to do authentication with non-acl mode, just add the following 
-configuration directive to all your Sentinel instances:
+The old pre-6.2 mode, just add the following configuration directive to all your Sentinel instances:
 
     requirepass "your_password_here"
 
@@ -805,13 +801,13 @@ This means that **you will have to configure the same `requirepass` password in 
 
 Before using this configuration make sure your client library is able to send the `AUTH` command to Sentinel instances.
 
-In order to do authentication with acl way, you need provide following configuration directive to all your Sentinel instances:
+Starting Redis 6.2 you can use ACL to directives to control the users and passwords which Sentinel allows to connect with (see https://redis.io/topics/acl), additionally you need provide following configuration directive to all your Sentinel instances:
 
     sentinel sentinel-user <username>
 
     sentinel sentinel-pass <password>
 
-This two configurations provide acl user and acl password to authenticate with other sentinels. you will need to make sure each Sentinel instances have configured same acl user and password with sufficient permissions. In this way, as long as acl user configured is not 'default' user, you don't need to configure the same `requirepass` password in all the Sentinel instances.
+These two configurations provide acl user and ACL password to authenticate with other sentinels. you will need to make sure each Sentinel instances have configured same ACL user and password with sufficient permissions.
 
 Sentinel clients implementation
 ---
--- a/topics/sentinel.md
+++ b/topics/sentinel.md
@@ -583,29 +583,39 @@ so forth.
 Sentinel commands
 ---
 
+The `SENTINEL` command, as of Redis 2.8, is the main API for Sentinel. The following is the list of its subcommands (minimal version is noted for where applicable):
+
+* **SENTINEL CKQUORUM `<master name>`** Check if the current Sentinel configuration is able to reach the quorum needed to failover a master, and the majority needed to authorize the failover. This command should be used in monitoring systems to check if a Sentinel deployment is ok.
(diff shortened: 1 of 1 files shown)


oranagra:
i tried to improve a bit, not sure if i did good.
@itamarhaber please have a look and comment.

topics/sentinel.md
```diff
@@ -801,7 +798,18 @@ When configured this way, Sentinels will do two things:
 
 This means that **you will have to configure the same `requirepass` password in all the Sentinel instances**. This way every Sentinel can talk with every other Sentinel without any need to configure for each Sentinel the password to access all the other Sentinels, that would be very impractical.
 
-Before using this configuration make sure your client library is able to send the `AUTH` command to Sentinel instances.
+Before using this configuration, make sure your client library can send the `AUTH` command to Sentinel instances.
+
+With Redis 6.2, Sentinel instances can be configured with ACL authentication, so clients need to provide both a user and a password. As with password-only authentication, the same ACL user is used to authenticate both incoming clients and outgoing Sentinel connections. You can create and manage the Sentinel ACL user on all instances with regular [ACL directives](/topics/acl), for example:
+
+    ACL SETUSER my-sentinel-user on >my-sentinel-password +@all
```
itamarhaber, line 805: @hwware is this a good example? Perhaps something more restrictive?

itamarhaber:
Also tried improving - one pending question.

oranagra:
@itamarhaber 
```
As with password-only authentication, the same ACL user is used to authenticate both incoming clients and outgoing Sentinel connections.
```
that seems wrong to me.

1. with the new `sentinel-pass` thingy, we now have separate control on the outgoing AUTH (`sentinel-pass`), and the incoming one (`requirepass` or `ACL`). that's true even if you don't use ACL users, and just have password, right?
2. i'd like to discourage the use of `requirepass` entirely, i.e. people are expected to only use `sentinel-pass` even on systems with a single user and no ACL.

although now that i think of it, it may be cumbersome to use (comparing to the old single field that controls all). so we should at least discourage the use of `requirepass` with ACL, since we know it causes problems (https://github.com/redis/redis/pull/7883)

topics/sentinel.md
hwware: @itamarhaber oh yes, maybe  give just enough permission for the sentinel user, let me test this.. thanks for catching up this.

hwware:
Hi @oranagra, I think the first point you mentioned make sense to me, since the sentinel-user and pass controls how this sentinel auth with others for the outbound traffic, and the inbound traffic will be controlled by acl or requirepass config in current sentinel. However normally since sentinels in a cluster need to talk with each other, normally they should be configured with the same acl user and same sentinel-user and pass.  Maybe we should say more specific on this. 
regarding the second one, I think we can mention something like it is not recommended that requirepass use together with ACL in https://github.com/redis/redis/blob/unstable/redis.conf#L795, how do you think?

itamarhaber:
Ok, I think we're good now - please review @hwware @oranagra (sorry, got a little carried away :))

topics/sentinel.md
```diff
@@ -583,26 +583,36 @@ so forth.
...
-* **SENTINEL ckquorum `<master name>`** Check if the current Sentinel configuration is able to reach the quorum needed to failover a master, and the majority needed to authorize the failover. This command should be used in monitoring systems to check if a Sentinel deployment is ok.
-* **SENTINEL flushconfig** Force Sentinel to rewrite its configuration on disk, including the current Sentinel state. Normally Sentinel rewrites the configuration every time something changes in its state (in the context of the subset of the state which is persisted on disk across restart). However sometimes it is possible that the configuration file is lost because of operation errors, disk failures, package upgrade scripts or configuration managers. In those cases a way to to force Sentinel to rewrite the configuration file is handy. This command works even if the previous configuration file is completely missing.
+* **ROLE** This command returns the string "sentinel" and a list of monitored masters. For more information refer to the `ROLE` command.
+* **SHUTDOWN** Shut down the Sentinel instance.
```
oranagra, line 596: was a little bit hard to review.
for some reason github shows that you changed all the lines about the SENTINEL commands.
but comparing them one by one i see they are unchanged, and all you did is add the section at the top with the generic management APIs.

I see you also added COMMAND which is not yet merged, but we can assume is gonna be soon.
i suppose we need a versioning paragraph here to mention that ACL and COMMAND are new to 6.2.

also, i see all the PUBSUB commands are missing from this list. they are documented below, but maybe this list should have some note and a link to the other section.

topics/sentinel.md
```diff
@@ -785,12 +795,33 @@ configured with `requirepass`, the Sentinel configuration must include the
...
+The default user is used by Sentinel to connect to other instances. You can provide the credentials of another superuser with the following configuration directives:
+
+    sentinel sentinel-user <username>
+    sentinel sentinel-pass <password>
+
+Where `<username>` and `<password>` are the Sentinel's ACL user and password, respectively.
+
+Lastly, for authenticating incoming client connections, you can create a restricted Sentinel user like this:
+
+    127.0.0.1:5000> ACL SETUSER sentinel-user ON >user-password -@all +@connection +SENTINEL|MASTERS +SENTINEL|MASTER +SENTINEL|GET-MASTER-ADDR-BY-NAME
```
oranagra, line 820: can someone confirm that this line would be enough for sentinels to talk to each other and to their job?

oranagra (commented):
Aren't we missing a mention about ACL in the top of the `Sentinel and Redis authentication` section? it looks like the `auth-user` option isn't documented at all!

topics/sentinel.md
itamarhaber: This is actually for incoming client connections and should be enough for discoverability, but an affirmation would be good.

topics/sentinel.md
itamarhaber: I uppercased the subcommands in these lines, perhaps that's why.

WRT COMMAND - exactly.

Pub/Sub - as it is covered in another section I didn't put it in, but I also felt it was missing in terms of completeness. Will add.

itamarhaber:
You're totally right about auth-user - I saw it was missing and forgot to add it. Fixing.

oranagra (approved):
LGTM..
i suppose someone needs to take another look at the live web version after this is merged to make sure the formatting works, and the reading flow and headers make sense.

PR #1418, opened 2020-10-13
update sentinel commands doc

hwware:
there are some sentinel command missing in the sentinel doc. this PR adds these commands.

### DIFF
--- a/docs/manual/sentinel.md
+++ b/docs/manual/sentinel.md
@@ -606,8 +606,8 @@ The `SENTINEL` command is the main API for Sentinel. The following is the list o
 * **SENTINEL FLUSHCONFIG** Force Sentinel to rewrite its configuration on disk, including the current Sentinel state. Normally Sentinel rewrites the configuration every time something changes in its state (in the context of the subset of the state which is persisted on disk across restart). However sometimes it is possible that the configuration file is lost because of operation errors, disk failures, package upgrade scripts or configuration managers. In those cases a way to force Sentinel to rewrite the configuration file is handy. This command works even if the previous configuration file is completely missing.
 * **SENTINEL FAILOVER `<master name>`** Force a failover as if the master was not reachable, and without asking for agreement to other Sentinels (however a new version of the configuration will be published so that the other Sentinels will update their configurations).
 * **SENTINEL GET-MASTER-ADDR-BY-NAME `<master name>`** Return the ip and port number of the master with that name. If a failover is in progress or terminated successfully for this master it returns the address and port of the promoted replica.
-* **SENTINEL INFO-CACHE** (`>= 3.2`) Return cached `INFO` output from masters and replicas.
-* **SENTINEL IS-MASTER-DOWN-BY-ADDR <ip> <port> <current-epoch> <runid>** Check if the master specified by ip:port is down from current Sentinel's point of view. This command is mostly for internal use.
+* **SENTINEL INFO-CACHE `<master name>`** (`>= 3.2`) Return cached `INFO` output from masters and replicas.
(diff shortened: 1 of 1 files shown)


oranagra (commented):
@hwware can you please have a closer look at `info-cache` implementation.
i think there documentation and/or code may be wrong.

specifically, this seems like dead code, maybe it meant to check `<3`:
```c
if (c->argc < 2) goto numargserr
```
and besides it looks like it's getting multiple arguments.

one other thing i'd like you to check, is when did any of these got introduced? maybe some of these existed since forever, and were not mentioned in the documentation on purpose (for being debug tools)?

also, i think we need to add a note about version compatibility (i.e. the `myid` will exist starting 6.2, but it may also be true for other additions).
for normal commands we already have a way to note that, but for sentinel we don't yet, and i suppose we need to come up with something.

hwware:
Hello @oranagra , the command of `sentinel info-cache` basicallly returning the cached info for all the specified masters their slaves, and if no master name provided, it simply returning all cached master and its slave info and its time for how long it was received.that's why it is ok to have 0, 1 or more additional arguments. I think we need to mention this in a better way to avoid people get confused. I will do a fix shortly.
Also I think you bring a good topic here, I also feel there are lots of documentation missing good explation for Sentinel part. Most of the sentinel code was written around 6 years ago (including the info-cache one) so there are missing lots of things. I would suggest we put the sentinel command also into the commands page (https://redis.io/commands). But that requires a lot of documentation change and not sure whether we worth to do it. How do you think?

oranagra:
I rather not mix Sentinel commands with Redis commands in the same table / index.
We can maybe create a similar section lile `redis.io/sentinal/commands/` which will behave the same (using the same infrastructure). but i'm not sure it worth the effort.
@itamarhaber any opinion on that?

For now, let's just try to improve the existing page.
But i just wonder if some of these commands were not documented on purpose.
If one of these commands exist from day one (before the documentation page was created), then it must mean that someone considered that an internal debug command and didn't bother to document it.
much like the Redis REPLCONF and (most of) DEBUG commands aren't mentioned.

hwware:
@oranagra  I see, and that is a very good point.. I checked this when at the very beginning Salvatore's commit https://github.com/redis/redis-doc/commit/78c6f7ec0182f4c67d31243a6f80f5594d9224cf#diff-4578785ae9106bc325486759f3f1eae0de1bf7c18af224a19d7c3146cc10b872R161 he mentioned the is-master-down-by-addr command in the doc, but later he removed it. Maybe the is-master-down-by-addr api is used for debugging purpose and  Salvatore deliberatey removed it in the doc.

oranagra:
looks like at some point Salvatore added a new sentinel doc that didn't include `is-master-down-by-addr` in the list of commands, and much later he removed the old doc that did have it.
173155f6d3b185348fb62c8dceb5616a33d51ffa
03c4ad501966606c0b1883486a876c15102279ae
however, even the new doc does refer to `is-master-down-by-addr` (it is mentioned a few times, just not listed), so i think we can add it back.

i see that:
`pending-scripts` exists since 2012 (3f194a9d2571df21717b635f037040407a51d50d)
`info-cache` exists since 2014 (f8c73e38b5f5a57d6d573fc6fdfed5b68d4879c8)
`simulate-failure` exists since 2015 (fb3af75f7478c6d0717d076289976f16b2389e28)
so it seems to me that the later two (being more recent addition) might have just been overlooked. but i'd guess that the first one isn't listed on purpose.

@itamarhaber maybe you can think of some old-timer that can shed more light on this?

itamarhaber:
Regrettably, I can't think of someone like that.

oranagra:
ok. so let's do what we think is right.
@hwware wanna refresh this PR and decide which of the unlisted commands should be listed and which ones can be skipped?

PR #1419, opened 2020-10-19
updating xadd command doc for adding nomkstream option

hwware:


zuiderkwast (changes requested):
This is useful, but the same information has already been added, I guess in other PRs. This PR can be closed.

The current XADD docs contains near the top of the page: "The creation of stream's key can be disabled with the NOMKSTREAM option." Under Return value: "The command returns a Null reply when used with the NOMKSTREAM option and the key doesn't exist." Under History: ">= 6.2: Added the NOMKSTREAM option, MINID trimming strategy and the LIMIT option."

itamarhaber:
Yep, sorry @hwware for missing yours and doing the work myself. OTOH, it was part of a larger effort (having nested optional in the syntax) so that's my excuse :)

Closing.

PR #1422, opened 2020-10-25
Some cleanups to streams doc

yoav-steinberg:


topics/streams-intro.md
```diff
@@ -56,13 +56,13 @@ Note that in this case, the minimum ID is 0-1 and that the command will not acce
...
+Now we are finally able to append entries in our stream via **XADD**. However, while appending data to a stream is quite obvious, the way streams can be queried in order to extract data is not so obvious. If we continue with the analogy of the log file, one obvious way is to mimic what we normally do with the Unix command `tail -f`, that is, we may start to listen in order to get the new messages that are appended to the stream. Note that unlike the blocking list operations of Redis, where a given element will reach a single client which is blocking in a *pop style* operation like **BLPOP**, with streams we want multiple consumers to see the new messages appended to the stream. The same way many `tail -f` processes can see what is added to a log. Using the traditional terminology we want the streams to be able to *fan out* messages to multiple clients.
```
itamarhaber, line 59: Putting the period here makes the context unclear - better use a pair of these instead.

```suggestion
Now we are finally able to append entries in our stream via **XADD**. However, while appending data to a stream is quite obvious, the way streams can be queried in order to extract data is not so obvious. If we continue with the analogy of the log file, one obvious way is to mimic what we normally do with the Unix command `tail -f`, that is, we may start to listen in order to get the new messages that are appended to the stream. Note that unlike the blocking list operations of Redis, where a given element will reach a single client which is blocking in a *pop style* operation like **BLPOP**, with streams we want multiple consumers to see the new messages appended to the stream (the same way many `tail -f` processes can see what is added to a log). Using the traditional terminology we want the streams to be able to *fan out* messages to multiple clients.
```

itamarhaber (commented):
Thanks @yoav-steinberg - one minor comment.

topics/streams-intro.md
yoav-steinberg: :+1:

itamarhaber (approved):

PR #1426, opened 2020-11-01
Adds the RESET command

itamarhaber:
Ref: https://github.com/redis/redis/pull/7982

Missing the 'Time complexity' which could be O(N+M+Q+P+...) though.

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -2580,6 +2580,11 @@
     "since": "1.0.0",
     "group": "generic"
   },
+  "RESET": {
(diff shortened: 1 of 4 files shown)


yossigo (approved):
I don't think time complexity for meta-data (i.e. watched keys, etc.) matters, and it is already ignored with similar commands (e.g. `DISCARD`).

itamarhaber:
@yossigo re complexity: https://github.com/redis/redis-doc/issues/1202

PR #1428, opened 2020-11-03
Add documentation for the new zdiff/zdiffstore commands

felipou:
Add documentation for the new commands added in https://github.com/redis/redis/pull/7961

I took the liberty to also add missing entries for the zinter and zunion commands in the "Movable Keys" section of COMMAND.

There's one thing I need help with: the complexity of these new commands. My math skills are rusty, so I didn't even try very hard to check it. I just adapted the complexity of SDIFF, but I'm thinking it is too broad, or maybe even incorrect. Basically, ZDIFF and SDIFF use the same algorithms, which can be one of two depending on the size of the input sets. I thought about just doing `O(min(a, b))`, but maybe there's a better way. The complexity stated in the documentation is the one from algorithm 2 in the code.

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -3547,6 +3547,52 @@
     "since": "2.0.0",
     "group": "sorted_set"
   },
+  "ZDIFF": {
+    "summary": "Subtract multiple sorted sets",
+    "complexity": "O(N)+O(M*log(M)) worst case with N is the total number of elements in all given sets and M being the number of elements in the resulting sorted set.",
+    "arguments": [
+      {
+        "name": "numkeys",
+        "type": "integer"
+      },
+      {
+        "name": "key",
+        "type": "key",
+        "multiple": true
+      },
+      {
+        "name": "withscores",
+        "type": "enum",
+        "enum": [
+          "WITHSCORES"
+        ],
+        "optional": true
+      }
+    ],
+    "since": "6.2.0",
+    "group": "sorted_set"
+  },
+  "ZDIFFSTORE": {
+    "summary": "Subtract multiple sorted sets and store the resulting sorted set in a new key",
+    "complexity": "O(N)+O(M*log(M)) worst case with N is the total number of elements in all given sets and M being the number of elements in the resulting sorted set.",
+    "arguments": [
+      {
+        "name": "destination",
+        "type": "key"
+      },
+      {
+        "name": "numkeys",
+        "type": "integer"
+      },
+      {
+        "name": "key",
+        "type": "key",
+        "multiple": true
+      }
+    ],
+    "since": "6.2.0",
+    "group": "sorted_set"
+  },
   "ZINCRBY": {
     "summary": "Increment the score of a member in a sorted set",
     "complexity": "O(log(N)) where N is the number of elements in the sorted set.",
--- a/commands.json
+++ b/commands.json
@@ -3574,7 +3574,7 @@
   },
   "ZDIFFSTORE": {
     "summary": "Subtract multiple sorted sets and store the resulting sorted set in a new key",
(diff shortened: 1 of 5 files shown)


oranagra (commented):
looks good to me.. 
i'll get back to you about the complexity after consulting with someone who knows better about these than me.

commands/zdiffstore.md
```diff
@@ -0,0 +1,25 @@
+Computes the difference between the first set and all the successive sorted
+sets given by the specified keys, and stores the result in `destination`.
+It is mandatory to provide the number of input keys (`numkeys`) before passing
+the input keys.
```
itamarhaber, line 4:
```suggestion
Computes the difference of `numkeys` sorted sets given by the specified keys,
between the first and all the successive sets, and stores the result in
`destination`.
```

itamarhaber (commented):
The big-O notation appears correct to my unmathematical eye, but another verificationl would be good.

commands/command.md
```diff
@@ -104,8 +104,12 @@ Cluster client needs to parse commands marked `movablekeys` to locate all releva
 Complete list of commands currently requiring key location parsing:
 
   - `SORT` - optional `STORE` key, optional `BY` weights, optional `GET` keys
+  - `ZUNION` - keys stop when `WEIGHT` or `AGGREGATE` starts
```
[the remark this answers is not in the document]
felipou: 🙂 👍

itamarhaber (approved):

commands/zdiffstore.md
```diff
@@ -1,7 +1,6 @@
-Computes the difference between the first set and all the successive sorted
-sets given by the specified keys, and stores the result in `destination`.
-It is mandatory to provide the number of input keys (`numkeys`) before passing
-the input keys.
```
oranagra: i think the previous text was clearer.
why did you change it?
i.e. what did you want to improve? maybe we can find another way to have both?

what i liked is that the first sentence was clear about that this command does, without getting into the fine (confusing) details.

i.e:
```
Computes the difference between the first set and all the successive sorted
sets given by the specified keys, and stores the result in `destination`.
```

maybe you were bothered by the fact that the "first" one is actually the destination?
so maybe something like:
```
Computes the difference between the first input set and all the successive input
sets given by the specified keys, and stores the result in `destination`.
```

itamarhaber: I adapted the text to follow the "convention" set by previous commands. I also didn't like the "It is mandatory..." part so that was another trigger for the change.

oranagra: @itamarhaber so the current:
```
Computes the difference of `numkeys` sorted sets given by the specified keys,
between the first and all the successive sets, and stores the result in
`destination`.
```
seems clear to you? (to me it looks confusing)

commands.json
```diff
@@ -3547,6 +3547,52 @@
     "since": "2.0.0",
     "group": "sorted_set"
   },
+  "ZDIFF": {
+    "summary": "Subtract multiple sorted sets",
+    "complexity": "O(N)+O(M*log(M)) worst case with N is the total number of elements in all given sets and M being the number of elements in the resulting sorted set.",
```
oranagra, line 3552: @itamarhaber i noticed that [SDIFF](https://redis.io/commands/sdiff) documentation just mentions:
```
Time complexity: O(N) where N is the total number of elements in all given sets.
```
although the code says:
```c
     * Algorithm 1 is O(N*M) where N is the size of the element first set
     * and M the total number of sets.
     *
     * Algorithm 2 is O(N) where N is the total number of elements in all
     * the sets.
```
do you think it's because the documentation wasn't updated when the algorithm was improved? or is it trying to state the worse case in a more simplistic way?

itamarhaber:
Also needs an update to notifications.md with the new `zdiffstore` event.

oranagra (approved):

PR #1431, opened 2020-11-08
The high-performance redis client supports .NETCORE/.NET4.0/.NET4.5

nnhy:
The high-performance redis client supports .NETCORE/.NET4.0/.NET4.5, which is specially optimized for big data and message queuing. The average daily call volume of single online application is 10 billion

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -1778,7 +1778,7 @@
     "url": "https://github.com/NewLifeX/NewLife.Redis",
     "repository": "https://github.com/NewLifeX/NewLife.Redis",
     "description": "The high-performance redis client supports .NETCORE/.NET4.0/.NET4.5, which is specially optimized for big data and message queuing. The average daily call volume of single online application is 10 billion",
-    "authors": ["BigStone"],
+    "authors": [],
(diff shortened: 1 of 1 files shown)


clients.json
```diff
@@ -1772,6 +1772,16 @@
     "active": true
   },
 
+  {
+    "name": "NewLife.Redis",
+    "language": "C#",
+    "url": "https://github.com/NewLifeX/NewLife.Redis",
+    "repository": "https://github.com/NewLifeX/NewLife.Redis",
+    "description": "The high-performance redis client supports .NETCORE/.NET4.0/.NET4.5, which is specially optimized for big data and message queuing. The average daily call volume of single online application is 10 billion",
+    "authors": ["BigStone"],
```
itamarhaber, line 1781: This should contain the Twitter handles of the author(s). It appears that the handle provided is inactive so you may just want to leave this empty like so.

```suggestion
    "authors": [],
```

itamarhaber (approved):
Thanks @nnhy - merging and best of luck with the new client!

PR #1433, opened 2020-11-08
more stream doc cleanups

yoav-steinberg:
Finished going over all of the streams doc. This includes all my fixes.


itamarhaber (approved):
Thanks for the massive editing @yoav-steinberg - have you considering teaching English as a profession? ;) Merging.

PR #1439, opened 2020-11-11
Adds the COPY command

itamarhaber:
Ref: https://github.com/redis/redis/pull/7953

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -976,7 +976,7 @@
   },
   "COPY": {
     "summary": "Copy a key",
-    "complexity": "O(1) to access the source and destination keys. An additional O(N*M) to iterate and copy the value, where N is the number of Redis objects composing the value and M their average size. For small string values the time complexity is thus O(1)+O(1*M) where M is small, so simply O(1). However for sorted set values the complexity is O(N*M*log(N)) because inserting values into sorted sets is O(log(N)).",
+    "complexity": "For small string values and ziplist encoded collections the time complexity is O(1). For most other collections the complexity is O(N) where N is the size of the collection.",
     "since": "6.2.0",
     "arguments": [
       {
--- a/commands.json
+++ b/commands.json
@@ -976,7 +976,7 @@
   },
   "COPY": {
     "summary": "Copy a key",
-    "complexity": "For small string values and ziplist encoded collections the time complexity is O(1). For most other collections the complexity is O(N) where N is the size of the collection.",
+    "complexity": "O(N) worst case for collections, where N is the number of nested items. O(1) for string values.",
     "since": "6.2.0",
     "arguments": [
       {
--- a/commands.json
+++ b/commands.json
@@ -974,6 +974,36 @@
     "since": "2.0.0",
     "group": "server"
   },
+  "COPY": {
+    "summary": "Copy a key",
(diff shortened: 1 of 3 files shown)


commands.json
```diff
@@ -974,6 +974,36 @@
     "since": "2.0.0",
     "group": "server"
   },
+  "COPY": {
+    "summary": "Copy a key",
+    "complexity": "O(1) to access the source and destination keys. An additional O(N*M) to iterate and copy the value, where N is the number of Redis objects composing the value and M their average size. For small string values the time complexity is thus O(1)+O(1*M) where M is small, so simply O(1). However for sorted set values the complexity is O(N*M*log(N)) because inserting values into sorted sets is O(log(N)).",
```
oranagra, line 979: @itamarhaber considering that we're inserting elements to the sorted set in the right order. i guess the O(log(N)) mention should be removed, right?

commands/copy.md
```diff
@@ -0,0 +1,24 @@
...
+
+* `1` if `source` was copied.
+* `0` if `source` was not copied.
+
+@examples
+
+```
+SET dolly "sheep"
+COPY dolly clone
+GET clone
```
oranagra, line 23: LOL

commands.json
itamarhaber: Re Streams: I think in most cases CG and such should be considerably smaller than the actual entries, so I'm tending not to mention that. As for stream-node-max-entries, this is basically a constant so O(1) :) 

list-max* and hash-max* are also constantish they are implicitly referred to by the opening sentence. I think we're ok as long as we want to relay the general gist of worst case complexity, rather than detailed per-type per-encoding description. Agree?

commands/copy.md
itamarhaber: https://www.youtube.com/watch?v=wgalixpASLo

oranagra: https://youtu.be/FavUpD_IjVY

commands.json
oranagra: @itamarhaber
hashes that are small (smaller than hash-max-ziplist-entries are O(1) covered by the beginning of the of the sentence, and larger ones are O(N) covered by the second part (where N is the number of entires).
But for lists the complexity ix O(N) where N is **not** the number of elements in the list.
And the same goes for streams.

p.s. The original content of this complexity block in this PR also attempted to refer to M (the size of the string), right? AFIAK we never mention these.

bottom line, i feel the current (revise) version of the text covers strings, sets, hashes properly. and i understand from your note that maybe we should skip the part that mention "ziplist encoding", not sure if it has any other reference in the commands docs.
but i feel that for lists and streams this text is lacking.

oranagra: summary of discussion on voice call:
`O(N) == O(N/100)`, so we can say that quicklist complexity is O(N).
same goes for streams, so we'll just say O(1) for strings, and O(N) for collections.

commands.json
```diff
@@ -974,6 +974,36 @@
     "since": "2.0.0",
     "group": "server"
   },
+  "COPY": {
+    "summary": "Copy a key",
+    "complexity": "For small string values and ziplist encoded collections the time complexity is O(1). For most other collections the complexity is O(N) where N is the size of the collection.",
```
itamarhaber, line 979: @oranagra wdut?

```suggestion
    "complexity": "O(N) worst case for collections, where N is the number of nested items. O(1) for string values.",
```

oranagra: LGTM

oranagra (approved):

PR #1441, opened 2020-11-12
add missing user information in client list

hwware:
after https://github.com/redis/redis-doc/pull/1440 I found user field is also missing in CLIENT LIST command.


commands/client-list.md
```diff
@@ -35,6 +35,7 @@ Here is the meaning of the fields:
 * `argv-mem`: incomplete arguments for the next command (already extracted from query buffer)
 * `tot-mem`: total memory consumed by this client in its various buffers
 * `redir`: client id of current client tracking redirection
+* `user`: current authenticated user of this client
```
itamarhaber, line 38: Thanks @hwware 

```suggestion
* `user`: the authenticated username of the client
```

Issue #1008, opened 2018-10-18
XPENDING syntax is wrong


itamarhaber:
According to the implementation (ref https://github.com/antirez/redis/pull/5459), should be:

```
XPENDING key group [start end count [consumer]]
```

instead of the current:

```
XPENDING key group [start end count] [consumer]
```

antirez:
Yep, can't be fixed.

PR #1443, opened 2020-11-20
Adds nested arguments

### LINKED ISSUE #1007: [Missing] XGROUP CREATE ... MKSTREAM

itamarhaber:
Relies on https://github.com/redis/redis-io/pull/221

Should fix #1007, #1008 and wrt to https://github.com/redis/redis/pull/7972#issuecomment-730867094

Also adds `XADD ... NOMKSTREAM` (https://github.com/redis/redis/pull/7910).

`XADD` before/after shots:
image
image

`XGROUP` before/after shots:
image
image

`XPENDING` before/after shots:
image
image

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -4341,8 +4341,33 @@
         "type": "key"
       },
       {
-        "name": "ID",
-        "type": "string"
+        "type": "block",
+        "block": [
+          {
+            "command": "MAXLEN"
+          },
+          {
+            "type": "block",
+            "block": [
+              {
+                "command": "~",
+                "optional": true
+              },
+              {
+                "name": "length",
+                "type": "integer"
+              }
+            ]
+          }
+        ],
+        "optional": true
+      },
+      {
+        "type": "enum",
+        "enum": [
+          "*",
+          "ID"
+        ]
       },
       {
         "name": [
@@ -4500,7 +4525,7 @@
         "multiple": true
       },
       {
-        "name": "id",
+        "name": "ID",
         "type": "string",
         "multiple": true
       }
@@ -4513,30 +4538,56 @@
     "complexity": "O(1) for all the subcommands, with the exception of the DESTROY subcommand which takes an additional O(M) time in order to delete the M entries inside the consumer group pending entries list (PEL).",
     "arguments": [
       {
-        "command": "CREATE",
-        "name": [
-          "key",
-          "groupname",
-          "id-or-$"
-        ],
-        "type": [
-          "key",
-          "string",
-          "string"
+        "type": "block",
+        "block": [
+          {
+            "command": "CREATE",
+            "name": [
+              "key",
+              "groupname"
+            ],
+            "type": [
+              "key",
+              "string"
+            ]
+          },
+          {
+            "name": "id",
+            "type": "enum",
+            "enum": [
+              "ID",
+              "$"
+            ]
+          },
+          {
+            "command": "MKSTREAM",
+            "optional": true
+          }
         ],
         "optional": true
       },
       {
-        "command": "SETID",
-        "name": [
-          "key",
-          "groupname",
-          "id-or-$"
-        ],
-        "type": [
-          "key",
-          "string",
-          "string"
+        "type": "block",
+        "block": [
+          {
+            "command": "SETID",
+            "name": [
+              "key",
+              "groupname"
+            ],
+            "type": [
+              "key",
+              "string"
+            ]
+          },
+          {
+            "name": "id",
+            "type": "enum",
+            "enum": [
+              "ID",
+              "$"
+            ]
+          }
         ],
         "optional": true
       },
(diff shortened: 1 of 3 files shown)


commands.json
```diff
@@ -4341,8 +4341,37 @@
...
+        "type": "block",
+        "block": [
+          {
+            "command": "MAXLEN"
+          },
+          {
+            "type": "block",
+            "block": [
+              {
+                "command": "~",
```
oranagra, line 4353: the text in t_stream.c says:
```
/* XADD key [MAXLEN [~|=] <count>] <ID or *> [field value] [field value] ... */
```
so i guess you need `=` in addition to `~`.

oranagra:
@itamarhaber i suppose that other than redis's help.h generator https://github.com/redis/redis/pull/8074/files there are other tools that might be using this json file and could break due to that change.
do you have any idea how to find these and warn them?

commands.json
itamarhaber: LOL - I didn't know/remember this was actually supported, maybe it wasn't documented intentionally? I pushed a corrective PR (and updated the XADD screenshot + the commands' pages) but since it is a) the default b) undocumented (and maybe c) MAXLEN isn't likely to evolve much), perhaps we should revert this.

itamarhaber:
> do you have any idea how to find these and warn them?

Good point. Other than my own scripts, I'm only aware of https://github.com/redis/redis-doc/pull/1231 via @mmkal. If this is merged, some eco-system breakage may indeed occur.

commands.json
oranagra: i don't think we should revert this.
it seems good that the exact (default) behavior can be explicitly requested rather than implicitly.

but i now see it's not covered by any test test, so maybe it's not at all working.
do you mind creating a PR to add it to the tests? if it'll not work, we can consider trimming it from the docs, but if it does, i think we must keep supporting it, have it in the tests and docs.

oranagra:
@redis/core-team maybe one of you is aware of any external tools that reply on our commands.json, and have an opinion here.
these tools may break the very moment we press the merge button (if like our help.h generator they fish the data directly from `https://raw.githubusercontent.com/redis/redis-doc/master/commands.json`)

yossigo:
@oranagra At least one I'm aware of is the Carmine Redis client for Clojure which AFAIR infers supported commands from a `commands.json` file it downloads.

commands.json
itamarhaber: https://github.com/redis/redis/pull/8083

oranagra:
@itamarhaber can you try to contact them, and since we see this turns out to be a popular practice, maybe publish this change somewhere?
and then let's give everyone a few weeks to respond or adjust before merging.

mmkal:
@oranagra I maintain a library which generates typescript from commands.json: https://github.com/mmkal/handy-redis. It checks out this repo as a git subtree so wouldn't break instantaneously. I can see what effect this change has and report back here.

oranagra:
@mmkal thanks, please make the necessary adjustments on your side and report back, we'll wait.
or if you find a problem, or have a suggestion for improvement please share.

itamarhaber:
cc @ptaoussanis (Carmine) @laixintao (IRedis) <- please review :)

itamarhaber:
@oranagra btw, 2.7K results for https://github.com/search?q=redis+%22commands.json%22&type=code

ptaoussanis:
Thanks for pinging about this 👍   

**No objections** for this change for Carmine (Clojure Redis client):

- Carmine _does make use_ of `commands.json`.
- But only when cutting a new release that manually updates to the latest `commands.json`.
  - I.e. **no users will be affected by the merge**.
- I'll just need to update the parser in Carmine for the next release, which is not a problem.

mmkal:
@oranagra @itamarhaber I manually merged in this branch. It didn't technically break anything (other than exposing a small bug in the generation code that was previously hidden), but since the prior code didn't understand `"type": "block"` at all, the generated types were just `unknown`. I was able to figure reverse-engineer how `block`s worked and generate something mostly sensible.

The actionable feedback I have is - can the new `block` and `enum` arguments include a string `name` property, like the others? Right now they're resulting in parameters called things like `xadd_block_1` which isn't super informative. [Example](https://github.com/mmkal/handy-redis/blob/08711e937df1d18355561bd728433446dac72fd5/src/generated/interface.ts#L7666-L7671) and [diff link](https://github.com/mmkal/handy-redis/compare/redis-doc/nested-args-handle-block#diff-19cefe200314d91021fb5153f096a401b299cca12790fd10b849ef1df4255c85R7683). 

___

The lack of a formal spec for `commands.json` is a bit of a maintenance concern, as a downstream consumer, though. It's not ideal having to reverse-engineer. Are there rules for what `block` looks like? Some of them have optional sub-arguments, which I mapped to union types, which seems right but hard to be confident won't turn out to be a wrong assumption at some point. Similarly, it's the `enum` ambiguity in the previously-linked PR.

Re-bumping the proposal I made to use json-schema for command arguments, which could solve that problem by using a well-defined format, no need for inventing things like `block`: https://github.com/redis/redis-doc/pull/1232. My client generates that format itself, here's what the xadd `block` argument becomes in json-schema: [it just uses standard syntax like `anyOf` and `array`](https://github.com/mmkal/handy-redis/blob/08711e937df1d18355561bd728433446dac72fd5/codegen/schema.json#L6797-L6828).

laixintao (approved):
Thanks for letting me know, IRedis using `commands.json` to display a help message on the bottom bar. 

image

The arguments' color is based on `"type"`. So I might need to handle the new type of `"block"`.

Don't worry, go ahead.

itamarhaber:
Thanks all - merging and let's keep our fingers crossed that the earth will not stop spinning :)

itamarhaber:
@mmkal - added the names for blocks via #1449

PR #1444, opened 2020-11-22
Adds exclusive ranges to XRANGE/XREVRANGE

itamarhaber:
See https://github.com/redis/redis/pull/8072

Also simplifies iteration examples as a result.

### DIFF
--- a/commands/xrange.md
+++ b/commands/xrange.md
@@ -72,7 +72,7 @@ of past events in a stream.
 The range is open (inclusive) by default, meaning that the reply can include
 entries with IDs matching the query's start and end intervals. It is possible
 to specify an open interval (exclusive) by prefixing the ID with the
-character `(`.
+character `(`. This is useful for iterating the stream, as explained below.
 
 ## Returning a maximum number of entries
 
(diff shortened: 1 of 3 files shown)


commands/xrange.md
```diff
@@ -67,6 +67,13 @@ Used in this way `XRANGE` works as a range query command to obtain entries
 in a specified time. This is very handy in order to access the history
 of past events in a stream.
 
+## Exclusive ranges
+
+The range is open (inclusive) by default, meaning that the reply can include
+entries with IDs matching the query's start and end intervals. It is possible
+to specify an open interval (exclusive) by prefixing the ID with the
+character `(`.
```
oranagra, line 75:
```suggestion
The range is open (inclusive) by default, meaning that the reply can includes
entries with IDs matching the query's start and end intervals. It is possible
to specify a closed interval (exclusive) by prefixing the ID with the
character `(`.
```

commands/xrevrange.md
```diff
@@ -14,54 +14,6 @@ enough to send:
 
     XREVRANGE somestream + - COUNT 1
 
-## Iterating with XREVRANGE
```
oranagra, line 17: do you think maybe there's a reason to keep it for the benefit of redis 5.0 and 6.0 users?

commands/xrange.md
```diff
@@ -67,6 +67,13 @@ Used in this way `XRANGE` works as a range query command to obtain entries
 in a specified time. This is very handy in order to access the history
 of past events in a stream.
 
+## Exclusive ranges
+
+The range is open (inclusive) by default, meaning that the reply can include
```
itamarhaber, line 72:
```suggestion
The range is close (inclusive) by default, meaning that the reply can include
```

commands/xrange.md
itamarhaber: Oops - thanks.

commands/xrevrange.md
itamarhaber: Added back as a subsection in XRANGE

oranagra (approved):
it would have been nice to put all that details inside some collapsible block

<details>
<summary>pre-6.2 iteration</summary>

bla
bla
bla

</details>

but we do what we can, so i guess what you did is enough.

PR #1452, opened 2020-12-03
Clients: Add forks of C and Erlang clients and mark old as inactive

zuiderkwast:
Redis and Redis Cluster in particular are used by Ericsson. Forks
of no longer active clients in C and Erlang are now maintained by
the Nordix Foundation with support by Ericsson.

This commit adds these forks and marks the original clients as inactive
and in some case as not recommented.

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -43,6 +43,14 @@
     "description": "Redis client with a focus on performance",
     "authors": ["wooga"],
     "recommended": true,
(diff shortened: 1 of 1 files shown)


itamarhaber (approved):
Thanks @zuiderkwast - good know and thanks for picking up the maintainership over these!

PR #1454, opened 2020-12-04
Fix modules intro links

smukil:
I noticed some of the links were broken while going through them. I've fixed them here.


itamarhaber (commented):
Thanks for the offer to help @smukil, much appreciated. However, the links aren't broken as the website renders them to the correct URLs - please see at: https://redis.io/topics/modules-intro

smukil:
Ah got it. Thank you. I'll close this then

PR #1457, opened 2020-12-07
Update the doc about maxmemory on replicas

gomex:
I can't find any mention about this ignore on documentation. I found this issue, but nothing besides that: https://github.com/redis/redis/issues/1875

### DIFF
--- a/topics/replication.md
+++ b/topics/replication.md
@@ -330,6 +330,6 @@ This behavior ensures that masters and replicas stay consistent, which is usuall
 
 Note that since the replica by default does not evict, it may up end using more memory than what is set via `maxmemory` (since there are certain buffers that may be larger on the replica, or data structures may sometimes take more memory and so forth). So make sure you monitor your replicas and make sure they have enough memory to never hit a real out-of-memory condition before the master hits the configured `maxmemory` setting.
 
-In order to change this behavior, it is possible to allow a replica don't ignore the maxmemory. The configuration directives to use is:
+In order to change this behavior, it is possible to allow a replica to not ignore the maxmemory. The configuration directives to use is:
 
     replica-ignore-maxmemory no
--- a/topics/replication.md
+++ b/topics/replication.md
@@ -324,7 +324,7 @@ down it, than can be restarted, and finally AOF can be enabled again.
 Maxmemory on replicas
 ---
 
-Replicas don't honor `maxmemory` because by default a replica will ignore this setting (unless it is promoted to master after a failover or manually). It means that the eviction of keys will be just handled by the master, sending the DEL commands to the replica as keys evict in the master side.
+Replicas don't honor `maxmemory` because by default a replica will ignore this setting (unless it is promoted to master after a failover or manually). It means that the eviction of keys will just be handled by the master, sending the DEL commands to the replica as keys evict in the master side.
 
 This behavior ensures that masters and replicas stay consistent, and is usually what you want, however if your replica is writable, or you want the replica to have a different memory setting, and you are sure all the writes performed to the replica are idempotent, then you may change this default (but be sure to understand what you are doing).
 
@@ -333,4 +333,3 @@ Note that since the replica by default does not evict, it may end using more mem
 In order to change this behavior, it is possible to allow a replica don't ignore the maxmemory. The configuration directives to use is:
 
     replica-ignore-maxmemory no
-
--- a/topics/replication.md
+++ b/topics/replication.md
@@ -328,7 +328,7 @@ Replicas don't honor `maxmemory` because by default a replica will ignore this s
 
 This behavior ensures that masters and replicas stay consistent, which is usually what you want. However, if your replica is writable, or you want the replica to have a different memory setting, and you are sure all the writes performed to the replica are idempotent, then you may change this default (but be sure to understand what you are doing).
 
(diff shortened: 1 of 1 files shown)


itamarhaber:
Hello @gomex 

Thanks for surfacing this - it is an important tidbit of info that should be covered in this doc. The behavior is already documented, however, in the redis.conf file (see the [`replica-ignore-maxmemory` directive](https://github.com/redis/redis/blob/6.0/redis.conf#L904-L922)) that I think does an excellent job at explaining the topic. I suggest you adopt (and perhaps adapt) that for this PR.

gomex:
@itamarhaber I have updated the description about it. Let me know if I need to add more info or modify something here.

topics/replication.md
```diff
@@ -321,3 +321,16 @@ It is not possible to partially resynchronize a replica that restarted via the
 AOF file. However the instance may be turned to RDB persistence before shutting
 down it, than can be restarted, and finally AOF can be enabled again.
 
+Maxmemory on replicas
+---
+
+Replicas don't honor `maxmemory` because by default a replica will ignore this setting (unless it is promoted to master after a failover or manually). It means that the eviction of keys will be just handled by the master, sending the DEL commands to the replica as keys evict in the master side.
```
zuiderkwast, line 327:
```suggestion
Replicas don't honor `maxmemory` because by default a replica will ignore this setting (unless it is promoted to master after a failover or manually). It means that the eviction of keys will just be handled by the master, sending the DEL commands to the replica as keys evict in the master side.
```

topics/replication.md
```diff
@@ -321,3 +321,16 @@ It is not possible to partially resynchronize a replica that restarted via the
...
+
+This behavior ensures that masters and replicas stay consistent, and is usually what you want, however if your replica is writable, or you want the replica to have a different memory setting, and you are sure all the writes performed to the replica are idempotent, then you may change this default (but be sure to understand what you are doing).
```
zuiderkwast, line 329:
```suggestion
This behavior ensures that masters and replicas stay consistent, which is usually what you want. However, if your replica is writable, or you want the replica to have a different memory setting, and you are sure all the writes performed to the replica are idempotent, then you may change this default (but be sure to understand what you are doing).
```

topics/replication.md
```diff
@@ -321,3 +321,16 @@ It is not possible to partially resynchronize a replica that restarted via the
...
+
+Note that since the replica by default does not evict, it may end using more memory than the one set via maxmemory (there are certain buffers that may be larger on the replica, or data structures may sometimes take more memory and so forth). So make sure you monitor your replicas and make sure they have enough memory to never hit a real out-of-memory condition before the master hits the configured maxmemory setting.
```
zuiderkwast, line 331:
```suggestion
Note that since the replica by default does not evict, it may up end using more memory than what is set via `maxmemory` (since there are certain buffers that may be larger on the replica, or data structures may sometimes take more memory and so forth). So make sure you monitor your replicas and make sure they have enough memory to never hit a real out-of-memory condition before the master hits the configured `maxmemory` setting.
```

topics/replication.md
```diff
@@ -321,3 +321,16 @@ It is not possible to partially resynchronize a replica that restarted via the
...
+
+In order to change this behavior, it is possible to allow a replica don't ignore the maxmemory. The configuration directives to use is:
```
zuiderkwast, line 333:
```suggestion
In order to change this behavior, it is possible to allow a replica to not ignore the maxmemory. The configuration directives to use is:
```

zuiderkwast (commented):
I quickly added some suggestions for (at least what I think) is slightly better English. Then, let's merge this soon!

gomex:
@zuiderkwast done

topics/replication.md
```diff
@@ -321,3 +321,15 @@ It is not possible to partially resynchronize a replica that restarted via the
...
 down it, than can be restarted, and finally AOF can be enabled again.
 
+Maxmemory on replicas
+---
+
+Replicas don't honor `maxmemory` because by default a replica will ignore this setting (unless it is promoted to master after a failover or manually). It means that the eviction of keys will just be handled by the master, sending the DEL commands to the replica as keys evict in the master side.
+
+This behavior ensures that masters and replicas stay consistent, which is usually what you want. However, if your replica is writable, or you want the replica to have a different memory setting, and you are sure all the writes performed to the replica are idempotent, then you may change this default (but be sure to understand what you are doing).
+
+Note that since the replica by default does not evict, it may up end using more memory than what is set via `maxmemory` (since there are certain buffers that may be larger on the replica, or data structures may sometimes take more memory and so forth). So make sure you monitor your replicas and make sure they have enough memory to never hit a real out-of-memory condition before the master hits the configured `maxmemory` setting.
```
zuiderkwast, line 331:
```suggestion
Replicas don't honor `maxmemory` because by default a replica will ignore this setting (unless it is promoted to master after a failover or manually).
It means that the eviction of keys will just be handled by the master, sending the DEL commands to the replica as keys evict in the master side.

This behavior ensures that masters and replicas stay consistent, which is usually what you want.
However, if your replica is writable, or you want the replica to have a different memory setting, and you are sure all the writes performed to the replica are idempotent, then you may change this default (but be sure to understand what you are doing).

Note that since the replica by default does not evict, it may end up using more memory than what is set via `maxmemory` (since there are certain buffers that may be larger on the replica, or data structures may sometimes take more memory and so forth).
So make sure you monitor your replicas and make sure they have enough memory to never hit a real out-of-memory condition before the master hits the configured `maxmemory` setting.
```

zuiderkwast:
Thanks @gomex! It's live now: https://redis.io/topics/replication#maxmemory-on-replicas

PR #1465, opened 2020-12-14
Add GEOSEARCH/GEOSEARCHSTORE command doc and explain geo search process

yangbodong22011:
refer https://github.com/redis/redis/pull/8094

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -1503,6 +1503,208 @@
     "since": "3.2.0",
     "group": "geo"
   },
+  "GEOSEARCH": {
+    "summary": "Query a sorted set representing a geospatial index to fetch members matching a box or circular.",
+    "complexity": "O(N+log(M)), where N is the number of elements in the bounding box of the circular area or box area, the circular area is defined by the center and radius, the box area is defined by the center, height and width, and M is the number of items inside the index.",
+    "arguments": [
+      {
+        "name": "key",
+        "type": "key"
+      },
+      {
+        "command": "FROMMEMBER",
+        "name": "member",
+        "type": "string",
+        "optional": true
+      },
+      {
+        "command": "FROMLOC",
+        "name": [
+          "long",
+          "lat"
+        ],
+        "type": [
+          "double",
+          "double"
+        ],
+        "optional": true
+      },
+      {
+        "command": "BYRADIUS",
+        "name": [
+          "radius",
+          "unit"
+        ],
+        "type": [
+          "double",
+          "string"
+        ],
+        "optional": true
+      },
+      {
+        "command": "BYBOX",
+        "name": [
+          "width",
+          "height",
+          "unit"
+        ],
+        "type": [
+          "double",
+          "double",
+          "string"
+        ],
+        "optional": true
+      },
+      {
+        "name": "withcoord",
+        "type": "enum",
+        "enum": [
+          "WITHCOORD"
+        ],
+        "optional": true
+      },
+      {
+        "name": "withdist",
+        "type": "enum",
+        "enum": [
+          "WITHDIST"
+        ],
+        "optional": true
+      },
+      {
+        "name": "withhash",
+        "type": "enum",
+        "enum": [
+          "WITHHASH"
+        ],
+        "optional": true
+      },
+      {
+        "command": "COUNT",
+        "name": "count",
+        "type": "integer",
+        "optional": true
+      },
+      {
+        "name": "order",
+        "type": "enum",
+        "enum": [
+          "ASC",
+          "DESC"
+        ],
+        "optional": true
+      }
+    ],
+    "since": "6.2",
+    "group": "geo"
+  },
+  "GEOSEARCHSTORE": {
+    "summary": "Query a sorted set representing a geospatial index to fetch members matching a box or circular, and stores the result in destination.",
+    "complexity": "O(N+log(M)), where N is the number of elements in the bounding box of the circular area or box area, the circular area is defined by the center and radius, the box area is defined by the center, height and width, and M is the number of items inside the index.",
+    "arguments": [
+      {
+        "name": "destination",
+        "type": "key"
+      },
+      {
+        "name": "key",
+        "type": "key"
+      },
+      {
+        "command": "FROMMEMBER",
+        "name": "member",
+        "type": "string",
+        "optional": true
+      },
+      {
+        "command": "FROMLOC",
+        "name": [
+          "long",
+          "lat"
+        ],
+        "type": [
+          "double",
+          "double"
+        ],
+        "optional": true
+      },
+      {
+        "command": "BYRADIUS",
+        "name": [
+          "radius",
+          "unit"
+        ],
+        "type": [
+          "double",
+          "string"
+        ],
+        "optional": true
+      },
+      {
+        "command": "BYBOX",
+        "name": [
+          "width",
+          "height",
+          "unit"
+        ],
+        "type": [
+          "double",
+          "double",
+          "string"
+        ],
+        "optional": true
+      },
+      {
+        "name": "withcoord",
+        "type": "enum",
+        "enum": [
+          "WITHCOORD"
+        ],
+        "optional": true
+      },
+      {
+        "name": "withdist",
+        "type": "enum",
+        "enum": [
+          "WITHDIST"
+        ],
+        "optional": true
+      },
+      {
+        "name": "withhash",
+        "type": "enum",
+        "enum": [
+          "WITHHASH"
+        ],
+        "optional": true
+      },
+      {
+        "command": "COUNT",
+        "name": "count",
+        "type": "integer",
+        "optional": true
+      },
+      {
+        "name": "order",
+        "type": "enum",
+        "enum": [
+          "ASC",
+          "DESC"
+        ],
+        "optional": true
+      },
+      {
+        "name": "storedist",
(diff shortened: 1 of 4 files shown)


commands.json
```diff
@@ -1503,6 +1503,208 @@
     "since": "3.2.0",
     "group": "geo"
   },
+  "GEOSEARCH": {
+    "summary": "Query a sorted set representing a geospatial index to fetch members matching a box or circular.",
```
oranagra, line 1507:
```suggestion
    "summary": "Query a sorted set representing a geospatial index to fetch members inside an area of box or circle.",
```

commands.json
```diff
@@ -1503,6 +1503,208 @@
...
+          "DESC"
+        ],
+        "optional": true
+      }
+    ],
+    "since": "6.2",
+    "group": "geo"
+  },
+  "GEOSEARCHSTORE": {
+    "summary": "Query a sorted set representing a geospatial index to fetch members matching a box or circular, and stores the result in destination.",
```
oranagra, line 1602:
```suggestion
    "summary": "Query a sorted set representing a geospatial index to fetch members inside an area of box or circle, and store the result another key.",
```

commands.json
```diff
@@ -1503,6 +1503,208 @@
     "since": "3.2.0",
     "group": "geo"
   },
+  "GEOSEARCH": {
+    "summary": "Query a sorted set representing a geospatial index to fetch members matching a box or circular.",
+    "complexity": "O(N+log(M)), where N is the number of elements in the bounding box of the circular area or box area, the circular area is defined by the center and radius, the box area is defined by the center, height and width, and M is the number of items inside the index.",
```
oranagra, line 1508: Trying to be a bit more accurate and also less verbose (not mention radius and center)
```suggestion
    "complexity": "O(N+log(M)), where N is the number of elements in the a grid aligned bounding box area around the provided circle or box, and M is the number of items inside the actual provided circle or box area.",
```

commands.json
```diff
@@ -1503,6 +1503,208 @@
...
+        ],
+        "optional": true
+      }
+    ],
+    "since": "6.2",
+    "group": "geo"
+  },
+  "GEOSEARCHSTORE": {
+    "summary": "Query a sorted set representing a geospatial index to fetch members matching a box or circular, and stores the result in destination.",
+    "complexity": "O(N+log(M)), where N is the number of elements in the bounding box of the circular area or box area, the circular area is defined by the center and radius, the box area is defined by the center, height and width, and M is the number of items inside the index.",
```
oranagra, line 1603: Trying to be a bit more accurate and also less verbose (not mention radius and center)
```suggestion
    "complexity": "O(N+log(M)), where N is the number of elements in the a grid aligned bounding box area around the provided circle or box, and M is the number of items inside the actual provided circle or box area.",
```

commands.json
```diff
@@ -1503,6 +1503,208 @@
...
+          "height",
+          "unit"
+        ],
+        "type": [
+          "double",
+          "double",
+          "string"
+        ],
+        "optional": true
+      },
```
oranagra, line 1557: @itamarhaber do these work? (providing an array for `name` and `type`)
if not, i guess we should use something like this:
```suggestion
      {
        "type": "block",
        "name": "circle",
        "block": [
          {
            "command": "BYRADIUS",
            "name": "radius"
            "type": "double"
          },
          {
            "name": "unit",
            "type": "enum",
            "enum": [
              "m",
              "km",
              "ft",
              "mi"
            ]
          },
        "optional": true
      },
      {
        "type": "block",
        "name": "box",
        "block": [
          {
            "command": "BYBOX",
            "name": "width",
            "type": "double",
          },
          {
            "name": "height",
            "type": "double",
          },
          {
            "name": "unit",
            "type": "enum",
            "enum": [
              "m",
              "km",
              "ft",
              "mi"
            ]
          },
        "optional": true
      }
```

commands.json
```diff
@@ -1503,6 +1503,208 @@
...
+  "GEOSEARCHSTORE": {
+    "summary": "Query a sorted set representing a geospatial index to fetch members matching a box or circular, and stores the result in destination.",
+    "complexity": "O(N+log(M)), where N is the number of elements in the bounding box of the circular area or box area, the circular area is defined by the center and radius, the box area is defined by the center, height and width, and M is the number of items inside the index.",
+    "arguments": [
+      {
+        "name": "destination",
+        "type": "key"
+      },
+      {
+        "name": "key",
```
oranagra, line 1610:
```suggestion
        "name": "source",
```

commands/geosearchstore.md
```diff
@@ -0,0 +1,3 @@
+This command is like `GEORADIUS`, but stores the result in destination.
+
+It supports the `WITHDIST` option, can refer to `GEORADIUS`, but no additional destination key parameter is needed because the destination key has been specified in `argv[1]`.
```
oranagra, line 3: I suppose you meant to reference GEOSEARCH, not GEORADIUS, right?
```suggestion
This command is like `GEOSEARCH`, but stores the result in destination key.

By default, it stores the results in a sorted set populated with their geospatial information into the destination key.

If `STOREDIST` is provided, it stores the items in a sorted set populated with their distance from the center of the circle or box, as a floating point number, in the same unit specified for that shape.
```

commands/geosearch.md
```diff
@@ -0,0 +1,31 @@
...
+
+* `FROMMEMBER`: It takes the name of a member already existing inside the geospatial index represented by the sorted set.
+* `FROMLOC`: Use parameters to pass longitude and latitude.
+
+The search area is determined by the following options:
+
+* `BYRADIUS`: Similar to GEORADIUS, searching for circular areas according to radius.
+* `BYBOX`: Axis-aligned rectangle, determined by height and width.
+
+`WITHCOORD`, `WITHDIST`, `WITHHASH`, `COUNT`, `ASC|DESC` option can refer to `GEORADIUS` commands.
```
oranagra, line 13: considering we're gonna deprecate GEORADIUS sooner or later, we probably wanna copy the description here rather than refer to it there.
maybe not just yet though.

commands/geosearch.md
```diff
@@ -0,0 +1,31 @@
...
+2. Estimate the geohash encoding bits by radius or max(height/2, width/2) value. Since geohash value represent a box, considering the edge case referenced in [Geohash Wiki](http://en.wikipedia.org/wiki/Geohash), to search fast, we need find the smallest geohash box with surrounding 8 geohash box that could cover all points in radius in the worst case. 
+3. According to the minimum bounding box, exclude areas that do not need to be searched, for example, when area.latitude.min < box.min_lat, which means that south, south_west, and south_east will not need to be searched.
+4. For each geohash box, we convert it to a score range.
+5. For each score range, use `ZRANGEBYSCORE key min max WITHSCORES` to retrieve all point's value and it's score.
+6. For each point value and it's score, we can decode the score to a GeoHash area by geohash-int and judge whether the point meets the search conditions. 
+
```
oranagra, line 23: i think we're missing a section about the response format?
something like:
```suggestion

@return

@array-reply, specifically:

* Without any `WITH` option specified, the command just returns a linear array like ["New York","Milan","Paris"].
* If `WITHCOORD`, `WITHDIST` or `WITHHASH` options are specified, the command returns an array of arrays, where each sub-array represents a single item.

When additional information is returned as an array of arrays for each item, the first item in the sub-array is always the name of the returned item. The other information is returned in the following order as successive elements of the sub-array.

1. The distance from the center as a floating point number, in the same unit specified in the shape.
2. The geohash integer.
3. The coordinates as a two items x,y array (longitude,latitude).

```

commands/geosearch.md
```diff
@@ -0,0 +1,31 @@
...
+1. Calculate the [Minimum bounding box](https://en.wikipedia.org/wiki/Minimum_bounding_box) of the search area. On the one hand, it is used to determine whether the point is in the box(When search box), and the other is to exclude GEOHASH boxes that do not need to be searched (a total of 9).
+2. Estimate the geohash encoding bits by radius or max(height/2, width/2) value. Since geohash value represent a box, considering the edge case referenced in [Geohash Wiki](http://en.wikipedia.org/wiki/Geohash), to search fast, we need find the smallest geohash box with surrounding 8 geohash box that could cover all points in radius in the worst case. 
+3. According to the minimum bounding box, exclude areas that do not need to be searched, for example, when area.latitude.min < box.min_lat, which means that south, south_west, and south_east will not need to be searched.
+4. For each geohash box, we convert it to a score range.
+5. For each score range, use `ZRANGEBYSCORE key min max WITHSCORES` to retrieve all point's value and it's score.
+6. For each point value and it's score, we can decode the score to a GeoHash area by geohash-int and judge whether the point meets the search conditions.
```
oranagra, line 22: Do we really want to get into the implementation details here?
i don't see that GEORADIUS did that.
i think we should focus on what it does and how to use it, not how it works.
@itamarhaber @yangbodong22011 what do you think?

yangbodong22011: I have deleted this part, if we need it later, we can write the topic separately.

commands/geosearch.md
yangbodong22011: done, copy it.

commands.json
oranagra: @itamarhaber if that's right, we need to apply that change to GEOSEARCHSTORE too

oranagra (approved):

PR #1468, opened 2020-12-19
change port to 7002 after redirection

tugberkugurlu:
I believe this port should be `7002` in the example, since it is shown after:

```
redis 127.0.0.1:7000> get foo
-> Redirected to slot [12182] located at 127.0.0.1:7002
"bar"
```


itamarhaber (approved):
Thanks, merging

PR #1472, opened 2020-12-27
Added documentation regarding error statistics (redis/redis#8217)

filipecosta90:
The following PR documents the added features in https://github.com/redis/redis/pull/8217 ( the new errorstats section and a new global error counter ).

### DIFF
--- a/commands/info.md
+++ b/commands/info.md
@@ -15,6 +15,7 @@ The optional parameter can be used to select a specific section of information:
 *   `modules`: Modules section
 *   `keyspace`: Database related statistics
 *   `modules`: Module related sections
+*   `errorstats`: Redis error statistics
 
 It can also take the following values:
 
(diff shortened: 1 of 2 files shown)


commands/info.md
```diff
@@ -324,12 +328,22 @@ Here is the meaning of all fields in the **cpu** section:
...
+ including the number of calls that reached command execution (not rejected),
+ the total CPU time consumed by these commands, the average CPU consumed
+ per command execution, the number of rejected calls
+ (errors prior command execution), and the number of failed calls
+ (errors within the command execution).
 
 For each command type, the following line is added:
 
-*   `cmdstat_XXX`: `calls=XXX,usec=XXX,usec_per_call=XXX`
+*   `cmdstat_XXX`: `calls=XXX,usec=XXX,usec_per_call=XXX,rejected_calls=XXX,failed_calls=XXX`
```
oranagra, line 339: @itamarhaber I know we don't currently document which info fields were added in each version, but maybe considering that this is a dict field we should? or maybe we should for all?

filipecosta90: @oranagra @itamarhaber should we add it as a note? example: Note: `rejected_calls` and `failed_calls` are available since XX

itamarhaber: Generally, we document the history, but in `INFO`'s case this has never been done. Better late than never? I don't think this applies in this case so we can merge as is.

PR #1473, opened 2020-12-28
Module API ref formatting

zuiderkwast:
The page https://redis.io/topics/modules-api-ref has some formatting errors because of improper Markdown syntax which is corrected here, e.g.

* Remove C-style comment markers (`/** .... */` docs copy-pasted directly from source-code?) e.g. [RedisModule_HoldString](https://redis.io/topics/modules-api-ref#coderedismoduleholdstringcode).
* Use proper lists instead of `1) ... 2) ...` which isn't rendered as a list.
* Correct indentation and blank lines inserteded to make code blocks show up as code blocks correctly (e.g. [RedisModule_SubscribeToServerEvent](https://redis.io/topics/modules-api-ref#coderedismodulesubscribetoservereventcode)).
* Make variouls constant/variable/function names as `code`.

To compare, it may be easiest to display the rendered pages [before](https://github.com/redis/redis-doc/blob/master/topics/modules-api-ref.md) and [after](https://github.com/zuiderkwast/redis-doc/blob/module-api-ref-formatting/topics/modules-api-ref.md) side-by-side.

### DIFF
--- a/topics/modules-api-ref.md
+++ b/topics/modules-api-ref.md
@@ -211,9 +211,9 @@ When enabled, automatic memory management tracks and automatically frees
 keys, call replies and Redis string objects once the command returns. In most
 cases this eliminates the need of calling the following functions:
 
-1) `RedisModule_CloseKey()`
-2) `RedisModule_FreeCallReply()`
-3) `RedisModule_FreeString()`
+1. `RedisModule_CloseKey()`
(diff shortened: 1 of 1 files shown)


zuiderkwast:
I realized the module API reference is generated from comments in the C code. I'll fix the formatting there instead.

PR #1474, opened 2020-12-29
geoadd command add CH/NX/XX option

kukey:
geoadd command add CH/NX/XX option

### DIFF
--- a/commands/geoadd.md
+++ b/commands/geoadd.md
@@ -1,47 +1,87 @@
 Adds the specified geospatial items (latitude, longitude, name) to the specified
-key. Data is stored into the key as a sorted set, in a way that makes it possible to later retrieve items using a query by radius with the `GEORADIUS` or `GEORADIUSBYMEMBER` commands.
+key.
+Data is stored into the key as a sorted set, in a way that makes it possible
+to later retrieve items using a query by radius with the `GEORADIUS` or
+`GEORADIUSBYMEMBER` commands, or use `GEOSEARCH` searching within rectangular
+areas.
 
-The command takes arguments in the standard format x,y so the longitude must
-be specified before the latitude. There are limits to the coordinates that
-can be indexed: areas very near to the poles are not indexable. The exact
-limits, as specified by EPSG:900913 / EPSG:3785 / OSGEO:41001 are the following:
+The command takes arguments in the standard format x,y so the longitude must be
+specified before the latitude.
+There are limits to the coordinates that can be indexed: areas very near to the
+poles are not indexable.
+The exact limits, as specified by EPSG:900913 / EPSG:3785 / OSGEO:41001 are the
+following:
 
 * Valid longitudes are from -180 to 180 degrees.
 * Valid latitudes are from -85.05112878 to 85.05112878 degrees.
 
-The command will report an error when the user attempts to index coordinates outside the specified ranges.
+The command will report an error when the user attempts to index coordinates
+outside the specified ranges.
 
-**Note:** there is no **GEODEL** command because you can use `ZREM` in order to remove elements. The Geo index structure is just a sorted set.
+**Note:** there is no **GEODEL** command because you can use `ZREM` in order to
+remove elements.
+The Geo index structure is just a sorted set.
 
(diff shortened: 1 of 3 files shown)


commands/geoadd.md
```diff
@@ -1,47 +1,87 @@
 Adds the specified geospatial items (latitude, longitude, name) to the specified
-key. Data is stored into the key as a sorted set, in a way that makes it possible to later retrieve items using a query by radius with the `GEORADIUS` or `GEORADIUSBYMEMBER` commands.
+key.
+Data is stored into the key as a sorted set, in a way that makes it possible
+to later retrieve items using a query by radius with the `GEORADIUS` or
+`GEORADIUSBYMEMBER` commands, or use `GEOSEARCH` searching within rectangular
```
oranagra, line 5: @itamarhaber do you think considering we're "deprecating" GEORADIUS* maybe we should only mention GEOSEARCH here? (which has references to GEORADIUS)

commands/geoadd.md
```diff
@@ -1,47 +1,87 @@
...
-The command takes arguments in the standard format x,y so the longitude must
-be specified before the latitude. There are limits to the coordinates that
-can be indexed: areas very near to the poles are not indexable. The exact
-limits, as specified by EPSG:900913 / EPSG:3785 / OSGEO:41001 are the following:
+The command takes arguments in the standard format x,y so the longitude must be
+specified before the latitude.
+There are limits to the coordinates that can be indexed: areas very near to the
+poles are not indexable.
+The exact limits, as specified by EPSG:900913 / EPSG:3785 / OSGEO:41001 are the
+following:
```
oranagra, line 13: @kukey why did you re-format this (and other blocks in this file)?
it's hard to see if there any real changes.

commands/geoadd.md
```diff
@@ -1,47 +1,87 @@
...
+
+* **XX**: Only update elements that already exist.
+  Never add elements.
+* **NX**: Don't update already existing elements.
+  Always add new elements.
+* **CH**: Modify the return value from the number of new elements added, to the
+  total number of elements changed (CH is an abbreviation of _changed_).
+  Changed elements are **new elements added** and elements already existing for
+  which **the score was updated**.
+  So elements specified in the command line having the same score as they had in
```
oranagra, line 39: the reference to "score" here is invalid. should be "coordinates"?

commands/geoadd.md
kukey: I use make, it`s auto re-format. I was suspicious at that time.

commands/geoadd.md
kukey: yeah, thanks

commands/geoadd.md
kukey: I fix and revert the format to the old version. Thanks!

kukey: sorry, i use rake format:cached comamnd due to this file format changed.

oranagra (commented):
i see the text still refers to "score" in two places (one of them existed before this PR).
both should be changed to "coordinate".

other than that LGTM.
Waiting for Itamer to respond about the reference to GEORADIUS (if we wanna drop it or not)

kukey:
> i see the text still refers to "score" in two places (one of them existed before this PR).
> both should be changed to "coordinate".
> 
> other than that LGTM.
> Waiting for Itamer to respond about the reference to GEORADIUS (if we wanna drop it or not)

ok

commands/geoadd.md
```diff
@@ -42,6 +54,11 @@ and most other applications. However in the worst case the error may be up to
 
 * The number of elements added to the sorted set, not including elements
   already existing for which the score was updated.
+  When use `CH` option, will return the number of elements added+updated.
```
itamarhaber, line 57:
```suggestion
* When used without optional arguments, the number of elements added to the sorted set excluding existing elements whose scores were updated.
* If the `CH` option is specified, the sum of elements added and updated by the operation.
```

itamarhaber:
Thanks @kukey for all the hard work 👍 merging after some additional edits.

PR #1475, opened 2020-12-30
XAUTOCLAIM docs

guybe7:


commands.json
```diff
@@ -5104,6 +5104,39 @@
...
+        "type": "string"
+      },
+      {
+        "name": "start",
+        "type": "string"
+      },
+      {
+        "command": "COUNT",
+        "name": "count",
+        "type": "integer"
```
oranagra, line 5134:
```suggestion
        "type": "integer"
        "optional": true
```

oranagra (commented):
not sure why the test fails, please look into that.
everything else looks good to me..
@itamarhaber ?

itamarhaber:
@guybe7 please review and tell me if it makes sense

oranagra:
@guybe7 @itamarhaber i pushed two commits.
first one is a refactoring to comply with the future #1485 (so that the diff in the second one is smaller)
second one is the new features, please review.

itamarhaber (approved):
Great stuff :)

PR #1476, opened 2020-12-31
Rename Presto SQL to Trino

martint:
PrestoSQL was rebranded as Trino: https://trino.io/blog/2020/12/27/announcing-trino.html


itamarhaber (approved):
Thanks @martint - merging.

PR #1482, opened 2021-01-01
New ZRANGESTORE and ZRANGE new args, deprecate others

oranagra:
i'm not very comfortable with how this came out.
documenting all the nuances and instructions of BYSCORE / BYLEX / BYRANK in one page can make it quite complex.
instead i added a note to refer to the documentation of the deprecated commands for details (which sounds like a bad idea).
maybe that's a hint that this change is all wrong?
i.e. there are a lot of details about inclusive / exclusive searches and how the lexicographic ordering works, maybe unifying them into one command is wrong?

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -4067,8 +4067,61 @@
     "since": "5.0.0",
     "group": "sorted_set"
   },
+  "ZRANGESTORE": {
+    "summary": "Store a range of members from sorted set into another key",
+    "complexity": "O(log(N)+M) with N being the number of elements in the sorted set and M the number of elements stored into the destination key.",
+    "arguments": [
+      {
+        "name": "dst",
+        "type": "key"
+      },
+      {
+        "name": "src",
+        "type": "key"
+      },
+      {
+        "name": "min",
+        "type": "string"
+      },
+      {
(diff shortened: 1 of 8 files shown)


oranagra:
@itamarhaber i merged your PR (https://github.com/oranagra/redis-doc/pull/1) and added another commit, please have a look.

I see not all the content from the ZRANGEBYSCORE and ZRANGEBYLEX pages was copied into the ZRANGE page.
so if we some day completely remove the pages about ZRANGEBYSCORE and ZRANGEBYLEX, some fine details will be missing.

i guess it's a good compromise, doesn't include all the details of the separate pages, but also better than what i did, adding at least some basic info, while still also referring to the other page
maybe we should add some text saying "please look at the <deprecated command" for more details" ?

do you think that's good to go now?

itamarhaber (approved):
I'm comfortable with it as is.

My intent was to include as many details as possible about each variant, without too much clutter. Other than patterns, I hope I included everything.

WRT to missing patterns: once we overhaul the docs, I think patterns should be moved out of specific commands' man pages, and into a "pattern/use case" topic thing.

PR #1483, opened 2021-01-04
hedis now supports cluster

michaelglass:
as of 0.14.0 hedis supports cluster. See: https://github.com/informatikr/hedis/blob/master/CHANGELOG


itamarhaber (approved):
Woot!

PR #1484, opened 2021-01-04
Adds complexity to SWAPDB

itamarhaber:


commands.json
```diff
@@ -3662,6 +3662,7 @@
   },
   "SWAPDB": {
     "summary": "Swaps two Redis databases",
+    "complexity": "O(N) where N is the count of clients watching or blocking on keys from both databases."
```
oranagra, line 3665: didn't we want to mention `M` the number of watched keys?
or did you conclude that isn't not needed?

itamarhaber: It is mentioned as a part of N so not really needed imo.

oranagra (approved):

oranagra (approved):

PR #1489, opened 2021-01-08
Fixed consistency issue with history versions

madolson:
Made @history consistent, since it was brought up in another PR. No need have people copy the wrong format.


itamarhaber (approved):
You rock... now all that's left is sprinkle the `@history` tag on all pages :)

PR #1490, opened 2021-01-11
update three to four

andy-k:
#1405

### DIFF
--- a/commands/set.md
+++ b/commands/set.md
@@ -13,7 +13,7 @@ The `SET` command supports a set of options that modify its behavior:
 * `KEEPTTL` -- Retain the time to live associated with the key.
 * `GET` -- Return the old value stored at key, or nil when key did not exist.
 
-Note: Since the `SET` command options can replace `SETNX`, `SETEX`, `PSETEX`, `GETSET`, it is possible that in future versions of Redis these four commands will be deprecated and finally removed.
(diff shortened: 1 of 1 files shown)


commands/set.md
```diff
@@ -13,7 +13,7 @@ The `SET` command supports a set of options that modify its behavior:
 * `KEEPTTL` -- Retain the time to live associated with the key.
 * `GET` -- Return the old value stored at key, or nil when key did not exist.
 
-Note: Since the `SET` command options can replace `SETNX`, `SETEX`, `PSETEX`, `GETSET`, it is possible that in future versions of Redis these three commands will be deprecated and finally removed.
+Note: Since the `SET` command options can replace `SETNX`, `SETEX`, `PSETEX`, `GETSET`, it is possible that in future versions of Redis these four commands will be deprecated and finally removed.
```
itamarhaber, line 16:
```suggestion
Note: Since the `SET` command options can replace `SETNX`, `SETEX`, `PSETEX`, `GETSET`, it is possible that in future versions of Redis these commands will be deprecated and finally removed.
```

itamarhaber (commented):
Thanks for noticing, reporting and fixing! I suggest we get rid of the number altogether :)

itamarhaber (approved):

Issue #1332, opened 2020-06-16
XADD documentation doesn't specifically outline what is returned when the operation fails


alexkhorton:
The XADD documentation (https://redis.io/commands/xadd) does not specifically state what is returned by a failed XADD.

Between this statement:
When a user specified an explicit ID to XADD, the minimum valid ID is 0-1, and the user must specify an ID which is greater than any other ID currently inside the stream, otherwise the command will fail.

And this statement:
The command returns the ID of the added entry.

It isn't clear what happens when the XADD fails and no entry is added.

Could you please update the document to specifically state whether what is returned on failure? I assume it's either (1) the ID provided to the XADD command or (2) some token indicating failure.

Thanks!

zuiderkwast:
Fixed by #1495

PR #1495, opened 2021-01-16
Feat: Added what happens when xadd fails

ayruslore:
#This commit is related to the issue opened - (https://github.com/redis/redis-doc/issues/1332) | #1332

### DIFF
--- a/commands/xadd.md
+++ b/commands/xadd.md
@@ -15,7 +15,10 @@ remove data from a stream.
 
 ## Specifying a Stream ID as an argument
 
-A stream entry ID identifies a given entry inside a stream.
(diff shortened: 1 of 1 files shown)


itamarhaber:
Thanks @ayruslore 

I don't think this fixes the original issue. The supposed ambiguity is with this:

> When a user specified an explicit ID to XADD, the minimum valid ID is 0-1, and the user must specify an ID which is greater than any other ID currently inside the stream, otherwise the command will fail.

In my opinion, "fail" is self explanatory, but perhaps it should be changed to "fail and return an error" to fix the original issue.

itamarhaber (approved):

PR #1508, opened 2021-02-01
improve GETDEL description

oranagra:
I feel this is needed since it says that it deletes the key (but it doesn't always do so).
maybe the part about the error return is excessive (we don't normally state that)

### DIFF
--- a/commands/getdel.md
+++ b/commands/getdel.md
@@ -1,9 +1,9 @@
 Get the value of `key` and delete the key.
-This command is similarly to `GET`, except for the fact that it also deletes the key on success (only if the key is string type).
+This command is similar to `GET`, except for the fact that it also deletes the key on success (if and only if the key's value type is a string).
 
 @return
 
-@bulk-string-reply: the value of `key`, `nil` when `key` does not exist, or error if the key is not a string type.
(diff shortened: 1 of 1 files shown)


commands/getdel.md
```diff
@@ -1,9 +1,9 @@
 Get the value of `key` and delete the key.
-This command is identical to `GET`, except for the fact that it also deletes the key.
+This command is similarly to `GET`, except for the fact that it also deletes the key on success (only if the key is string type).
```
itamarhaber, line 2:
```suggestion
This command is similarl to `GET`, except for the fact that it also deletes the key on success (if and only if the key's value type is a string).
```

commands/getdel.md
```diff
@@ -1,9 +1,9 @@
...
 
 @return
 
-@bulk-string-reply: the value of `key`, or `nil` when `key` does not exist.
+@bulk-string-reply: the value of `key`, `nil` when `key` does not exist, or error if the key is not a string type.
```
itamarhaber, line 6:
```suggestion
@bulk-string-reply: the value of `key`, `nil` when `key` does not exist, or an error if the key's value type isn't a string.
```

itamarhaber (approved):

PR #1511, opened 2021-02-10
clients.json: marked node-redis as inactive

mjomble:
With no activity in the repo for a year, it hasn't met the "Clients with some activity in the official repository within the latest six months" criteria for some time.
I kept the `recommended` flag for now, as I'm not sure who decides that and based on what.

Also updated the name from node_redis to node-redis.

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -822,13 +822,13 @@
   },
   
   {
-    "name": "node_redis",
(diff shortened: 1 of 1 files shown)


itamarhaber (approved):
Thanks - merging.

PR #1516, opened 2021-02-14
Little tweak on INCR rate limiter 1

sazzad16:


ayruslore (approved):

itamarhaber:
Hi @sazzad16 - thanks for the suggestion.

I'm not sure I understand why this tweak is needed - please explain.

sazzad16:
@itamarhaber 
1. It reduces round trip (before api call).
2. It limits extra calls. For example: many processes read (GET) `current`=9; so all of those perform api call. This overshoots the limit. This change avoids this scenario.

itamarhaber (approved):

itamarhaber:
Thanks - makes sense now, merging.

PR #1524, opened 2021-02-26
Persistence: Define acronyms RDB and AOF

boukeversteegh:
As per common convention: explain the meaning of an acronym the first time it is introduced.

I couldn't find the definitions of RDB and AOF and had to google for it. This will make the page more easy to follow for newcomers.

### DIFF
--- a/topics/persistence.md
+++ b/topics/persistence.md
@@ -5,10 +5,10 @@ Redis Persistence
 
 Redis provides a different range of persistence options:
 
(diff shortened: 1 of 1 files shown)


itamarhaber (commented):
Thanks for adding the acronym's expanded form.

WRT mdash - although editorially correct, we don't have any of these in the docs so for consistency please revert these.

boukeversteegh:
Done!

itamarhaber (approved):
🙇

PR #1527, opened 2021-03-02
Update lru-cache.md

abrookins:
The existing language implies that `maxmemory` applies only to the data in Redis, which does not appear to be true.

### DIFF
--- a/topics/lru-cache.md
+++ b/topics/lru-cache.md
@@ -19,7 +19,7 @@ Maxmemory configuration directive
 ---
 
 The `maxmemory` configuration directive is used in order to configure Redis
-to use a specified amount of memory for the data set. It is possible to
+to use a specified amount of memory. It is possible to
 set the configuration directive using the `redis.conf` file, or later using
 the `CONFIG SET` command at runtime.
 


abrookins:
@itamarhaber Could you review?

itamarhaber:
Sure - I don't agree with the change proposal as `maxmemory` controls the data set's size, not the entire footprint of the process. What's the motivation for this?

abrookins:
A user asked because of the following behavior:

```
127.0.0.1:6379> flushdb
127.0.0.1:6379> config set maxmemory 2kb

127.0.0.1:6379> set test hey
(error) OOM command not allowed when used memory > 'maxmemory'.

127.0.0.1:6379> info
...
# Memory
used_memory:1103120
used_memory_human:1.05M
...

127.0.0.1:6379> config set maxmemory 2mb
OK
127.0.0.1:6379> set test hey
OK
```

Or in other words: what's going on here exactly? It seems we can't write to Redis unless `maxmemory` is above the total memory used by Redis, even with no data in the database.

After some close reading of the other `maxmemory` references (aside from the LRU page) in the Redis docs, I don't see anywhere else that we call this the maximum memory used for the _data set_. Elsewhere we seem to imply that it's the max memory used by Redis:

> Redis has built-in protections allowing the user to set a max limit to memory usage, using the maxmemory option in the configuration file to put a limit to the memory Redis can use.
-- https://redis.io/topics/faq

Curious about your thoughts here/explanation.

abrookins:
I agree with you @itamarhaber and will close this PR.

At the same time, if I set maxmemory to 1 mb on an empty database, I don't see why I should get an OOM trying to set a string. Especially if `used_memory_dataset` is 67600. But maybe I'm just not thinking that through well enough -- and it's a bit of an edge case anyway. I will follow up in the redis repo.

PR #1533, opened 2021-03-15
Add documentation for Lua behavior under low memory conditions

QuChen88:
Add documentation for Lua behavior under low memory conditions, namely around ways that Lua can cause Redis maxmemory to be exceeded, and how to mitigate this.

### DIFF
--- a/commands/eval.md
+++ b/commands/eval.md
@@ -194,6 +194,18 @@ returned in the format specified above (as a Lua table with an `err` field).
 The script can pass the exact error to the user by returning the error object
 returned by `redis.pcall()`.
 
+## Running Lua under low memory conditions
+
+When the memory usage in Redis exceeds the `maxmemory` limit, the first write command encountered in the Lua script that uses additional memory will cause the script to abort (unless `redis.pcall` was used).
+However, one thing to caution here is that if the first write command does not use additional memory such as DEL, LREM, or SREM, etc, Redis will allow it to run and all subsequent commands in the Lua script will execute to completion for atomicity.
+If the subsequent writes in the script generate additional memory, the Redis memory usage can go over `maxmemory`.
+
+Another possible way for Lua script to cause Redis memory usage to go above `maxmemory` happens when the script execution starts when Redis is slightly below `maxmemory` so the first write command in the script is allowed.
+As the script executes, subsequent write commands continue to generate memory and causes the Redis server to go above `maxmemory`.
+
+In those scenarios, it is recommended to configure the `maxmemory-policy` not to use `noeviction`.
+Also Lua scripts should be short so that evictions of items can happen in between Lua scripts.
+
 ## Bandwidth and EVALSHA
 
 The `EVAL` command forces you to send the script body again and again.


commands/eval.md
```diff
@@ -194,6 +194,24 @@ returned in the format specified above (as a Lua table with an `err` field).
 The script can pass the exact error to the user by returning the error object
 returned by `redis.pcall()`.
 
+## Running Lua under low memory conditions
+
+When the memory usage in Redis exceeds the `maxmemory` limit, the first write command
+encountered in the Lua script that uses additional memory will cause the script to
+abort. However, one thing to caution here is that if the first write command does
```
oranagra, line 201: technically, it'll cause the script to abort if `redis.call` was used, but not if `redis.pcall` was used.
```suggestion
encountered in the Lua script that uses additional memory will fail with -OOM error.
However, one thing to caution here is that if the first write command does
```

commands/eval.md
```diff
@@ -194,6 +194,24 @@ returned in the format specified above (as a Lua table with an `err` field).
...
+atomicity of the script execution. If the subsequent writes in the script generate additional
+memory, the Redis memory usage can go over `maxmemory`.
+
+Another possible way for Lua script to cause Redis memory usage to go above `maxmemory`
+happens when the script execution starts when Redis is slightly below `maxmemory` so the
+first write command in the script is allowed. As the script executes, subsequent write
+commands continue to generate memory and causes the Redis server to go above `maxmemory`.
+
+In those scenarios, it is recommended to configure the `maxmemory-policy` appropriately
+to allow the proper evictions of items.
```
oranagra, line 213: i didn't understand what you meant by that?
the script will "properly" execute regardless of the maxmemory-policy.
one can even argue that it ignores it and bloats the memory in cases where it shouldn't have.

oranagra (commented):
please have a look at the new styling guidelines in README.md.
let's split each sentence to one line, regardless of it's length.
this way subsequent changes will likely to only cause a diff to one line.

@itamarhaber please take a look too

commands/eval.md
QuChen88: Well, the point I am trying to make here is that in between Lua scripts, Redis can evict items to reclaim memory if the `maxmemory-policy` is configured properly. We have seen some user who has configured `maxmemory-policy` to be `no-eviction` when the Lua script keeps blowing up memory, and there was no way redis can evict items to recover.

commands/eval.md
QuChen88: I think Redis will abort the script regardless of whether it is `lua.call` or `lua.pcall`. See https://github.com/redis/redis/blob/unstable/src/scripting.c#L663-L676

oranagra: Right. Sorry, I mixed it up with commands that return an error.

oranagra (approved):

commands/eval.md
MeirShpilraien: @oranagra I think you are correct, just tested, if redis.pcall is used the script is not aborted:
```
127.0.0.1:6379> set x 1
(error) OOM command not allowed when used memory > 'maxmemory'.
127.0.0.1:6379> EVAL "redis.call('set', 'x', '1'); return 1" 0
(error) ERR Error running script (call to f_05c76db2bc9e1d0e1c646a7959c5d6ee8177c884): @user_script:1: @user_script: 1: -OOM command not allowed when used memory > 'maxmemory'.
127.0.0.1:6379> EVAL "redis.pcall('set', 'x', '1'); return 1" 0
(integer) 1
```

oranagra: Thanks.. Good to know. (assuming I'll remember)

MeirShpilraien: But its not only that, the script might also decide to catch the error raised by redis.call so even on redis.call it might not be aborted:
```
127.0.0.1:6379> EVAL "local f = function() redis.call('set', 'x', '1'); end; f() ; return 1" 0
(error) ERR Error running script (call to f_de1355275acdfc1349db5a1a34a0f5db57ff2642): @user_script:1: @user_script: 1: -OOM command not allowed when used memory > 'maxmemory'.
127.0.0.1:6379> EVAL "local f = function() redis.call('set', 'x', '1'); end; pcall(f); return 1" 0
(integer) 1
```

oranagra: Yes, that's actually all very logical, it was my initial intuition (so no need to remember). 
I was just too busy / lazy to check.

QuChen88: I just ran the same Lua test. I noticed that when you do `redis.pcall()`, the script actually aborts internally without generating an error response back to the client. So it looks like the client command succeeded, but the actual write fails in Redis silently. 

```
127.0.0.1:6379> set a 1
(error) OOM command not allowed when used memory > 'maxmemory'.
127.0.0.1:6379> EVAL "redis.call('set', 'x', '1'); return 1" 0
(error) ERR Error running script (call to f_05c76db2bc9e1d0e1c646a7959c5d6ee8177c884): @user_script:1: @user_script: 1: -OOM command not allowed when used memory > 'maxmemory'.
127.0.0.1:6379> scan 0
1) "0"
2) (empty array)
127.0.0.1:6379> EVAL "redis.pcall('set', 'x', '1'); return 1" 0
(integer) 1
127.0.0.1:6379> scan 0
1) "0"
2) (empty array)
```

MeirShpilraien: Yes, pcall will not failed your script. If you check the return value of pcall you will see the error and you have to handle the error in your script.

QuChen88: Do you think it is a good idea if I add another note here that Redis silently fails the write if you are not catching the error when using `lua.pcall()`? I feel that is already a general functionality of `lua.pcall()` that it will not fail your script unless you handle it inside your script.

oranagra: @QuChen88 i suppose more details are always better, but considering the section about the difference between `call` and `pcall` is right above this one, i guess it should be short, maybe something like:
```diff
- will cause the script to abort
+ will cause the script to abort (unless `redis.pcall` was used)
```
this way it's just a tip, sending the reader to get more details in the other section.

also, looking again at the last sentence:
> In those scenarios, it is recommended to configure the `maxmemory-policy` appropriately to allow the proper evictions of items to free up memory in between Lua scripts.

i think it's still unclear what you mean to say, i suggest extending it a bit, maybe give an example of the danger and how to avoid it.

QuChen88: Fixed.

commands/eval.md
```diff
@@ -194,6 +194,17 @@ returned in the format specified above (as a Lua table with an `err` field).
...
+## Running Lua under low memory conditions
+
+When the memory usage in Redis exceeds the `maxmemory` limit, the first write command encountered in the Lua script that uses additional memory will cause the script to abort (unless `redis.pcall` was used).
+However, one thing to caution here is that if the first write command does not use additional memory such as DEL, LREM, or SREM, etc, Redis will allow it to run and all subsequent commands in the Lua script will execute to completion for atomicity.
+If the subsequent writes in the script generate additional memory, the Redis memory usage can go over `maxmemory`.
+
+Another possible way for Lua script to cause Redis memory usage to go above `maxmemory` happens when the script execution starts when Redis is slightly below `maxmemory` so the first write command in the script is allowed.
+As the script executes, subsequent write commands continue to generate memory and causes the Redis server to go above `maxmemory`.
+
+In those scenarios, it is recommended to configure the `maxmemory-policy` to not use `noeviction` to allow evictions of items to free up memory in between Lua scripts.
```
oranagra, line 206: my English isn't perfect but something seemed broken in that sentence.
```suggestion
In those scenarios, it is recommended to configure the `maxmemory-policy` not to use `noeviction` and allow evictions of items to free up memory in between Lua scripts.
```
anyway, isn't that true in any case? (with or without lua)
maybe you want to say that scripts should be short, so that eviction happens in between script executions?

QuChen88: Took the suggestion.

oranagra (approved):

PR #1541, opened 2021-03-26
Fix wrong order of geospatial tuple

iamhesir:
The prompt of command GEOADD in redis-cli (6.2.1) is:
> GEOADD key [NX|XX] [CH] longitude latitude member [longitude latitude member ...]


itamarhaber (approved):
Thanks for catching and fixing this inconsistency @iamhesir - merging.

PR #1547, opened 2021-04-08
update transactions.md

ezerror:
WATCH no longer ignores keys which have expired for MULTI/EXEC, but the document is not updated.


zuiderkwast (changes requested):
Please describe the behaviour rather than referring to a pull request.

From what I can understand after skimming through the PR, there is a difference if the key is expired before or after the WATCH command is sent...? Please explain what happens in each case.

Focus on the most important information first. The latest Redis version is more important, so describe the behaviour of the latest version first and afterwards, the behaviour in previous versions.

Also, use a space after period and comma (not before) and a capital letter to start a sentence.

zuiderkwast:
Replaced by #1590. Thanks anyway!

PR #1549, opened 2021-04-12
Add more details about configuration format

deztructor:
Describe strings format in `redis.conf` according to the way how it is
treated by the configuration parsing logic.


itamarhaber (approved):
Thanks! Merging.

PR #1552, opened 2021-04-13
Remove cli tag from XACK

avimredis:
redis.io page was showing interactive cli for XACK which is blocked and through an error in the response. I removed the cli tag and fixed the example to show how it would look to the user


itamarhaber (approved):
Thanks - merging.

PR #1553, opened 2021-04-15
Update bitfield.md

pjambet:
`OVERFLOW` subcommands also affect subsequent `SET` subcommands, not just `INCRBY`

See: https://github.com/redis/redis/blob/6.2.1/src/bitops.c#L1061-L1103


itamarhaber (approved):
Thanks for catching this slip @pjambet - merging.

PR #1559, opened 2021-04-22
SPOP with COUNT returns empty list (not nil)

sazzad16:
Related code: https://github.com/redis/redis/blob/f5235b2d7660b163c739325ec9284f97d5b963be/src/t_set.c#L465-L466

### DIFF
--- a/commands/spop.md
+++ b/commands/spop.md
@@ -14,7 +14,7 @@ When called without the `count` argument:
 
 When called with the `count` argument:
 
-@array-reply: the removed members, or an empty list when `key` does not exist.
+@array-reply: the removed members, or an empty set when `key` does not exist.
 
 @history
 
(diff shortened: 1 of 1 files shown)


commands/spop.md
```diff
@@ -14,7 +14,7 @@ When called without the `count` argument:
 
 When called with the `count` argument:
 
-@array-reply: the removed members, or `nil` when `key` does not exist.
+@array-reply: the removed members, or an empty set when `key` does not exist.
```
itamarhaber, line 17:
```suggestion
@array-reply: the removed members, or an empty array when `key` does not exist.
```

itamarhaber (approved):

itamarhaber:
Thanks @sazzad16 - changed "set" to "array" for consistency with other commands' replies.

sazzad16:
@itamarhaber I don't as long as it says empty, not nil.

sazzad16:
@itamarhaber I don't mind as long as it says empty, not nil.

PR #1561, opened 2021-04-26
Remove WITHCOORD, WITHDIST and WITHHASH

avimredis:
WITHCOORD, WITHDIST and WITHHASH are not meant to be supported since this command a STORE command, which doesn't support those args.

Fixes issue #8864

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -1857,30 +1857,6 @@
         "optional": true
       },
       {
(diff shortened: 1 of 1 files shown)


avimredis (commented):
Fixes issue #8864

oranagra (approved):

PR #1562, opened 2021-04-28
Add documentation for `firstkey`, `lastkey` and `keystep` parameters of `RedisModule_CreateCommand`

edufelipe:
These parameters of `RedisModule_CreateCommand` were previously undocumented but they are needed for ACL to check permission on keys and by Redis Cluster to figure our how to route the command.

### DIFF
--- a/topics/modules-api-ref.md
+++ b/topics/modules-api-ref.md
@@ -170,7 +170,7 @@ order to report keys, like in the following example:
                                   int keystep);
 
 Register a new command in the Redis server, that will be handled by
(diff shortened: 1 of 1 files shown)


itamarhaber:
Hi @edufelipe 

Thanks for spotting this gap in the docs. Please note, however, that the API's reference is automatically generated from the inline comments of https://github.com/redis/redis/blob/unstable/src/module.c. Please close this PR and create a new one with the same fix against that file.

edufelipe:
@itamarhaber Thanks for the help! Submitted to redis repo.

PR #1564, opened 2021-05-01
fix modules-cluster api description for RedisModule_GetClusterNodesList

iakkus:
* replace incorrect `REDISMODULE_NODE_ID_SIZE` with the correct `REDISMODULE_NODE_ID_LEN` in the cluster api reference for modules.
* fix the call to `RedisModule_Log()` with the correct arguments in the example.

### DIFF
--- a/topics/modules-api-ref.md
+++ b/topics/modules-api-ref.md
@@ -3156,7 +3156,7 @@ known cluster node, `REDISMODULE_ERR` is returned.
     char **RedisModule_GetClusterNodesList(RedisModuleCtx *ctx, size_t *numnodes);
 
 Return an array of string pointers, each string pointer points to a cluster
-node ID of exactly `REDISMODULE_NODE_ID_SIZE` bytes (without any null term).
+node ID of exactly `REDISMODULE_NODE_ID_LEN` bytes (without any null term).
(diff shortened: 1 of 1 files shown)


itamarhaber:
Hello @iakkus and thanks for your first contribution to the project!

Please note that the API's documentation is automatically generated from the [`module.c`](https://github.com/redis/redis/blob/unstable/src/module.c) file. Please make a PR against that file and close this one.

iakkus:
I totally missed that. Will do, thanks!

itamarhaber:
No worries, this is a repeated "mistake", which means it should be addressed better. I wonder if a GitHub issue template would alleviate this. Or maybe a big non-displayed header in the HTML about it. Or maybe both.

itamarhaber:
Ah. The header is already there :)

iakkus:
Yeah, I think perhaps the bigger reason is that the auto-generated API documentation and manually written tutorials/readme files are in the same folder. As a result, it becomes easy to mistake one for the other. I need to check how many of such auto-generated files are in this repo, but perhaps a separation might reduce such "mistakes" in the future.

itamarhaber:
This one's the only one here :)

PR #1565, opened 2021-05-05
Update SET documentation for NX and GET

fabianeichinger:
Merge after https://github.com/redis/redis/pull/8906

## ~~WIP~~
~~I think it would also make sense to clear up the segment "Return value" as it is currently ambiguous whether `EX`/`NX` or `GET` have priority on what's returned.~~


yoav-steinberg (changes requested):
I think you're right, would be good to add a subsection in the `return` section detailing behavior when `GET` is passed.

fabianeichinger:
@yoav-steinberg I think I'm done here. Would love to see this reviewed.

oranagra (approved):
LGTM.
@itamarhaber care to take a look?

yoav-steinberg (approved):
:+1:

PR #1569, opened 2021-05-11
Rephrase description of XX and NX options of ZADD

sazzad16:
Reference https://stackoverflow.com/questions/67486957/syntax-error-when-making-redis-call-from-lua

> That "Always" was confusing to me, should probably be "Only"


zuiderkwast (approved):
I agree. This is much better.

PR #1571, opened 2021-05-17
Use 'nil' in LPOS command docs

sazzad16:


zuiderkwast (approved):
`nil` is used in command docs but "null" is used on the protocol page. So, at least `nil` is consistent with other commands here.

Should we change the protocol page to `nil`? It also needs to be updated to RESP3...

zuiderkwast:
Thanks @sazzad16!

sazzad16:
@zuiderkwast Pleasure!

PR #1572, opened 2021-05-17
MEMORY USAGE command may return 'nil'

sazzad16:


zuiderkwast (approved):
LGTM

("nil" is used in the docs of commands like GET, but "null" is used on the protocol page.)

madolson (approved):

PR #1573, opened 2021-05-18
Remove <CR> from protocol as it doesn't work

ktsiligkiris:
I have tested this in redis in a linux server, and adding <CR> in the
protocol file to pipe to redis-cli --pipe just doesn't work. It only
works with "normal" line endings of <LF> only.

However the examples with printf require \r to work. Why the behavior
of redis-cli --pipe is different?

### DIFF
--- a/topics/mass-insert.md
+++ b/topics/mass-insert.md
@@ -66,28 +66,32 @@ The Redis protocol is extremely simple to generate and parse, and is
 the goal of mass insertion you don't need to understand every detail of the
 protocol, but just that every command is represented in the following way:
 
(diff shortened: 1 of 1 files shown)


zuiderkwast:
The redis protocol (RESP) is using `\r\n` as a delimiter. See https://redis.io/topics/protocol. Maybe you're using a client which translates `\n` to `\r\n` automatically? Did you run `redis-cli` directly in a terminal, in shell script or something else?

If you think there is a bug, please file a bug report at https://github.com/redis/redis/issues with detailed steps how to reproduce the bug.

PR #1575, opened 2021-05-24
Update client-side-caching.md

mashun4ek:
Nit picking: use plural form.


itamarhaber (approved):
Not a nit imo, many (plural) thanks @mashun4ek.

PR #1576, opened 2021-05-25
ZREMRANGEBYLEX : Replace "return" with "remove"

Skn0tt:
I find that "remove" better suits the actual meaning here, as `ZRANGEBYLEX` doesn't actually return any values.


itamarhaber (approved):
Thanks @Skn0tt - merging.

PR #1579, opened 2021-05-29
Add missing flag nofailover in "cluster nodes" command output.

huangzhw:
`nofailover` flag  not updated.


zuiderkwast (approved):
Good.

For reference, `nofailover` has been there since 2018 and is handled here: https://github.com/redis/redis/blob/unstable/src/cluster.c#L244

PR #1586, opened 2021-06-14
Adding ecredis - Erlang Redis Cluster client.

NikolaBorisov:
Adding a new redis cluster client in erlang.

Some key features
* uses eredis as underlying implementation
* allows connections to multiple clusters at the same time
* queries don't get handled on the cluster-gen-server process they get directed to right eredis client.

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -76,6 +76,14 @@
     "description": "Wrapper around eredis providing process pools and consistent hashing.",
     "authors": ["jeremyong"]
   },
+  {
+    "name": "ecredis",
+    "language": "Erlang",
+    "repository": "https://github.com/HalloAppInc/ecredis",
+    "description": "Redis Cluster client that allows for connections to multiple clusters. Queries are send directly to eredis clients allowing for large throughput.",
+    "authors": ["HalloAppInc"]
+  },
+
 
   {
     "name": "exredis",


zuiderkwast (approved):
This looks interesting! (I'm maintaining Nordix/eredis and Nordix/eredis_cluster with @bjosv. I'll have a look at ecredis later.)

I think we should also remove some of the ancient Erlang clients from this list of clients, the ones that haven't been maintained for the last 5 years. Feel free...

zuiderkwast:
It can be mentioned (here or in the README of ecredis itself) that ecredis uses one connection per cluster node, as opposed to eredis_cluster which has a connection pool for each node.

NikolaBorisov:
> It can be mentioned (here or in the README of ecredis itself) that ecredis uses one connection per cluster node, as opposed to eredis_cluster which has a connection pool for each node.

@zuiderkwast Thanks for the feedback. I will update the README of ecredis itself. Should we use Nordix/eredis instead of wooga/eredis. What are the major improvements?

zuiderkwast:
> What are the major improvements?

TLS support, AUTH support, TCP error handling corrections and other fixes. See https://github.com/Nordix/eredis/blob/master/CHANGELOG.md for details. We have also officially taken over Hex package from wooga, so Nordix is the official continuation of the project.

NikolaBorisov:
Ok, I will try to switch ecredis to Nordix.

zuiderkwast:
@NikolaBorisov I think your client should be added to redis.io website either way. The maintainers will merge this sooner or later. (I can't.)

zuiderkwast:
> The maintainers will merge this sooner or later. (I can't.)

I can now. :-)

PR #1587, opened 2021-06-15
Add redis-interval-module

ogama:
Add the redis-interval-module repository in modules.json

### DIFF
--- a/modules.json
+++ b/modules.json
@@ -386,6 +386,15 @@
        "chayim"
     ],
     "stars":0
-  }
-
+  },
+  {
+    "name": "redis-interval-module",
+    "license": "MIT",
+    "repository": "https://github.com/ogama/redis-interval-module",
+    "description": "Redis module for creation and manipulation of interval set.",
+    "authors": [
+      "ogama"
+    ],
+    "stars": 3
+  }	
 ]


zuiderkwast (approved):
Nice!

For reference, this module started after this discussion: https://github.com/redis/redis/issues/8368

modules.json
```diff
@@ -386,6 +386,15 @@
...
     ],
     "stars":0
-  }
-
+  },
+  {
+    "name": "redis-interval-module",
+    "license": "MIT",
+    "repository": "https://github.com/ogama/redis-interval-module",
+    "description": "Redis module for creation and manipulation of interval set.",
```
madolson, line 394: interval sets*

zuiderkwast:
You decided to use `] ... [` for open intervals, which is fine, but typically Redis commands use `( ... )` for open intervals. Some commands like [`XRANGE`](https://redis.io/commands/xrange) use `(` as a prefix to a number to indicate an open interval.

I just wanted to point this out. I will still merge this PR. :-)

modules.json
zuiderkwast: I agree, but it's not very important, so I decided to merge as-is.

madolson: Sure, I agree

PR #1590, opened 2021-06-24
Improve wording around expire + transact

madolson:
Clarifying transaction documentation after we fixed the expire bug.

### DIFF
--- a/topics/transactions.md
+++ b/topics/transactions.md
@@ -194,13 +194,17 @@ there's no need to repeat the operation.
 
 So what is `WATCH` really about? It is a command that will
 make the `EXEC` conditional: we are asking Redis to perform
-the transaction only if none of the `WATCH`ed keys were modified.
-(But they might be changed by the same client inside the transaction
-without aborting it. [More on this](https://github.com/redis/redis-doc/issues/734).)
-Otherwise the transaction is not entered at
-all. (Note that if you `WATCH` a volatile key and Redis expires
-the key after you `WATCH`ed it, `EXEC` will still work. [More on
-this](http://code.google.com/p/redis/issues/detail?id=270).)
+the transaction only if none of the `WATCH`ed keys were modified. This includes
+modifications made by the client, like writes commands, and by Redis itself,
+like expiration or eviction. If keys were modified between when they were
+`WATCH`ed and when the `EXEC` was recieved, the entire transaction will be aborted
+instead.
+
+**NOTE**
+* In Redis versions before 6.0.9, an expired key would not cause a transaction
+to be aborted. [More on this](https://github.com/redis/redis/pull/7920))
+* Commands executed within the transaction are allowed without aborting the 
+transaction. [More on this](https://github.com/redis/redis-doc/issues/734))
 
 `WATCH` can be called multiple times. Simply all the `WATCH` calls will
 have the effects to watch for changes starting from the call, up to
--- a/topics/transactions.md
+++ b/topics/transactions.md
@@ -203,8 +203,8 @@ instead.
 **NOTE**
 * In Redis versions before 6.0.9, an expired key would not cause a transaction
 to be aborted. [More on this](https://github.com/redis/redis/pull/7920)
-* Commands executed within the transaction are allowed without aborting the 
-transaction. [More on this](https://github.com/redis/redis-doc/issues/734)
+* Commands within a transaction wont trigger the `WATCH` condition since they
+are only queued until the `EXEC` is sent.
 
 `WATCH` can be called multiple times. Simply all the `WATCH` calls will
 have the effects to watch for changes starting from the call, up to
--- a/topics/transactions.md
+++ b/topics/transactions.md
@@ -202,9 +202,9 @@ instead.
 
 **NOTE**
 * In Redis versions before 6.0.9, an expired key would not cause a transaction
-to be aborted. [More on this](https://github.com/redis/redis/pull/7920))
+to be aborted. [More on this](https://github.com/redis/redis/pull/7920)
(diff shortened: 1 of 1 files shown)


topics/transactions.md
```diff
@@ -194,13 +194,17 @@ there's no need to repeat the operation.
...
-this](http://code.google.com/p/redis/issues/detail?id=270).)
+the transaction only if none of the `WATCH`ed keys were modified. This includes
+modifications made by the client, like write commands, and by Redis itself,
+like expiration or eviction. If keys were modified between when they were
+`WATCH`ed and when the `EXEC` was received, the entire transaction will be aborted
+instead.
+
+**NOTE**
+* In Redis versions before 6.0.9, an expired key would not cause a transaction
+to be aborted. [More on this](https://github.com/redis/redis/pull/7920)
```
zuiderkwast, line 205: Good!

zuiderkwast (changes requested):
Nice!

Are we going to add another note when redis/redis#9068 is solved too? A note for that would looks something like

* In Redis versions before x.y.z, if a watched key expires between MULTI and EXEC, it does not cause a transaction to be aborted. [More on this](https://github.com/redis/redis/issues/9068)

topics/transactions.md
```diff
@@ -194,13 +194,17 @@ there's no need to repeat the operation.
...
+* Commands executed within the transaction are allowed without aborting the 
+transaction. [More on this](https://github.com/redis/redis-doc/issues/734)
```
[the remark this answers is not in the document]
madolson: Sure, we can drop the read more about #734, but I'll keep the clarification.

madolson:
Yeah, I would add another note for redis/redis#9068, it seems like an important caveat.

zuiderkwast (approved):

topics/transactions.md
```diff
@@ -194,13 +194,17 @@ there's no need to repeat the operation.
...
+the transaction only if none of the `WATCH`ed keys were modified. This includes
+modifications made by the client, like write commands, and by Redis itself,
+like expiration or eviction. If keys were modified between when they were
+`WATCH`ed and when the `EXEC` was received, the entire transaction will be aborted
+instead.
+
+**NOTE**
+* In Redis versions before 6.0.9, an expired key would not cause a transaction
+to be aborted. [More on this](https://github.com/redis/redis/pull/7920)
+* Commands within a transaction wont trigger the `WATCH` condition since they
```
oranagra, line 206: shouldn't it be `won't` instead of `wont`?

btw, please note the styling guides in the readme, try to avoid manual word wrap (and put each sentence in a line of it's own), this can cause a small change to a single sentence to affect multiple lines or even the whole paragraph, and is harder to review.

zuiderkwast: You're right, it should be won't.

Line breaks: I thought the important thing is to not reflow text as that's what causes a large diff, but ok, I will comment on this next time. :)

oranagra: there are two things:
one is not to reflow (not to break the line at a random point due to length and continue on the next line), since that would cause the next edit that added ore removed two words to move some words from one one line to the other.

the second is to start each sentence in a line of it's own, so lines are short, and when a line is changed, there's not a lot of words changed (easier to diff the two lines and realize what's changed).
i.e. break lines manually at `.`, or `,`. and if it's too long we can even break it at some (no-random) "logical" place, which will increase the chance that a future edit will either affect one half or the other but not both.

zuiderkwast: I understand these points. I just wasn't sure how strictly you'd want to enforce them. My first prio was to get the things in which have been waiting for years, but I'll start enforcing these guidelines from now on.

oranagra: you're right. we don't need to enforce it strictly, the "code" base is already full of these overflows, we can just gradually get rid of them to make future changes easier to review.
I just saw an opportunity to comment and make sure we understand / agree on the guidelines.

madolson: I was aware of the guidelines, but the rest of the file doesn't do it so I left it as is. I really don't like "best effort" guidelines, if we want to fix it's easy to write a script to fix them all.

oranagra: I wouldn't mind merging such a commit (if we cam make sure it didn't screw things up), but I rather not invest my time on such effort. 
I don't mind adding new blocks with overflows, or editing old ones without reformatting them. 

Bottom line.. I don't particularly mind about any of that...
Was just trying to make reviewing of modifications easier... 
I hope we agree that such a style would be better, if not feel free to argue.

madolson: I 100% agree that your proposed style is better :)

zuiderkwast: How long did we have these guidelines? If it's a long time, then they're probably not intuitive since nobody seems to follow them (at least not in the PRs I've seen lately). What if we just forbid reflowing and don't care about the rest? I think reflowing the only thing that makes a diff hard to read.

oranagra: exactly 2 months today..
most PRs are probably from before they were written down.
and even if not, existing contributors to the repo don't bother re-reading the readme file before every commit 8-).
i suppose that's one of the reasons why we need to nudge people a lot in the near future, and maybe less in the far future.

the other reason if obviously that people write new docs to look similar to existing lines in the file they edit or file they copy from... and if we make a massive refactory commit it'll help. i just don't think it is a good use of our time to write such script...
all 3 of us better use more time in the redis repo than this one.

zuiderkwast: > Yeah, I would add another note for [redis/redis#9068](https://github.com/redis/redis/issues/9068), it seems like an important caveat.

@madolson For reference, see https://github.com/redis/redis/pull/9194#issuecomment-876380545

PR #1592, opened 2021-06-27
Some minor fix in cluster

huangzhw:
- Fix redis-cli reference instead of redis-trib;
- redis-cli use multiple keys MIGRATE instead of single key MIGRATE;
- One failover target error in cluster spec.

### DIFF
--- a/topics/cluster-spec.md
+++ b/topics/cluster-spec.md
@@ -470,7 +470,7 @@ is that:
 * All queries about non-existing keys in A are processed by "B", because "A" will redirect clients to "B".
 
 This way we no longer create new keys in "A".
-In the meantime, a special program called `redis-cli` used during reshardings
+In the meantime, `redis-cli` used during reshardings
 and Redis Cluster configuration will migrate existing keys in
 hash slot 8 from A to B.
 This is performed using the following command:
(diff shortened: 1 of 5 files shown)


zuiderkwast (approved):

topics/cluster-spec.md
```diff
@@ -470,20 +470,20 @@ is that:
 * All queries about non-existing keys in A are processed by "B", because "A" will redirect clients to "B".
 
 This way we no longer create new keys in "A".
-In the meantime, a special program called `redis-trib` used during reshardings
+In the meantime, a special program called `redis-cli` used during reshardings
 and Redis Cluster configuration will migrate existing keys in
 hash slot 8 from A to B.
```
zuiderkwast, line 475: Is `redis-cli` a "special" program? I think it is the normal CLI client shipped with Redis...

zuiderkwast (commented):
It looks good to replace redis-trib with redis-cli.

I thread is related: https://stackoverflow.com/questions/52880996/redis-trib-rb-is-not-longer-available-but-redis-cli-cluster-create-throws-unre

topics/cluster-spec.md
huangzhw: Remove it.

PR #1593, opened 2021-06-28
New client: redis-async

lukepalmer:
Add a new client: redis-async for ocaml.
Document at dependency on the ruby gem 'batch'


zuiderkwast (approved):

zuiderkwast:
Thanks @lukepalmer!

PR #1596, opened 2021-07-03
Change args of BITPOS from [start] [end] to [start [end]]

huangzhw:
BITPOS key bit [start [end]] instead of BITPOS key bit [start] [end].

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -288,14 +288,20 @@
         "type": "integer"
       },
       {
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
LGTM

zuiderkwast:
Merged and looks on on redis.io. :100:

PR #1601, opened 2021-07-20
Render lists properly in CLIENT PAUSE doc

sazzad16:
Without the blank line, all elements in the list is rendered as a paragraph.

### DIFF
--- a/commands/client-pause.md
+++ b/commands/client-pause.md
@@ -7,10 +7,12 @@ The command performs the following actions:
 * When the specified amount of time has elapsed, all the clients are unblocked: this will trigger the processing of all the commands accumulated in the query buffer of every client during the pause.
 
 Client pause currently supports two modes:
+
 * `ALL`: This is the default mode. All client commands are blocked.
 * `WRITE`: Clients are only blocked if they attempt to execute a write command.
 
 For the `WRITE` mode, some commands have special behavior:
+
 * `EVAL`/`EVALSHA`: Will block client for all scripts.
 * `PUBLISH`: Will block client.
 * `PFCOUNT`: Will block client.


sazzad16:
image

itamarhaber (approved):

oranagra:
@itamarhaber looks like this PR broke the spell checker CI for other PRs, not sure why.
see #1613

madolson:
Sad, it rendered locally in my VSCode window. I guess that is something check out for, it looks like spell checker doesn't correctly decompile sections that aren't valid markdown: https://github.com/redis/redis-doc/pull/1618

oranagra:
that's bad (that the spell checker skips things that are invalid markdown),
but what's worse is that it didn't detect that problem in the CI of this PR, and only detected after it was merged.

p.s. if it is possible to detect these invalid markdown issues, it would be good to:
1. scan the entire repo and fix these.
2. add a CI step that errors when a new one is introduced. (i added one in #1609 and realized that only after merging)

i was under the impression that it's not invalid, just produced a different result than what i meant. but if the spell checker is able to detect that as invalid, then ideally it would have also warned us.

anyway, i don't have time to deal with that, but if someone else has, that's welcome..

oranagra:
btw, anyone already know #1613 got hit by that merge? it was branched before this was merged and appeared ok, and then got the error when i edited one of the files in GH web UI, so i assume the CI runs on the merged result and not on the raw PR.
but now, after the fix was merged, i re-triggered the CI, and it still fails.

zuiderkwast:
Actually there was a an error in the spellchecker job in this PR which I noticed just after merging it. Sorry for that.

oranagra:
ohh, that's good to know.. at least we know we can rely on it.

oranagra:
p.s. i eventually had to add another commit to the other PR to cause it to notice that the master branch has changed so that the CI passes.

PR #1604, opened 2021-07-21
Clarifies that timeout is in seconds.

simonprickett:


yoav-steinberg (changes requested):
This is good.
Can you clarify that the value can have a decimal point to specify fractions of seconds as well. See `BLPOP`'s documentation.

simonprickett:
> This is good.
> Can you clarify that the value can have a decimal point to specify fractions of seconds as well. See `BLPOP`'s documentation.

I've had a go at making this line up more with `BLPOP` thanks!

yoav-steinberg (approved):
:+1:

PR #1613, opened 2021-07-26
Add doc for new options of expire commands

sunng87:
Doc changes for feature https://github.com/redis/redis/pull/2795

### DIFF
--- a/commands/expire.md
+++ b/commands/expire.md
@@ -40,6 +40,9 @@ The `EXPIRE` command supports a set of options since Redis 7.0:
 * `GT` -- Set expiry only when the new expiry is greater than current one
 * `LT` -- Set expiry only when the new expiry is less than current one
 
+A non-volatile key is treated as an infinite TTL for the purpose of `GT` and `LT`.
+The `GT`, `LT` and `NX` options are mutually exclusive.
+
 ## Refreshing expires
 
 It is possible to call `EXPIRE` using as argument a key that already has an
@@ -62,7 +65,8 @@ are now fixed.
 @integer-reply, specifically:
 
 * `1` if the timeout was set.
-* `0` if `key` does not exist or the expiry was not applied due to provided option.
+* `0` if the timeout was not set. e.g. key doesn't exist, or operation skipped
+  due to the provided arguments.
 
 @examples
 
@@ -77,6 +81,10 @@ TTL mykey
 EXPIRE mykey 10 NX
 TTL mykey
 ```
+
+@history
+
+* `>= 7.0`: Added options: `NX`, `XX`, `GT` and `LT`.
 
 ## Pattern: Navigation session
 
--- a/commands/expire.md
+++ b/commands/expire.md
@@ -65,8 +65,7 @@ are now fixed.
 @integer-reply, specifically:
 
 * `1` if the timeout was set.
-* `0` if the timeout was not set. e.g. key doesn't exist, or operation skipped
-  due to the provided arguments.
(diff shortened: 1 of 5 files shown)


oranagra:
@sunng87 thank you.
please also update commands.json (that's what generates the command syntax at the top of each page).
and please add a history section for each of these at the bottom (see the docs for SET command).

to avoid repeated complex description, we can fill the details only in one command and refer the others to it (e.g. see how https://redis.io/commands/zrangestore is documented).
unlike ZRANGE, i think that in this case, it is ok to repeat the description since it's not long or complex.
but in anyway, the history and return value section should be present in each.

oranagra:
p.s. maybe the GT and LT ones, should mention that a non-volatile key is equivalent to an infinite TTL?

yoav-steinberg (changes requested):
Looks good.
Missing update to `EXPIRE` commands in https://github.com/redis/redis-doc/blob/master/commands.json

sunng87:
The content is a little bit repetitive but I feel each page should have sufficient information for user to know all the detail.

yoav-steinberg (approved):
lgtm

commands/expire.md
```diff
@@ -31,6 +31,18 @@ will be `del`, not `expired`).
...
+## Options
+
+The `EXPIRE` command supports a set of options since Redis 7.0:
+
+* `NX` -- Set expiry only when the key has no expiry
+* `XX` -- Set expiry only when the key has an existing expiry
+* `GT` -- Set expiry only when the new expiry is greater than current one
+* `LT` -- Set expiry only when the new expiry is less than current one
+
+A non-volatile key is treated as an infinite ttl.
```
oranagra, line 43: maybe TTL should be upper case? (being an acronym)

oranagra: and maybe it would be clearer if we add "for the purpose of GT / LT"?
i.e.
```
A non-volatile key is treated as an infinite TTL for the purpose of GT and LT.
```

oranagra (approved):

commands/pexpireat.md
```diff
@@ -1,12 +1,24 @@
...
+A non-volatile key is treated as an infinite TTL for the purpose of `GT` and `LT`.
+The `GT`, `LT` and `NX` options are mutually exclusive.
+
 @return
 
 @integer-reply, specifically:
 
 * `1` if the timeout was set.
-* `0` if `key` does not exist.
+* `0` if `key` does not exist or the expiry was not applied due to provided option.
```
madolson, line 21: Would this be clearer as?
```suggestion
* `0` if the timeout was not set.
```

I don't know if we need to enumerate the different ways that can happen.

oranagra: i think it can save some confusion.
maybe we want to split it to format part and then examples that make it clear? like:
```suggestion
- * `0` if `key` does not exist or the expiry was not applied due to provided option.
+ * `0` if the timeout was not set. e.g. key doesn't exist, or operation skipped due to the provided arguments.
```

sunng87: I would recommend to provide such examples for why timeout could not be set.

oranagra (approved):
i removed some line breaks..
our new styling guidelines state that we should avoid breaking lines at arbitrary positions of length, only on periods or stops.
this is intended to make it easier to review future changes.

i we don't refactor existing code though..

oranagra (approved):

sunng87:
Thank you @oranagra

PR #1617, opened 2021-08-02
Add current_cow_peak field in INFO command

ShooterIT:
In https://github.com/redis/redis/pull/8974, we add `current_cow_peak field` in INFO command


yoav-steinberg (approved):
:+1:

PR #1621, opened 2021-08-08
Add SORT_RO documentation

eduardobr:
In addition, corrected use of hyphen in "read-only" for similar GEORADIUS doc

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -3796,7 +3796,7 @@
   },
   "SORT": {
     "summary": "Sort the elements in a list, set or sorted set",
-    "complexity": "O(N+M*log(M)) where N is the number of elements in the list or set to sort, and M the number of returned elements. When the elements are not sorted, complexity is currently O(N) as there is a copy step that will be avoided in next releases.",
+    "complexity": "O(N+M*log(M)) where N is the number of elements in the list or set to sort, and M the number of returned elements. When the elements are not sorted, complexity is O(N).",
     "arguments": [
       {
         "name": "key",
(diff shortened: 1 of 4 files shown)


oranagra:
@itamarhaber @yoav-steinberg i see that GEORADIUS_RO isn't mentioned in commands.json, however, EVAL_RO is.
Do you have an opinion about that?

yoav-steinberg:
This is the current status as far as I can see:
commnad | commands.json | dedicated page | documented in r/w page
---|---|---|---
`BITFIED_RO` | | |
`SORT_RO` | | |
`EVAL_RO` | :heavy_check_mark: | :heavy_check_mark: |
`EVALSHA_RO` | :heavy_check_mark: | :heavy_check_mark: |
`GEORADIUS_RO` | | | :heavy_check_mark:
`GEORADIUSBYMEMBER_RO` | | | :heavy_check_mark:

My opinion is that:
1) We should aim for these commands to **be documented**.
2) We should aim for these commands to be documented **the same way**.
3) The way I prefer is the `EVAL_RO` approach where the command appears in commands.json and has its own page. Note that it's also visible via the "related commands" in `EVAL`.
4) Ideally this PR should do that, or at a minimum do it for `SORT_RO`.

oranagra:
thanks @yoav-steinberg , i completely forgot about BITFIELD_RO myself.
GEORADIUS_RO is actually deprecated now, we prefer to send new use cases to GEOSEARCH, but i suppose that as long as the command exists as a command, it's better to list it.

@eduardobr wanna modify this PR to follow the EVAL_RO approach?

eduardobr:
@oranagra Sure

eduardobr:
In current SORT definition from commands.json:
"When the elements are not sorted, complexity is currently O(N) as there is a copy step that will be avoided in next releases."

@oranagra 
Which next releases? Isn't it O(N) anymore?

oranagra:
> In current SORT definition from commands.json:
"When the elements are not sorted, complexity is currently O(N) as there is a copy step that will be avoided in next releases."
> @oranagra
Which next releases? Isn't it O(N) anymore?

interesting... i see it was added in edb18d07
The related changes i see after that time in redis are:
https://github.com/redis/redis/commit/9a914a632d
https://github.com/redis/redis/commit/c908774b9e
https://github.com/redis/redis/commit/96abf65900

But that doesn't change fact the elements are still copied (twice) on their way to the output buffers.
@itamarhaber @yoav-steinberg @yossigo are you aware of anything else about that topic?
can we just remove the reference to "next releases" and leave it "O(N)"?

yoav-steinberg:
> can we just remove the reference to "next releases" and leave it "O(N)"?

I'm for removing the "next releases".

oranagra:
Just to be clear.. We're not removing just the "next releases" because it was already done, but rather because we don't think it was done or could be done.. So the complexity when not sorting is O(N).

oranagra (approved):
LGTM

yoav-steinberg (approved):
:+1:

PR #1622, opened 2021-08-09
Avoid confusing use of term "pure function"

### LINKED ISSUE #693: Confusing use of the term "pure function"

zuiderkwast:
Also cleaned up and clarified other text in the
replication section of the EVAL command.

Co-Authored-By: Viktor Söderqvist <viktor.soderqvist@est.tech>

Replaces #695. Fixes #693.

### DIFF
--- a/commands/eval.md
+++ b/commands/eval.md
@@ -338,18 +338,18 @@ SCRIPT currently accepts three different commands:
     not violate the scripting engine's guaranteed atomicity).
     See the next sections for more information about long running scripts.
 
-## Scripts as pure functions
+## Scripts with deterministic writes
 
 *Note: starting with Redis 5, scripts are always replicated as effects and not sending the script verbatim. So the following section is mostly applicable to Redis version 4 or older.*
 
-A very important part of scripting is writing scripts that are pure functions.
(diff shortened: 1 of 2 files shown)


zuiderkwast:
This is just a rebase of #695 with the merge conflicts solved.

PR #1631, opened 2021-08-19
Remove cli tag from GETBIT

avimredis:
redis.io page was showing interactive cli for GETBIT which is blocked and threw an error in the response. I removed the cli tag and fixed the example to show how it would look to the user


madolson:
Related to https://github.com/redis/redis-doc/pull/1631.

itamarhaber:
This should be resolved via https://github.com/redis/redis-io/pull/250 - thanks for the effort though!

PR #1632, opened 2021-08-19
Remove cli tag from SETBIT

avimredis:
redis.io page was showing interactive cli for SETBIT which is blocked and threw an error in the response. I removed the cli tag and fixed the example to show how it would look to the user

### DIFF
--- a/commands/setbit.md
+++ b/commands/setbit.md
@@ -26,10 +26,13 @@ the same _key_ will not have the allocation overhead.
 
 @examples
 
-```cli
-SETBIT mykey 7 1
-SETBIT mykey 7 0
-GET mykey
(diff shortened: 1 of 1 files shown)


madolson:
I think the real problem is that this: (https://github.com/redis/redis-io/blob/7420e4017d6d1a77bd181e1a0f30ad99c3f95c9a/scripts/generate_interactive_commands.rb#L13) needs to include the bitmap category.

itamarhaber:
Exactly what @madolson said.

itamarhaber:
This should be resolved via https://github.com/redis/redis-io/pull/250 - thanks for the effort though!

PR #1635, opened 2021-08-30
Escape markdown in my_random_value

Rodolffo:
Escape markdown in the string "my_random_value" in the distributed locks page.


enjoy-binbin:
https://redis.io/topics/distlock#correct-implementation-with-a-single-instance
How about use Use `` , like `my_random_value` ?  (this one can wait for others opinion)

image

yoav-steinberg (approved):
Thanks!

PR #1636, opened 2021-08-30
Add new LMPOP and BLMPOP commands.

enjoy-binbin:
WIP: haven't tried local rendering, not sure about the output render.
I am also not very good at long sentences, not sure I have covered everything.

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -420,6 +420,7 @@
         ]
       },
       {
+        "command": "COUNT",
         "name": "count",
         "type": "integer",
         "optional": true
@@ -455,6 +456,7 @@
         ]
       },
       {
+        "command": "COUNT",
         "name": "count",
         "type": "integer",
         "optional": true
--- a/commands.json
+++ b/commands.json
@@ -397,6 +397,72 @@
     "since": "6.2.0",
     "group": "list"
   },
+  "LMPOP": {
+    "summary": "Pop elements from a list",
+    "complexity": "O(N+M) where N is the number of provided keys and M is the number of elements returned.",
+    "arguments": [
+      {
+        "name": "numkeys",
+        "type": "integer"
+      },
+      {
+        "name": "key",
+        "type": "key",
+        "optional": true,
+        "multiple": true
+      },
+      {
+        "name": "where",
+        "type": "enum",
+        "enum": [
+          "LEFT",
+          "RIGHT"
+        ]
+      },
+      {
+        "name": "count",
+        "type": "integer",
+        "optional": true
+      }
+    ],
+    "since": "7.0.0",
+    "group": "list"
+  },
+  "BLMPOP": {
+    "summary": "Pop elements from a list, or block until one is available",
+    "complexity": "O(N+M) where N is the number of provided keys and M is the number of elements returned.",
+    "arguments": [
+      {
+        "name": "timeout",
+        "type": "double"
+      },
+      {
+        "name": "numkeys",
+        "type": "integer"
+      },
+      {
+        "name": "key",
+        "type": "key",
+        "optional": true,
+        "multiple": true
+      },
+      {
+        "name": "where",
+        "type": "enum",
+        "enum": [
+          "LEFT",
+          "RIGHT"
+        ]
+      },
+      {
+        "name": "count",
+        "type": "integer",
+        "optional": true
+      }
+    ],
+    "since": "7.0.0",
+    "group": "list"
+  },
   "BZPOPMIN": {
     "summary": "Remove and return the member with the lowest score from one or more sorted sets, or block until one is available",
     "complexity": "O(log(N)) with N being the number of elements in the sorted set.",
--- a/commands/blmpop.md
+++ b/commands/blmpop.md
@@ -11,9 +11,8 @@ See `LMPOP` for more information.
 
 @array-reply: specifically:
 
-* A `nil` multi-bulk when no element could be popped, and timeout is reached.
-* A two-element multi-bulk with the first element being the name of the key where
-  elements was popped, and the second element is a multi-bulk with an array of elements.
+* A `nil` when no element could be popped, and timeout is reached.
+* A two-element array with the first element being the name of the key from which elements where popped, and the second element is an array of elements.
 
 @examples
 
--- a/commands/blmpop.md
+++ b/commands/blmpop.md
@@ -2,9 +2,7 @@
 
 When lists contains elements, this command behaves exactly like `LMPOP`.
 When used inside a `MULTI`/`EXEC` block, this command behaves exactly like `LMPOP`.
-When lists are empty, Redis will block the connection until another client
-pushes to it or until `timeout` (a double value specifying the maximum number
-of seconds to block) is reached.
+When lists are empty, Redis will block the connection until another client pushes to it or until `timeout` (a double value specifying the maximum number of seconds to block) is reached.
 A `timeout` of zero can be used to block indefinitely.
 
 See `LMPOP` for more information.
--- a/commands/blmpop.md
+++ b/commands/blmpop.md
@@ -1,8 +1,8 @@
 `BLMPOP` is the blocking variant of `LMPOP`.
 
-When lists contains elements, this command behaves exactly like `LMPOP`.
+When any of the lists contains elements, this command behaves exactly like `LMPOP`.
 When used inside a `MULTI`/`EXEC` block, this command behaves exactly like `LMPOP`.
-When lists are empty, Redis will block the connection until another client pushes to it or until `timeout` (a double value specifying the maximum number of seconds to block) is reached.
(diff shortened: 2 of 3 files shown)


commands/blmpop.md
```diff
@@ -0,0 +1,33 @@
...
+
+See `LMPOP` for more information.
+
+@return
+
+@array-reply: specifically:
+
+* A `nil` multi-bulk when no element could be popped, and timeout is reached.
+* A two-element multi-bulk with the first element being the name of the key where
+  elements was popped, and the second element is a multi-bulk with an array of elements.
```
oranagra, line 18:
```suggestion
* A `nil` when no element could be popped, and timeout is reached.
* A two-element array with the first element being the name of the key from which elements where popped, and the second element is an array of elements.
```

commands/blmpop.md
```diff
@@ -0,0 +1,33 @@
+`BLMPOP` is the blocking variant of `LMPOP`.
+
+When lists contains elements, this command behaves exactly like `LMPOP`.
+When used inside a `MULTI`/`EXEC` block, this command behaves exactly like `LMPOP`.
+When lists are empty, Redis will block the connection until another client
+pushes to it or until `timeout` (a double value specifying the maximum number
+of seconds to block) is reached.
```
oranagra, line 7: our new style guides (see readme) ask that we don't split sentences to multiple lines.
this is intended to make diffs smaller and code review easier.
i know it is confusing since most of the existing docs are written differently.
```suggestion
When lists are empty, Redis will block the connection until another client pushes to it or until `timeout` (a double value specifying the maximum number of seconds to block) is reached.
```

commands/lmpop.md
```diff
@@ -0,0 +1,34 @@
+`LPOP` or `RPOP` takes one key, and can return multiple elements.
+`BLPOP` or `BRPOP` takes multiple keys, but returns one element from just one key.
+`LMPOP` or `BLMPOP` can take multiple keys and return multiple elements from just one key.
```
oranagra, line 3:
```suggestion
Pops one ore more elements from the first non-empty list key from the list of provided key names.

Similar other commands have limitations:
- `LPOP` or `RPOP` takes one key, and can return multiple elements.
- `BLPOP` or `BRPOP` takes multiple keys, but returns one element from just one key.

See `BLMPOP` for the blocking variant of this command.
```

commands/lmpop.md
```diff
@@ -0,0 +1,34 @@
...
+The number of returned elements will consist of up to `count`(default 1),
+depending on the list's length.
+
+@return
+
+@array-reply: specifically:
+
+* A `nil` multi-bulk when no element could be popped.
+* A two-element multi-bulk with the first element being the name of the key where
+  elements was popped, and the second element is a multi-bulk with an array of elements.
```
oranagra, line 17:
```suggestion
* A `nil` when no element could be popped.
* A two-element array with the first element being the name of the key from which elements where popped, and the second element is an array of elements.
```

oranagra (commented):
@yoav-steinberg please take it from here..

commands/blmpop.md
enjoy-binbin: i did take a look with the new guides and the existing code also confusing me...
anyway, i realize this now, and I will deal with it this way in the future

commands/lmpop.md
```diff
@@ -0,0 +1,34 @@
+Pops one ore more elements from the first non-empty list key from the list of provided key names.
+
+Similar other commands have limitations:
```
yoav-steinberg, line 3:
```suggestion
LMPOP and BLMPOP deprecate the following commands:
```

commands/lmpop.md
```diff
@@ -0,0 +1,34 @@
...
+- `LPOP` or `RPOP` takes one key, and can return multiple elements.
```
yoav-steinberg, line 4:
```suggestion
- `LPOP` or `RPOP` which take only one key, and can return multiple elements.
```

commands/lmpop.md
```diff
@@ -0,0 +1,34 @@
...
+- `BLPOP` or `BRPOP` takes multiple keys, but returns one element from just one key.
```
yoav-steinberg, line 5:
```suggestion
- `BLPOP` or `BRPOP` which take multiple keys, but return only one element from just one key.
```

commands/lmpop.md
```diff
@@ -0,0 +1,34 @@
...
+
+See `BLMPOP` for the blocking variant of this command.
+
+Search the list from left to right, find the first non-empty list, return the corresponding list key name, pop from the left or right depending on the argument given, and return the elements.
+The number of returned elements will be limited to `count`(default 1) and the list length.
```
yoav-steinberg, line 10: Elements are popped from either the left or right of the first non-empty list based on the `where` argument. Number of returned elements is limited to the lower between the non-empty list's length and the `count` argument (which defaults to 1).

commands/blmpop.md
```diff
@@ -0,0 +1,30 @@
+`BLMPOP` is the blocking variant of `LMPOP`.
+
+When lists contains elements, this command behaves exactly like `LMPOP`.
```
yoav-steinberg, line 3:
```suggestion
When any of the lists contains elements, this command behaves exactly like `LMPOP`.
```

commands/blmpop.md
```diff
@@ -0,0 +1,30 @@
...
+When used inside a `MULTI`/`EXEC` block, this command behaves exactly like `LMPOP`.
+When lists are empty, Redis will block the connection until another client pushes to it or until `timeout` (a double value specifying the maximum number of seconds to block) is reached.
```
yoav-steinberg, line 5:
```suggestion
When all lists are empty, Redis will block the connection until another client pushes to it or until the `timeout` (a double value specifying the maximum number of seconds to block) elapses.
```

commands/blmpop.md
```diff
@@ -0,0 +1,30 @@
...
+A `timeout` of zero can be used to block indefinitely.
+
+See `LMPOP` for more information.
+
+@return
+
+@array-reply: specifically:
+
+* A `nil` when no element could be popped, and timeout is reached.
+* A two-element array with the first element being the name of the key from which elements where popped, and the second element is an array of elements.
```
yoav-steinberg, line 15:
```suggestion
* A two-element array with the first element being the name of the key from which elements were popped, and the second element is an array of elements.
```

commands/lmpop.md
```diff
@@ -0,0 +1,34 @@
...
+
+@return
+
+@array-reply: specifically:
+
+* A `nil` when no element could be popped.
+* A two-element array with the first element being the name of the key from which elements where popped, and the second element is an array of elements.
```
yoav-steinberg, line 17:
```suggestion
* A two-element array with the first element being the name of the key from which elements were popped, and the second element is an array of elements.
```

yoav-steinberg (changes requested):
Added some suggestions.
Also need to add deprecation notices for `B[RL]POP` and `[RL]POP` commands.
(assuming we're really deprecating them). If not please let me know and I'll fix my suggestions accordingly.
See question here: https://github.com/redis/redis/pull/9373#issuecomment-909050639

commands/lmpop.md
enjoy-binbin: About the `where` arg
Oran said this in the conversation above: I don't think we wanna refer to where IIRC the argument name isn't shown.

commands.json
```diff
@@ -397,6 +397,72 @@
...
+      {
+        "name": "where",
+        "type": "enum",
+        "enum": [
+          "LEFT",
+          "RIGHT"
+        ]
+      },
+      {
+        "name": "count",
```
yoav-steinberg, line 458: I think we need to explicitly add the `COUNT` keyword before the count, so this should be
```suggestion
        "command": "COUNT",
        "name": "count",
```

commands.json
```diff
@@ -397,6 +397,72 @@
...
```
yoav-steinberg, line 423: I think we need to explicitly add the `COUNT` keyword before the count, so this should be
```suggestion
        "command": "COUNT",
        "name": "count",
```

yoav-steinberg (changes requested):

commands/lmpop.md
yoav-steinberg: Elements are popped from either the left or right of the first non-empty list based on the passed argument. The number of returned elements is limited to the lower between the non-empty list's length and the `count` argument (which defaults to 1).

commands/lmpop.md
yoav-steinberg: @enjoy-binbin removed the deprecation mention after we agreed we're not deprecating the simple commands. Please review the suggestion above.

yoav-steinberg (approved):

commands/lmpop.md
enjoy-binbin: Took it! thanks for the all effort

PR #1641, opened 2021-09-09
Update MONITOR command

bjosv:
Updating `MONITOR` docs to match that some commands are now included in the output.
Changed since 6.2.4 via https://github.com/redis/redis/pull/8859


zuiderkwast (approved):
LGTM

bjosv:
The command `EXEC` has been tested to be logged by `MONITOR` in following versions:
`5.0.13`, `6.0.15`, `6.2.5`

using manual tests like: `MULTI`, `SET a 1`, `SET b 2`, `EXEC`

zuiderkwast:
I tested on 3.0.0 just now. (2.0.0 didn't but I didn't try very hard.) EXEC is logged, so I think we can assume it has always been logged.

PR #1642, opened 2021-09-09
Add ASKING command

zuiderkwast:


huangzhw:
If we add this, shuold we add `restore-asking`?

itamarhaber (approved):
LGTM

itamarhaber:
Coincenetaly, I just mapped some of the missing commands in case it matters:

BITFIELD_RO: not found in commands.json
SUBSTR: not found in commands.json <- deprecated since 1.0 iirc
REPLCONF: not found in commands.json
RESTORE-ASKING: not found in commands.json
ASKING: not found in commands.json <-- that's yours!
GEORADIUS_RO: not found in commands.json
GEORADIUSBYMEMBER_RO: not found in commands.json
PFSELFTEST: not found in commands.json <- perhaps internal
PFDEBUG: not found in commands.json <- same
XSETID: not found in commands.json <- definitely internal for replication, undoc intentionally
POST: not found in commands.json <- HTTP DDoS
HOST:: not found in commands.json <- HTTP DDoS

zuiderkwast:
I actually needed to look up ASKING when I implemented -ASK redirects in a client (eredis_cluster, Erlang) a year ago. (I was wondering if you prefix ASKING to another command in the same multibulk (like how you invoke Unix commands like `nice`, `time` and `gdb`) or if it's a separate command.)

Restore-asking is probably for debugging. I can't see why any client would need it. Do you?

For the others, I suppose the `_RO` commands (BITFIELD_RO, GEO*_RO) are useful to add.

huangzhw:
`RESTORE-ASKING` is used in `MIGRATE` in cluster. I think it is used for internal purpose.

huangzhw:
For GEO*_RO, they are deprecated by `GEOSEARCH`, so maybe we should not document them.

PR #1648, opened 2021-09-13
Escapes keywords that are also commands

itamarhaber:


yoav-steinberg (approved):

commands/migrate.md
```diff
@@ -48,9 +48,9 @@ uses pipelining in order to migrate multiple keys between instances without
 incurring in the round trip time latency and other overheads that there are
 when moving each key with a single `MIGRATE` call.
 
-In order to enable this form, the `KEYS` option is used, and the normal *key*
+In order to enable this form, the `!KEYS` option is used, and the normal *key*
```
zuiderkwast, line 51: So will this be rendered as `KEYS` in the website? That must be a non-standard markdown escape...

yoav-steinberg: Not really standard, but then again the auto linking of commands isn't a markdown standard either. The logic can be found here: 
https://github.com/redis/redis-io/blob/5c17c2beaf40558d0645c15f7628b7944fe32fab/lib/template.rb#L26-L39

zuiderkwast: Thanks @yoav-steinberg! Very nice. That documentation isn't easy to find. :-) (The same can be said about the module.c to markdown script. You have to read the source code to see what it's doing.)

zuiderkwast (approved):

PR #1653, opened 2021-09-20
Remove the description about String 512MB limit

hayashier:
512 MB max value limit was removed. https://github.com/redis/redis/issues/757
Maybe proto-max-bulk-len was introduced with Redis 4.0.7. So this limit seems to be removed at that time.

So I updated this description.

### DIFF
--- a/topics/data-types-intro.md
+++ b/topics/data-types-intro.md
@@ -55,7 +55,6 @@ A few other rules about keys:
 * Try to stick with a schema. For instance "object-type:id" is a good
   idea, as in "user:1000". Dots or dashes are often used for multi-word
   fields, as in "comment:1234:reply.to" or "comment:1234:reply-to".
-* The maximum allowed key size is 512 MB.
 
 <a name="strings"></a>
 Redis Strings
@@ -83,7 +82,7 @@ already stored into the key, in the case that the key already exists, even if
 the key is associated with a non-string value. So `SET` performs an assignment.
 
 Values can be strings (including binary data) of every kind, for instance you
-can store a jpeg image inside a value. A value can't be bigger than 512 MB.
+can store a jpeg image inside a value.
 
 The `SET` command has interesting options, that are provided as additional
 arguments. For example, I may ask `SET` to fail if the key already exists,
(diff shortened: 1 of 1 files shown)


itamarhaber:
Hello @hayashier - thanks for this.

The 512MB limit has indeed been relaxed, but it is still the default behavior. To support bigger strings, one would first need to configure (`proto-max-bulk-len` directive) the server accordingly, so I'd rather have a clear explanation of that rather than deleting the relevant sentences.

Also, if you look through the docs, you'll find other references to the 512MB limit (e.g. `SETBIT`, `SETRANGE` and topics/protocol) so this would be a good chance to update all with this information.

hayashier:
Hello itamarhaber,

I agree. I found the following parts which describe 512MB limit by using grep command.
And I agree we should explain about `proto-max-bulk-len` directive.

1. https://redis.io/commands/client-setname
> There is no limit to the length of the name that can be assigned if not the usual limits of the Redis string type (512 MB). 

2. https://redis.io/commands/setbit
> The offset argument is required to be greater than or equal to 0, and smaller than 232 (this limits bitmaps to 512MB). When the string at key is grown, added bits are set to 0.

> Warning: When setting the last possible bit (offset equal to 232 -1) and the string value stored at key does not yet hold a string value, or holds a small string value, Redis needs to allocate all intermediate memory which can block the server for some time. On a 2010 MacBook Pro, setting bit number 232 -1 (512MB allocation) takes ~300ms, setting bit number 230 -1 (128MB allocation) takes ~80ms, setting bit number 228 -1 (32MB allocation) takes ~30ms and setting bit number 226 -1 (8MB allocation) takes ~8ms. 

3. https://redis.io/commands/setrange
> Note that the maximum offset that you can set is 229 -1 (536870911), as Redis Strings are limited to 512 megabytes. If you need to grow beyond this size, you can use multiple keys.

> On a 2010 MacBook Pro, setting byte number 536870911 (512MB allocation) takes ~300ms, setting byte number 134217728 (128MB allocation) takes ~80ms, setting bit number 33554432 (32MB allocation) takes ~30ms and setting bit number 8388608 (8MB allocation) takes ~8ms.

4. https://redis.io/topics/data-types
> A String value can be at max 512 Megabytes in length.

5. https://redis.io/topics/modules-api-ref
> The function returns REDISMODULE_OK on success, and REDISMODULE_ERR on error, that is, the key is not open for writing, is not a string or resizing for more than 512 MB is requested.

6. https://redis.io/topics/protocol
> Bulk Strings are used in order to represent a single binary safe string up to 512 MB in length.

PR #1654, opened 2021-09-20
Add new ZMPOP and BZMPOP commands.

enjoy-binbin:
https://github.com/redis/redis/pull/9484

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -544,7 +544,6 @@
       {
         "name": "key",
         "type": "key",
-        "optional": true,
         "multiple": true
       },
       {
@@ -4628,7 +4627,6 @@
       {
         "name": "key",
         "type": "key",
-        "optional": true,
         "multiple": true
       },
       {
--- a/commands.json
+++ b/commands.json
@@ -529,6 +529,42 @@
     "since": "5.0.0",
     "group": "sorted_set"
   },
+  "BZMPOP": {
+    "summary": "Remove and return members with scores in a sorted set or block until one is available",
+    "complexity": "O(K) + O(N*log(M)) where K is the number of provided keys, N being the number of elements in the sorted set, and M being the number of elements popped.",
+    "arguments": [
+      {
+        "name": "timeout",
+        "type": "double"
+      },
+      {
+        "name": "numkeys",
+        "type": "integer"
+      },
+      {
+        "name": "key",
+        "type": "key",
+        "optional": true,
+        "multiple": true
+      },
+      {
+        "name": "where",
+        "type": "enum",
+        "enum": [
+          "MIN",
+          "MAX"
+        ]
+      },
+      {
+        "command": "COUNT",
+        "name": "count",
+        "type": "integer",
+        "optional": true
+      }
+    ],
+    "since": "7.0.0",
+    "group": "sorted_set"
+  },
   "CLIENT CACHING": {
     "summary": "Instruct the server about tracking or not keys in the next request",
     "complexity": "O(1)",
@@ -4579,6 +4615,38 @@
       }
     ],
     "since": "5.0.0",
+    "group": "sorted_set"
+  },
+  "ZMPOP": {
+    "summary": "Remove and return members with scores in a sorted set",
+    "complexity": "O(K) + O(N*log(M)) where K is the number of provided keys, N being the number of elements in the sorted set, and M being the number of elements popped.",
+    "arguments": [
+      {
+        "name": "numkeys",
+        "type": "integer"
+      },
+      {
+        "name": "key",
+        "type": "key",
+        "optional": true,
+        "multiple": true
+      },
+      {
+        "name": "where",
+        "type": "enum",
+        "enum": [
+          "MIN",
+          "MAX"
+        ]
+      },
+      {
+        "command": "COUNT",
+        "name": "count",
+        "type": "integer",
+        "optional": true
+      }
+    ],
+    "since": "7.0.0",
     "group": "sorted_set"
   },
   "ZRANDMEMBER": {
--- a/commands/bzmpop.md
+++ b/commands/bzmpop.md
@@ -14,17 +14,3 @@ See `ZMPOP` for more information.
 * A `nil` when no element could be popped.
 * A two-element array with the first element being the name of the key from which elements were popped, and the second element is an array of the popped elements. Every entry in the elements array is also an array that contains the member and its score.
 
-@examples
-
-```cli
-DEL myzset myzset2
-ZADD myzset 1 "one" 2 "two" 3 "three"
-BZMPOP 1 1 myzset MIN COUNT 2
-ZRANGE myzset 0 -1 WITHSCORES
-ZADD myzset2 4 "four" 5 "five" 6 "six"
-BZMPOP 1 2 myzset myzset2 MIN COUNT 1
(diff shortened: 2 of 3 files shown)


commands/zmpop.md
```diff
@@ -0,0 +1,33 @@
+Pops one or more elements(member / score pair) from the first non-empty sorted set key from the list of provided key names.
```
itamarhaber, line 1:
```suggestion
Pops one or more elements, that are member-score pairs, from the first non-empty sorted set in the provided list of key names.
```

commands/zmpop.md
```diff
@@ -0,0 +1,33 @@
...
+
+Elements are popped from either the min or max of the first non-empty sorted set based on the passed argument.
+The number of returned elements is limited to the lower between the non-empty sorted set's length, and the count argument (which defaults to 1).
+
+@return
+
+@array-reply: specifically:
+
+* A `nil` when no element could be popped.
+* A two-element array with the first element being the name of the key from which elements were popped, and the second element is an array of elements, and each element is also an array that contains the member and score.
```
itamarhaber, line 17:
```suggestion
* A two-element array with the first element being the name of the key from which elements were popped, and the second element is an array of the popped elements. Every entry in the elements array is also an array that contains the member and its score.
```

commands/bzmpop.md
```diff
@@ -0,0 +1,30 @@
...
+A `timeout` of zero can be used to block indefinitely.
+
+See `ZMPOP` for more information.
+
+@return
+
+@array-reply: specifically:
+
+* A `nil` when no element could be popped.
+* A two-element array with the first element being the name of the key from which elements were popped, and the second element is an array of elements, and each element is also an array that contains the member and score.
```
itamarhaber, line 15:
```suggestion
* A two-element array with the first element being the name of the key from which elements were popped, and the second element is an array of the popped elements. Every entry in the elements array is also an array that contains the member and its score.
```

commands/bzmpop.md
```diff
@@ -0,0 +1,30 @@
...
+ZADD myzset 1 "one" 2 "two" 3 "three"
+BZMPOP 1 1 myzset MIN COUNT 2
+ZRANGE myzset 0 -1 WITHSCORES
+ZADD myzset2 4 "four" 5 "five" 6 "six"
+ZMPOP 2 myzset myzset2 MIN COUNT 1
+ZRANGE myzset 0 -1 WITHSCORES
+ZMPOP 2 myzset myzset2 MAX count 10
+ZRANGE myzset2 0 -1 WITHSCORES
+EXISTS myzset myzset2
+```
```
itamarhaber, line 30: No need to repeat the same example. Also, since this is a blocking command it won't be able to really run in the interactive website, so the example needs to be static (i.e. not tagged w/ the `cli` thing).

commands/zmpop.md
```diff
@@ -0,0 +1,33 @@
...
+
+@examples
+
+```cli
+ZMPOP 2 non1 non2 MIN COUNT 10
```
itamarhaber, line 22: I think it is best to start as simple as possible.

```suggestion
ZMPOP 1 notsuchkey MIN
```

commands/bzmpop.md
```diff
@@ -0,0 +1,30 @@
+`BZMPOP` is the blocking variant of `ZMPOP`.
+
+When any of the sorted sets contains elements, this command behaves exactly like `ZMPOP`.
+When used inside a `MULTI`/`EXEC` block, this command behaves exactly like `ZMPOP`.
+When all sorted sets are empty, Redis will block the connection until another client adds to it or until the `timeout` (a double value specifying the maximum number of seconds to block) elapses.
```
itamarhaber, line 5:
```suggestion
When all sorted sets are empty, Redis will block the connection until another client adds members to one of the keys or until the `timeout` (a double value specifying the maximum number of seconds to block) elapses.
```

commands/bzmpop.md
enjoy-binbin: oh... my bad. I actually want to perform the `BZMPOP` in non-blocking path... (due to the copy-paste bug)

oranagra:
@itamarhaber i noticed this isn't merged yet (already in unstable long ago), please make sure to merge it.

commands/bzmpop.md
```diff
@@ -0,0 +1,30 @@
...
+ZADD myzset 1 "one" 2 "two" 3 "three"
+BZMPOP 1 1 myzset MIN COUNT 2
+ZRANGE myzset 0 -1 WITHSCORES
+ZADD myzset2 4 "four" 5 "five" 6 "six"
+BZMPOP 1 2 myzset myzset2 MIN COUNT 1
+ZRANGE myzset 0 -1 WITHSCORES
+BZMPOP 1 2 myzset myzset2 MAX COUNT 10
+ZRANGE myzset2 0 -1 WITHSCORES
+EXISTS myzset myzset2
+```
```
itamarhaber, line 30: A blocking command should be run in the interactive example - removing as the same example exists in zmpop.

```suggestion
```

itamarhaber (approved):

commands.json
```diff
@@ -4581,6 +4617,38 @@
...
+      {
+        "name": "numkeys",
+        "type": "integer"
+      },
+      {
+        "name": "key",
+        "type": "key",
+        "optional": true,
+        "multiple": true
+      },
```
itamarhaber, line 4633: Since the number of keys must be gt 0, the first key is actually mandatory.

```suggestion
      {
        "name": "key",
        "type": "key",
      },
      {
        "name": "key",
        "type": "key",
        "optional": true,
        "multiple": true
      },
```

commands.json
```diff
@@ -529,6 +529,42 @@
...
```
itamarhaber, line 549: Since the number of keys must be gt 0, the first key is actually mandatory.

```suggestion
      {
        "name": "key",
        "type": "key",
      },
      {
        "name": "key",
        "type": "key",
        "optional": true,
        "multiple": true
      },
```

itamarhaber (changes requested):

commands.json
enjoy-binbin: Right... we may also need to fix LMPOP/BLMPOP.  The `optional` one and the BLMPOP examples ?

https://github.com/redis/redis-doc/blob/master/commands.json#L441
https://github.com/redis/redis-doc/blob/master/commands/blmpop.md

commands/zmpop.md
```diff
@@ -0,0 +1,35 @@
+Pops one or more elements, that are member-score pairs, from the first non-empty sorted set in the provided list of key names.
+
+`ZMPOP` and `BZMPOP` are similar to the following, more limited, commands:
```
itamarhaber, line 3: Newline needed to render the list

```suggestion
`ZMPOP` and `BZMPOP` are similar to the following, more limited, commands:

```

itamarhaber (changes requested):

itamarhaber:
lol @enjoy-binbin - you're online and applying my suggestions as I make them :)

enjoy-binbin:
@itamarhaber yes. I happen to be online LOL

itamarhaber:
Yes, indeed, LMPOP/BLMPOP have the same issues and need to be fixed. Do you want to open a PR for that?

itamarhaber (approved):

PR #1662, opened 2021-10-08
Update doc for new cluster command: CLUSTER DELSLOTSRANGE and CLUSTER ADDSLOTSRANGE

hwware:
Upstream: https://github.com/redis/redis/pull/9445

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -832,7 +832,7 @@
   },
   "CLUSTER ADDSLOTSRANGE ": {
     "summary": "Assign new hash slots to receiving node",
-    "complexity": "O(N) where N is the total number of the slots between the start slot and end slot arguments. The complexity is ultimately determined by the sum of multiply slot ranges",
+    "complexity": "O(N) where N is the total number of the slots between the start slot and end slot arguments.",
     "arguments": [
       {
         "name": [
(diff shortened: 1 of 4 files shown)


commands.json
```diff
@@ -830,6 +830,25 @@
     "since": "3.0.0",
     "group": "cluster"
   },
+  "CLUSTER ADDSLOTSRANGE ": {
+    "summary": "Assign new hash slots to receiving node",
+    "complexity": "O(N) where N is the total number of the slots between the start slot and end slot arguments. The complexity is ultimately determined by the sum of multiply slot ranges",
```
madolson, line 835: Let's drop the second sentence here, I'm not sure it really clarifies anything.

commands/cluster-addslotsrange.md
```diff
@@ -0,0 +1,51 @@
+This command is useful in order to modify a node's view of the cluster
```
madolson, line 1: We typically don't duplicate the information between two similar commands. I would include the syntax information here, and then a link to the `ADDSLOTS` subargument.

madolson: I don't think this was addressed. I think we can simplify this page to just talk about how it's different from cluster addslots, since I don't want to duplicate so much context.

hwware:
@madolson Hi Madolson, I simplify the description for these 2 new commands, please take a look, Thanks

itamarhaber (approved):

PR #1667, opened 2021-10-18
example for replicaof command

chayim:


itamarhaber:
Nope, that's not what I'm talking about. Please, open any command page with an example like the one you're trying to add and try typing there... e.g.:

image

chayim:
Added

PR #1668, opened 2021-10-18
Add a third argument to eval.md to avoid confusion

DianaOlympos:
Eval need the number of keys as the second argument. We have seen people thinking it needs to be the number of arguments instead.

This small change of the example hopefully may help people disambiguate and understand that the `2` correspond to the number of keys.


itamarhaber (approved):
Thanks for improving this @DianaOlympos - merging.

PR #1685, opened 2021-11-19
Added redli to list of clients

toote:
As described in the title

### DIFF
--- a/clients.json
+++ b/clients.json
@@ -1964,5 +1964,5 @@
     "description": "Simple Redis Client for Dart",
     "authors": ["cerussite127"]
   }
-  
+
 ]
--- a/clients.json
+++ b/clients.json
@@ -1964,5 +1964,14 @@
     "description": "Simple Redis Client for Dart",
     "authors": ["cerussite127"]
   }
+  
+  {
+    "name": "redli",
+    "language": "Go",
+    "url": "https://github.com/IBM-Cloud/redli",
+    "repository": "https://github.com/IBM-Cloud/redli",
+    "description": "A humane alternative to the Redis-cli and TLS",
+    "authors": ["IBM"]
+  }
 
 ]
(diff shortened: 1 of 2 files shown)


itamarhaber:
Hello @toote 

Thanks for making the PR and I, too, think Redli deserves a toot from redis.io.

That said, Redli isn't a programmatically-approachable client such as those on this list. It should be mentioned along redis-cli (and possibly https://github.com/laixintao/iredis & https://github.com/mylxsw/redis-tui as well). We don't quite have such a place yet though. Maybe the topics/rediscli page?

zuiderkwast:
I think interactive clients can be listed under tools, i.e. "Higher level libraries and tools" on the https://redis.io/clients page. `iredis` is there already. (It's a H1 but there's no link anchor for it. :thinking: )

clients.json
```diff
@@ -1964,5 +1964,14 @@
     "description": "Simple Redis Client for Dart",
     "authors": ["cerussite127"]
   }
+  
+  {
```
zuiderkwast, line 1968: Comma is missing after the previous client. That's why there is a JSON parse error and build failure.

Move this to `tools.json`, then I think we can merge it.

zuiderkwast (changes requested):

tools.json
```diff
@@ -127,6 +127,14 @@
     "description": "Ruby/Redis based recommendation engine (collaborative filtering).",
     "authors": ["paulasmuth"]
   },
+  {
+    "name": "redli",
+    "language": "Go",
+    "url": "https://github.com/IBM-Cloud/redli",
+    "repository": "https://github.com/IBM-Cloud/redli",
```
zuiderkwast, line 134: I don't think there is a benefit adding the same URL twice. People would click on both just to find out they point to the same page.

```suggestion
    "repository": "https://github.com/IBM-Cloud/redli",
```

zuiderkwast (approved):
@itamarhaber I hope you agree this should be merged.

PR #1686, opened 2021-11-20
Add new async_loading field in INFO command

enjoy-binbin:
Added in https://github.com/redis/redis/pull/9323

### DIFF
--- a/commands/info.md
+++ b/commands/info.md
@@ -172,6 +172,7 @@ by referring to the `MEMORY STATS` command and the `MEMORY DOCTOR`.
 Here is the meaning of all fields in the **persistence** section:
 
 *   `loading`: Flag indicating if the load of a dump file is on-going
+*   `async_loading`: To differentiate from the blocking loading - Added in Redis 7.0.
 *   `current_cow_peak`: The peak size in bytes of copy-on-write memory
      while a child fork is running
 *   `current_cow_size`: The size in bytes of copy-on-write memory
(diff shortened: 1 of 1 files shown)


oranagra:
Thanks, indeed e should have documented it and forgot. 
But I think the text should be improved to add more details, I.e. State that in this mode redis is in the process of loading a new dataset, and but commands are allowed and the `loading` flag is off. Also it need sot refer to `repl-diskless-load swapdb`

commands/info.md
```diff
@@ -172,6 +172,7 @@ by referring to the `MEMORY STATS` command and the `MEMORY DOCTOR`.
 Here is the meaning of all fields in the **persistence** section:
 
 *   `loading`: Flag indicating if the load of a dump file is on-going
+*   `async_loading`: To differentiate from the blocking loading - Added in Redis 7.0.
```
yoav-steinberg, line 175:
```suggestion
*   `async_loading`: Currently busy loading replication data-set asynchronously while serving old data. This means `repl-diskless-load` is enabled and set to `swapdb`. Added in Redis 7.0.
```

yoav-steinberg (commented):
@oranagra changed it based on your comments. Let me know if it makes sense.

commands/info.md
oranagra: LTGM, one suggestion: maybe drop the word "busy", (not to be confused with `-BUSY` error).
the text "currently loading..." would be enough..

PR #1693, opened 2021-11-28
Correct XGROUP SETID enum type argument definition in comments.json

rueian:
Hi, I found a little inconsistent in the comments.json. I think `id` argument in the `XGROUP SETID` should be labeled as `type=enum` to align with others.

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -5862,7 +5862,7 @@
       },
       {
         "name": "id",
-        "type": "string",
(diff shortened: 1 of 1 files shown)


itamarhaber (approved):

itamarhaber:
Hi @rueian - indeed seems like an inconsistency, merging & thanks.

PR #1698, opened 2021-12-02
Update for multiparam config set.

yoav-steinberg:
Update `CONFIG SET` docs given https://github.com/redis/redis/pull/9748

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -1222,15 +1222,18 @@
     "group": "server"
   },
   "CONFIG SET": {
-    "summary": "Set a configuration parameter to the given value",
+    "summary": "Set configuration parameters to the given values",
     "arguments": [
       {
-        "name": "parameter",
-        "type": "string"
-      },
-      {
-        "name": "value",
-        "type": "string"
+        "name": [
+          "parameter",
+          "value"
+        ],
+        "type": [
+          "string",
+          "string"
+        ],
+        "multiple": true
       }
     ],
     "since": "2.0.0",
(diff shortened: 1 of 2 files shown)


yoav-steinberg:
Pending for https://github.com/redis/redis/pull/9656 to be merged so we won't override the commands.json modifications in this pr.

itamarhaber (approved):

yoav-steinberg:
@itamarhaber I'd like to merge this bug I'd like you thumbs up on me deleting all that stuff about `save` and byte size notation which seemed irrelevant and obsolete.
Also there's a reference to the v6.0 _redis.conf_ file in this doc which I think we should change to simply reference the latest _redis.conf_ in unstable since our docs are always in sync with _unstable_ and not with a specific version. WDYT?

itamarhaber:
I approved after thinking about the save removal (+1) and trying to find when the kb/mb was added (went as far as 2.2) but couldn't (so another +1). As for the link, makes sense to me - adding it and merging.

(also note that I added a history section for this change :))

itamarhaber:
Coming to rethink about it, maybe we should keep the save note for the sake of users coming with less-than-7 versions.

yoav-steinberg:
> Coming to rethink about it, maybe we should keep the save note for the sake of users coming with less-than-7 versions.

I think there's not reason since the change from pre-7, in that respect, is in the config file not in `config set`. So if you're pre 7 and trying to configure your redis's `save` you should read the relevant comments in `redis.conf`. Perhaps we can add a note about how the documentation in `redis.conf` changes from version to verion? I'll try to push something, let me know what you think.

itamarhaber (approved):
LGTM

PR #1711, opened 2021-12-14
Update SHUTDOWN and Signal Handling

zuiderkwast:
Doc update for redis/redis#9872

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -3912,6 +3912,22 @@
           "SAVE"
         ],
         "optional": true
+      },
+      {
+        "name": "now",
+        "type": "enum",
+        "enum": [
+          "NOW"
+        ],
+        "optional": true
+      },
+      {
+        "name": "force",
+        "type": "enum",
+        "enum": [
+          "FORCE"
+        ],
+        "optional": true
       }
     ],
     "since": "1.0.0",
--- a/commands.json
+++ b/commands.json
@@ -10738,6 +10738,13 @@
         "summary": "Synchronously save the dataset to disk and then shut down the server",
         "since": "1.0.0",
         "group": "server",
+        "complexity": "O(N) when saving, where N is the total number of keys in all databases when saving data, otherwise O(1)",
+        "history": [
+            [
+                "7.0",
+                "Added the `NOW`, `FORCE` and `ABORT` modifiers. Introduced waiting for lagging replicas before exiting."
+            ]
+        ],
         "acl_categories": [
             "@admin",
             "@slow",
--- a/commands/info.md
+++ b/commands/info.md
@@ -72,7 +72,7 @@ Here is the meaning of all fields in the **server** section:
 *   `executable`: The path to the server's executable
 *   `config_file`: The path to the config file
 *   `io_threads_active`: Flag indicating if I/O threads are active
-*   `shutdown_in_seconds`: The maximum time remaining for replicas to catch up the replication before shutting completing the shutdown sequence.
+*   `shutdown_in_seconds`: The maximum time remaining for replicas to catch up the replication before completing the shutdown sequence.
     This field is only present during shutdown.
 
 Here is the meaning of all fields in the **clients** section:
--- a/commands/info.md
+++ b/commands/info.md
@@ -71,6 +71,9 @@ Here is the meaning of all fields in the **server** section:
 *   `lru_clock`: Clock incrementing every minute, for LRU management
 *   `executable`: The path to the server's executable
 *   `config_file`: The path to the config file
+*   `io_threads_active`: Flag indicating if I/O threads are active
+*   `shutdown_in_seconds`: The maximum time remaining for replicas to catch up the replication before shutting completing the shutdown sequence.
+    This field is only present during shutdown.
 
 Here is the meaning of all fields in the **clients** section:
 
@@ -89,7 +92,6 @@ Here is the meaning of all fields in the **clients** section:
      `BRPOP`, `BRPOPLPUSH`, `BLMOVE`, `BZPOPMIN`, `BZPOPMAX`)
 *   `tracking_clients`: Number of clients being tracked (`CLIENT TRACKING`)
 *   `clients_in_timeout_table`: Number of clients in the clients timeout table
-*    `io_threads_active`: Flag indicating if I/O threads are active
 
 Here is the meaning of all fields in the **memory** section:
 
--- a/commands/shutdown.md
+++ b/commands/shutdown.md
@@ -41,15 +41,14 @@ system is in a state that does not allow to safely immediately persist
 on disk.
 
 Normally if there is an AOF child process performing an AOF rewrite, Redis
-will simply kill it and exit. However there are three conditions where it is
-unsafe to do so, and the **SHUTDOWN** command will be refused with an error
-instead. This happens when:
+will simply kill it and exit.
+However, there are situations where it is unsafe to do so and, unless the **FORCE** modifier is specified, the **SHUTDOWN** command will be refused with an error instead.
+This happens in the following situations:
 
 * The user just turned on AOF, and the server triggered the first AOF rewrite in order to create the initial AOF file. In this context, stopping will result in losing the dataset at all: once restarted, the server will potentially have AOF enabled without having any AOF file at all.
 * A replica with AOF enabled, reconnected with its master, performed a full resynchronization, and restarted the AOF file, triggering the initial AOF creation process. In this case not completing the AOF rewrite is dangerous because the latest dataset received from the master would be lost. The new master can actually be even a different instance (if the **REPLICAOF** or **SLAVEOF** command was used in order to reconfigure the replica), so it is important to finish the AOF rewrite and start with the correct data set representing the data set in memory when the server was terminated.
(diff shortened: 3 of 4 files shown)


commands/info.md
```diff
@@ -71,6 +71,9 @@ Here is the meaning of all fields in the **server** section:
 *   `lru_clock`: Clock incrementing every minute, for LRU management
 *   `executable`: The path to the server's executable
 *   `config_file`: The path to the config file
+*   `io_threads_active`: Flag indicating if I/O threads are active
+*   `shutdown_in_seconds`: The maximum time remaining for replicas to catch up the replication before shutting completing the shutdown sequence.
```
yoav-steinberg, line 75:
```suggestion
*   `shutdown_in_seconds`: The maximum time remaining for replicas to catch up the replication before completing the shutdown sequence.
```

commands/shutdown.md
```diff
@@ -1,35 +1,40 @@
...
+* If there are any replicas lagging behind in replication:
+  * Pause clients attempting to write by performing a `CLIENT PAUSE` with the `WRITE` option.
+  * Wait up to the configured `shutdown-timeout` (default 10 seconds) for replicas to catch up the replication offset.
 * Stop all the clients.
 * Perform a blocking SAVE if at least one **save point** is configured.
 * Flush the Append Only File if AOF is enabled.
 * Quit the server.
 
 If persistence is enabled this commands makes sure that Redis is switched off
 without the lost of any data.
```
yoav-steinberg, line 12:
```suggestion
without any data loss.
```

commands/shutdown.md
```diff
@@ -1,35 +1,40 @@
...
-This is not guaranteed if the client uses simply `SAVE` and then `QUIT` because
-other clients may alter the DB data between the two commands.
```
yoav-steinberg, line 11: Why did you remove this comment? Because it goes without saying and isn't related to `SHUTDOWN` or for some other reason?

commands/shutdown.md
```diff
@@ -1,35 +1,40 @@
...
-* **SHUTDOWN SAVE** will force a DB saving operation even if no save points are
-  configured.
-* **SHUTDOWN NOSAVE** will prevent a DB saving operation even if one or more
-  save points are configured.
-  (You can think of this variant as an hypothetical **ABORT** command that just
-  stops the server).
+* **SAVE** will force a DB saving operation even if no save points are configured.
+* **NOSAVE** will prevent a DB saving operation even if one or more save points are configured.
+* **NOW** skips waiting for lagging replicas, i.e. it bypasses the first step in the shutdown sequence.
+* **FORCE** ignores any errors that would normally prevent the server from exiting.
```
yoav-steinberg, line 30: Not clear that such "any errors" can be? Maybe add "details below" or reference the next section?

commands/shutdown.md
```diff
@@ -41,11 +46,28 @@ instead. This happens when:
```
yoav-steinberg, line 46: See above:
> there are two conditions

Should now be "three conditions".

commands/shutdown.md
```diff
@@ -41,11 +46,28 @@ instead. This happens when:
...
+The first command will turn off the AOF if needed, and will terminate the AOF rewriting child if there is one active.
+The second command will not have any problem to execute since the AOF is no longer enabled.
 
-There are conditions when we want just to terminate a Redis instance ASAP, regardless of what its content is. In such a case, the right combination of commands is to send a **CONFIG appendonly no** followed by a **SHUTDOWN NOSAVE**. The first command will turn off the AOF if needed, and will terminate the AOF rewriting child if there is one active. The second command will not have any problem to execute since the AOF is no longer enabled.
+## Minimize the risk of data loss
+
+Since Redis 7.0, the server waits for lagging replicas up to a configurable `shutdown-timeout`, by default 10 seconds, before shutting down.
+This provides a best effort minimizing the risk of data loss in a situation where no save points are configured and AOF is disabled.
+Before version 7.0, shutting down a heavily loaded master node in a diskless setup is more likely to result in data loss.
```
yoav-steinberg, line 61:
```suggestion
Before version 7.0, shutting down a heavily loaded master node in a diskless setup was more likely to result in data loss.
```

topics/signals.md
```diff
@@ -22,14 +22,24 @@ executed just after the script is killed, or if terminates spontaneously.
...
 * The server exits with an exit code of zero.
 
-In case the RDB file can't be saved, the shutdown fails, and the server continues to run in order to ensure no data loss. Since Redis 2.6.11 no further attempt to shut down will be made unless a new `SIGTERM` will be received or the `SHUTDOWN` command issued.
+In case the RDB file can't be saved, the shutdown fails, and the server continues to run in order to ensure no data loss.
+Likewise, if the user just turned on AOF, and the server triggered the first AOF rewrite in order to create the initial AOF file and this file can't be saved, the shutdown fails and the server continues to run.
+Since Redis 2.6.11 no further attempt to shut down will be made unless a new `SIGTERM` is received or the `SHUTDOWN` command issued.
+
+Since Redis 7.0, the server waits for lagging replicas up to a configurable `shutdown-timeout`, by default 10 seconds, before shutting down.
+This provides a best effort minimizing the risk of data loss in a situation where no save points are configured and AOF is disabled.
+Before version 7.0, shutting down a heavily loaded master node in a diskless setup is more likely to result in data loss.
```
yoav-steinberg, line 41:
```suggestion
Before version 7.0, shutting down a heavily loaded master node in a diskless setup was more likely to result in data loss.
```

commands/shutdown.md
zuiderkwast: I think the comment was confusing and irrelevevant. QUIT is just closing the client's connection. It's not shutting down the server.

commands.json
```diff
@@ -3912,6 +3912,22 @@
           "SAVE"
         ],
         "optional": true
+      },
+      {
+        "name": "now",
```
zuiderkwast, line 3917: IIUC, this file is generated from redis and I should not touch this file. I'll revert it. Is that right?

commands/shutdown.md
zuiderkwast: Right. I was thinking about users who continue to run old versions. For them it "is" still the case, but if we use "Before" we should also use "was". Another way to describe it would be "In versions prior to 7.0, ... is ....", but I like "before" because it's simpler language.

commands.json
yoav-steinberg: As long as we're still using this version of the docs/redis-io (which doesn't use the generated file from the redis repo) we should keep both the redis repo and the redis-doc repo manually in sycn. So don't revert.

yoav-steinberg (approved):

commands/shutdown.md
itamarhaber: "several"?

commands/shutdown.md
```diff
@@ -1,51 +1,74 @@
...
 
 @return
 
 @simple-string-reply on error.
 On success nothing is returned since the server quits and the connection is
 closed.
+
+@history
+
+* `>= 7.0`: NOW and FORCE modifiers added.
```
itamarhaber, line 73:
```suggestion
* `>= 7.0.0`: `NOW` and `FORCE `modifiers added.
```

itamarhaber (approved):

commands/shutdown.md
zuiderkwast: "all of the conditions below"... Many possibilitites

zuiderkwast: Looking at this again, it doesn't seem like conditions that conditions have to be fulfilled. It's rather two different *situations* where exit is prevented. Maybe it should instead be sth like

"However there are two situations where it is unsafe to do so, and unless the **FORCE** modifier is specified, the **SHUTDOWN** command will be refused with an error instead. This happens when:"

WDYT?

commands/shutdown.md
```diff
@@ -70,5 +70,5 @@ closed.
 
 @history
 
-* `>= 7.0`: NOW and FORCE modifiers added.
+* `>= 7.0.0`: `NOW` and `FORCE `modifiers added.
```
yoav-steinberg, line 73:
```suggestion
* `>= 7.0.0`: `NOW` and `FORCE` modifiers added.
```

commands/shutdown.md
zuiderkwast: See commit "conditions -> situations"

commands/shutdown.md
```diff
@@ -1,51 +1,74 @@
...
-On success nothing is returned since the server quits and the connection is
-closed.
+@simple-string-reply: `OK` if `ABORT` was specified and shutdown was aborted.
+On successful shutdown, nothing is returned since the server quits and the connection is closed.
+On failure, an error is returned.
+
+@history
+
+* `>= 7.0.0`: `NOW`, `FORCE` and `ABORT` modifiers added.
+  Waiting for lagging replicas before exiting added.
```
itamarhaber, line 74: History is now recorded in the Redis repo, in each command's src/commands JSON file. Please move this tidbit there.

zuiderkwast: ACK. I'll need to add it in the Redis PR then.

itamarhaber (approved):

PR #1697, opened 2021-12-01
Refresh COMMAND, add topics: command-args, key-specs

guybe7:
Add new COMMAND command features implemented in https://github.com/redis/redis/pull/9656 and it's predecessor PRs (key-specs, and sub-commands)
Plus some other cleanup and improvements.

TODO:
- [x] Document COMMAND LIST
- [x] Write the key-specs topic
- [x] Write the command hints
- [x] Document key-spec-index (new field for command argument)
- [x] update the list of command flags

PR #1714, opened 2021-12-16
Replaces commands.json with new syntax

itamarhaber:
Besides pulling in the new 'commands.json', this also:

* Adds man pages for some pure container commands, such as `CONFIG` and `SCRIPT` (put differently, all containers except `COMMAND`)
* Adds man pages for commands with help, e.g. `CLIENT HELP` and `SCRIPT HELP`
* Adds man pages for system (internal) commands: `DEBUG`, `PFSELFTEST`, `PFDEBUG`, `XSETID`, `REPLCONF` and `RESTORE-ASKING`
* Adds a man page for the deprecated `SUBSTR` command
* Adds man pages for missing (deprecated) _RO commands: `GEORADIUS_RO`, `GEORADIUSBYMEMBER`
* Opportunistically fixes the wrong use of `@example` tags in some `LATENCY` man pages (should be `@examples`)

This goes hand in hand with https://github.com/redis/redis-io/pull/259

- [ ] WIP `COMMAND LIST` (#1697)
- [ ] WIP Functions


commands/substr.md
```diff
@@ -0,0 +1,22 @@
+Returns the substring of the string value stored at `key`, determined by the
```
oranagra, line 1: Missing depreciation notice?

oranagra (commented):
So other than replacing the json file, this PR does:
1. Add entries for container commands (like CONFIG, DEBUG, SCRIPT, etc) 
2. Add the HELP sub commands
3. Rename "example" to "examples"
4. Add SUBSTR, GEORADIUS_RO, 
5. Anything else? 
Please list in the top comment.

commands/substr.md
oranagra: Or maybe it's suppose to be auto generated, and we need to remove it (and other Metadata like history and complexity) from all other files?

itamarhaber:
@oranagra added.

commands/substr.md
itamarhaber: Exactly. I can actually make the change to redis-io to use the new deprecation metadata now, but I want to clear it from the files first. Perhaps we'll leave a deprecation notice in substr just for now?

itamarhaber, line 1:
```suggestion
**Note**: as of Redis 2.0, this command was renamed to `GETRANGE`.

Returns the substring of the string value stored at `key`, determined by the
```

oranagra: if you wish to do it in steps, ok..
but let's not forget.

PR #1726, opened 2021-12-30
Readme cleanup

oranagra:


README.md
```diff
@@ -34,7 +34,9 @@ For example:
 
 ## Commands
 
-Redis commands are described in the `commands.json` file.
+Redis commands are described in the `commands.json` file that is auto generated
+from the redis repo based on the JSON files in the commands folder.
+See: https://github.com/redis/redis/tree/unstable/src/commands
```
itamarhaber, line 39: We don't want peeps to go there. How about
```suggestion
from the output of a Redis server.
See: https://github.com/redis/redis/tree/unstable/src/utils/generate-commands-json.py
```

itamarhaber (approved):

README.md
oranagra: Maybe we need both? 
In the past we referred people to a file they can edit and submit a PR. 
So now, if anyone finds an issue in Redis.io commands, he'll need to submit a PR to redis.

itamarhaber: I guess both makes most sense for now. We can also have issue templates saying something similar.

PR #1728, opened 2022-01-01
Update ping.md

### LINKED ISSUE #1727: Ping command returns either Simple String or Bulk String response based on argument.
Hi, the Ping command returns Simple string +PONG if no argument is given to ping,
else it returns the message as bulk string message.

```
ping server
$6
server
```
Should the return value in the [Ping](https://github.com/redis/redis-doc/blob/master/commands/ping.md) doc be updated ?

Observed this on Redis 6.2.6

crackthecodeabhi:
Ping returns Bulk string when an argument is provided.

This request adds the @bulk-string-reply to the documentation.

Fixes #1727

### DIFF
--- a/commands/ping.md
+++ b/commands/ping.md
@@ -10,9 +10,9 @@ of the argument.
 
 @return
 
-@simple-string-reply, if no argument is provided
+@simple-string-reply, and specifically `PONG`, when no argument is provided.
 
-@bulk-string-reply, copy of the argument if provided
+@bulk-string-reply the argument provided, when applicable.
 
 @examples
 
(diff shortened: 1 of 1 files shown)


commands/ping.md
```diff
@@ -10,7 +10,9 @@ of the argument.
 
 @return
 
-@simple-string-reply
+@simple-string-reply, if no argument is provided
+
+@bulk-string-reply, copy of the argument if provided
```
itamarhaber, line 15:
```suggestion
@simple-string-reply, and specifically `PONG`, when no argument is provided.

@bulk-string-reply the argument provided, when applicable.
```

itamarhaber (approved):

PR #1732, opened 2022-01-05
Add dump_payload_sanitizations field in INFO

enjoy-binbin:
Added in https://github.com/redis/redis/pull/7807

PS: the other changes doesn't break the blame log (look like)
image

> but it also can be removed

### DIFF
--- a/commands/info.md
+++ b/commands/info.md
@@ -291,7 +291,7 @@ Here is the meaning of all fields in the **stats** section:
 *   `total_error_replies`: Total number of issued error replies, that is the sum of
     rejected commands (errors prior command execution) and
     failed commands (errors within the command execution)
-*   `dump_payload_sanitizations`: Total number of deep dump payloads integrity validations.
+*   `dump_payload_sanitizations`: Total number of deep dump payloads integrity validations (see `sanitize-dump-payload` config).
 *   `total_reads_processed`: Total number of read events processed
 *   `total_writes_processed`: Total number of write events processed
 *   `io_threaded_reads_processed`: Number of read events processed by the main and I/O threads
--- a/commands/info.md
+++ b/commands/info.md
@@ -291,10 +291,11 @@ Here is the meaning of all fields in the **stats** section:
 *   `total_error_replies`: Total number of issued error replies, that is the sum of
     rejected commands (errors prior command execution) and
     failed commands (errors within the command execution)
-*    `total_reads_processed`: Total number of read events processed
-*    `total_writes_processed`: Total number of write events processed
-*    `io_threaded_reads_processed`: Number of read events processed by the main and I/O threads
-*    `io_threaded_writes_processed`: Number of write events processed by the main and I/O threads
+*   `dump_payload_sanitizations`: Total number of deep dump payloads integrity validations.
+*   `total_reads_processed`: Total number of read events processed
+*   `total_writes_processed`: Total number of write events processed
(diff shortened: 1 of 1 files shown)


commands/info.md
```diff
@@ -291,10 +291,11 @@ Here is the meaning of all fields in the **stats** section:
 *   `total_error_replies`: Total number of issued error replies, that is the sum of
     rejected commands (errors prior command execution) and
     failed commands (errors within the command execution)
-*    `total_reads_processed`: Total number of read events processed
-*    `total_writes_processed`: Total number of write events processed
-*    `io_threaded_reads_processed`: Number of read events processed by the main and I/O threads
-*    `io_threaded_writes_processed`: Number of write events processed by the main and I/O threads
+*   `dump_payload_sanitizations`: Total number of deep dump payloads integrity validations.
```
oranagra, line 294: maybe we should add some reference to the `sanitize-dump-payload` config so users have a chance to understand what this is about.

enjoy-binbin: yes, good point

Adding this: 
> See `dump_payload_sanitizations` configuration in config file.

or 
> See `dump_payload_sanitizations` configuration in config file for more details.

somethings like this

oranagra:
```suggestion
*   `dump_payload_sanitizations`: Total number of deep dump payloads integrity validations (see `sanitize-dump-payload` config).
```

oranagra: the one with the `_` is the variable.. we're referring to the user config name.

yoav-steinberg:
@oranagra What exactly does `dump_payload_sanitizations` count and why do we need this stat?

oranagra:
in redis 6.2 there's a feature that sanitizes RDB files and RESTORE commands.
it can work in two ways, shallow and deep (scanning listpacks in O(N)).
this counts how many deeps scans it performed.

It is currently disabled by default, but it also has a mode in which it'll be applied only for client connections, and not for RDB file loading.
also ACL has a feature to exempt certain users from this overhead.
there's also a plan that cluster nodes that use MIGRATE won't go through there.

so the metric would count just the non-exempt cases.

yoav-steinberg:
I mean why do we need this, what's the value of it. I'd say it's more interesting to know how many fails deep sanitizations we had not how many sanitization attempts there were.

commands/info.md
```diff
@@ -291,10 +291,11 @@ Here is the meaning of all fields in the **stats** section:
 *   `total_error_replies`: Total number of issued error replies, that is the sum of
     rejected commands (errors prior command execution) and
     failed commands (errors within the command execution)
-*    `total_reads_processed`: Total number of read events processed
-*    `total_writes_processed`: Total number of write events processed
-*    `io_threaded_reads_processed`: Number of read events processed by the main and I/O threads
-*    `io_threaded_writes_processed`: Number of write events processed by the main and I/O threads
+*   `dump_payload_sanitizations`: Total number of deep dump payloads integrity validations (see `sanitize-dump-payload` config).
```
yoav-steinberg, line 294:
```suggestion
*   `dump_payload_sanitizations`: Total number of dump payload deep integrity validations (see `sanitize-dump-payload` config).
```

oranagra:
the metric is there to give some visibility on performance.
could be nice to add another one for the failures. but arguably it is somewhat covered by `INFO ERRORSTAT`

PR #1736, opened 2022-01-10
zrange min->start max ->stop

slorello89:
Accompanying PR to redis/redis#10097 to change `min` -> `start` and `max` -> `stop` in `ZRANGE` see other PR for explanation for change.

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -14807,11 +14807,11 @@
                 "key_spec_index": 0
             },
             {
-                "name": "min",
+                "name": "start",
                 "type": "string"
             },
             {
-                "name": "max",
+                "name": "stop",
                 "type": "string"
             },
             {
--- a/commands/zrange.md
+++ b/commands/zrange.md
@@ -56,7 +56,6 @@ Will return all the elements with `5 < score < 10` (5 and 10 excluded).
 Using the `REV` option reverses the sorted set, with index 0 as the element with the highest score.
 
 By default, `<start>` must be less than or equal to `<stop>` to return anything.
-
 However, if the `BYSCORE`, or `BYLEX` options are selected, the `<start>` is the highest score to consider, and `<stop>` is the lowest score to consider, therefore `<start>` must be greater than or equal to `<stop>` in order to return anything.
 
 For example:
@@ -82,7 +81,7 @@ Note that lexicographical ordering relies on all elements having the same score.
 Valid `<start>` and `<stop>` must start with `(` or `[`, in order to specify
 whether the range interval is exclusive or inclusive, respectively.
 
-The special values of `+` or `-` for `<start>` and `<stop>` mean positive and negative infinite strings, respectively, so for instance the command **ZRANGE myzset - + BYLEX** is guaranteed to return all the elements in the sorted set, providing that all the elements have the same score.
+The special values of `+` or `-` for `<start>` and `<stop>` mean positive and negative infinite strings, respectively, so for instance the command `ZRANGE myzset - + BYLEX` is guaranteed to return all the elements in the sorted set, providing that all the elements have the same score.
 
 The `REV` options reverses the order of the `<start>` and `<stop>` elements, where `<start>` must be lexicographically greater than `<stop>` to produce a non-empty result.
 
--- a/commands/zrange.md
+++ b/commands/zrange.md
@@ -51,6 +51,24 @@ ZRANGE zset (5 (10 BYSCORE
 
 Will return all the elements with `5 < score < 10` (5 and 10 excluded).
 
+## Rev ranges
+
+Using the `REV` option reverses the sorted set, with index 0 as the element with the highest score. In this case, `<start>` must be less than or equal to `<stop>` to return anything. However, if the `BYSCORE` option is selected, the `<start>` is the highest score to consider, and `<stop>` is the lowest score to consider, therefore `<start>` must be less than or equal to `<stop>`
+
+For example:
+
+```
+ZRANGE zset 5 10 REV
+```
+
+Will return the elements between index 5 and 10 in the reversed index.
+
+```
+ZRANGE zset 10 5 REV BYSCORE
+```
+
+Will return all elements with scores less than 10 and greater than 5.
+
 ## Lexicographical ranges
 
 When the `BYLEX` option is used, the command behaves like `ZRANGEBYLEX` and returns the range of elements from the sorted set between the `<start>` and `<stop>` lexicographical closed range intervals.
--- a/commands/zrange.md
+++ b/commands/zrange.md
@@ -51,9 +51,13 @@ ZRANGE zset (5 (10 BYSCORE
 
 Will return all the elements with `5 < score < 10` (5 and 10 excluded).
 
-## Rev ranges
+## Reverse ranges
 
-Using the `REV` option reverses the sorted set, with index 0 as the element with the highest score. In this case, `<start>` must be less than or equal to `<stop>` to return anything. However, if the `BYSCORE` option is selected, the `<start>` is the highest score to consider, and `<stop>` is the lowest score to consider, therefore `<start>` must be less than or equal to `<stop>`
+Using the `REV` option reverses the sorted set, with index 0 as the element with the highest score.
+
+By default, `<start>` must be less than or equal to `<stop>` to return anything.
+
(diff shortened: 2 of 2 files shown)


slorello89:
@itamarhaber

commands.json
```diff
@@ -14807,11 +14807,11 @@
                 "key_spec_index": 0
             },
             {
-                "name": "min",
+                "name": "start",
```
zuiderkwast, line 14810: These JSON changes need to be done also in src/commands/zrange.json in the Redis main repository.

commands/zrange.md
```diff
@@ -17,23 +17,23 @@ The optional `WITHSCORES` argument supplements the command's reply with the scor
...
 
 The indexes can also be negative numbers indicating offsets from the end of the sorted set, with `-1` being the last element of the sorted set, `-2` the penultimate element, and so on.
 
 Out of range indexes do not produce an error.
 
-If `<min>` is greater than either the end index of the sorted set or `<max>`, an empty list is returned.
+If `<start>` is greater than either the end index of the sorted set or `<stop>`, an empty list is returned.
 
-If `<max>` is greater than the end index of the sorted set, Redis will use the last element of the sorted set.
+If `<stop>` is greater than the end index of the sorted set, Redis will use the last element of the sorted set.
```
zuiderkwast, line 28: I think we should update this explanation so it is correct also with REV.

commands/zrange.md
```diff
@@ -53,14 +53,14 @@ Will return all the elements with `5 < score < 10` (5 and 10 excluded).
...
+When the `BYLEX` option is used, the command behaves like `ZRANGEBYLEX` and returns the range of elements from the sorted set between the `<start>` and `<stop>` lexicographical closed range intervals.
 
 Note that lexicographical ordering relies on all elements having the same score. The reply is unspecified when the elements have different scores.
 
-Valid `<min>` and `<max>` must start with `(` or `[`, in order to specify
+Valid `<start>` and `<stop>` must start with `(` or `[`, in order to specify
 whether the range interval is exclusive or inclusive, respectively.
 
-The special values of `+` or `-` `<min>` and `<max>` mean positive and negative infinite strings, respectively, so for instance the command **ZRANGEBYLEX myzset - +** is guaranteed to return all the elements in the sorted set, providing that all the elements have the same score.
+The special values of `+` or `-` `<start>` and `<stop>` mean positive and negative infinite strings, respectively, so for instance the command **ZRANGEBYLEX myzset - +** is guaranteed to return all the elements in the sorted set, providing that all the elements have the same score.
```
zuiderkwast, line 63: Missing "for" in this sentence. (Not new.)

```suggestion
The special values of `+` or `-` for `<start>` and `<stop>` mean positive and negative infinite strings, respectively, so for instance the command **ZRANGEBYLEX myzset - +** is guaranteed to return all the elements in the sorted set, providing that all the elements have the same score.
```

zuiderkwast (commented):
Good idea.

Is it explained somewhere that REV needs start=max and stop=min? Maybe add that somewhere...

It would be good to add an example with REV as well.

commands/zrange.md
slorello89: These statements still appear to be accurate, if `BYSCORE` is not used it still min/max would still apply in the same positions, it just reverses the result. Might warrant some further discussion about consistency - I will write up something for the `REV` argument with some example usage.

slorello89:
@zuiderkwast - added another piece about the `REV` argument.

commands/zrange.md
zuiderkwast: Oh, good. It's confusing indeed, but your last commit explains it well.

commands/zrange.md
```diff
@@ -51,16 +51,34 @@ ZRANGE zset (5 (10 BYSCORE
 
 Will return all the elements with `5 < score < 10` (5 and 10 excluded).
 
+## Rev ranges
```
zuiderkwast, line 54:
```suggestion
## Reverse ranges
```

commands/zrange.md
```diff
@@ -51,16 +51,34 @@ ZRANGE zset (5 (10 BYSCORE
...
+
+Using the `REV` option reverses the sorted set, with index 0 as the element with the highest score. In this case, `<start>` must be less than or equal to `<stop>` to return anything. However, if the `BYSCORE` option is selected, the `<start>` is the highest score to consider, and `<stop>` is the lowest score to consider, therefore `<start>` must be less than or equal to `<stop>`
```
zuiderkwast, line 56: Line breaks between every sentense for new text. (See CONTRIBUTING.md in this repo.)

Also `.` is missing in the end.

```suggestion
Using the `REV` option reverses the sorted set, with index 0 as the element with the highest score.
In this case, `<start>` must be less than or equal to `<stop>` to return anything.
However, if the `BYSCORE` option is selected, the `<start>` is the highest score to consider, and `<stop>` is the lowest score to consider, therefore `<start>` must be less than or equal to `<stop>`.
```

zuiderkwast (commented):
The added section about reverse ranges explains it very well.

Question: Does BYLEX REV also reverse the start and stop lexicographic order?

slorello89:
@zuiderkwast - added the bit on `BYLEX` also addressed the couple stylistic things you pointed out. Did I miss something? I don't see a `CONTRIBUTING.md` in this repo.

zuiderkwast:
> @zuiderkwast - added the bit on `BYLEX` also addressed the couple stylistic things you pointed out. Did I miss something? I don't see a `CONTRIBUTING.md` in this repo.

Sorry, it's in the readme: https://github.com/redis/redis-doc#styling-guidelines

No need to add a blank line which will cause a new paragraph. A single linebreak in the markdown is enough and it doesn't cause any linebreak in the rendered HTML. The purpose is just to make the diff smaller when some part of the text is changed, i.e. easier to review changes.

commands/zrange.md
```diff
@@ -51,9 +51,13 @@ ZRANGE zset (5 (10 BYSCORE
...
 
-## Rev ranges
+## Reverse ranges
 
-Using the `REV` option reverses the sorted set, with index 0 as the element with the highest score. In this case, `<start>` must be less than or equal to `<stop>` to return anything. However, if the `BYSCORE` option is selected, the `<start>` is the highest score to consider, and `<stop>` is the lowest score to consider, therefore `<start>` must be less than or equal to `<stop>`
+Using the `REV` option reverses the sorted set, with index 0 as the element with the highest score.
+
+By default, `<start>` must be less than or equal to `<stop>` to return anything.
+
+However, if the `BYSCORE`, or `BYLEX` options are selected, the `<start>` is the highest score to consider, and `<stop>` is the lowest score to consider, therefore `<start>` must be greater than or equal to `<stop>` in order to return anything.
```
zuiderkwast, line 60: No need to add a blank line between them. A blank line causes a new paragraph. A single linebreak does not produce a linebreak in the rendered HTML, but it's only for a smaller diff in future changes.

At least the second and third sentences should be in the same paragraph, I think.

commands/zrange.md
```diff
@@ -78,7 +82,9 @@ Note that lexicographical ordering relies on all elements having the same score.
 Valid `<start>` and `<stop>` must start with `(` or `[`, in order to specify
 whether the range interval is exclusive or inclusive, respectively.
 
-The special values of `+` or `-` `<start>` and `<stop>` mean positive and negative infinite strings, respectively, so for instance the command **ZRANGEBYLEX myzset - +** is guaranteed to return all the elements in the sorted set, providing that all the elements have the same score.
+The special values of `+` or `-` `<start>` and `<stop>` mean positive and negative infinite strings, respectively, so for instance the command **ZRANGE myzset - + BYLEX** is guaranteed to return all the elements in the sorted set, providing that all the elements have the same score.
+
+The `REV` options reverses the order of the `<start>` and `<stop>` elements, where `<start>` must be lexicographically greater than `<stop>` to produce a non-empty result.
```
zuiderkwast, line 87: Good!

zuiderkwast:
The build error is the spellcheck which doesn't like the word BYLEX. If the code example is surrounded by backticks I think it can escape the spell check. Otherwise, we need it to the `wordlist` file.

zuiderkwast (approved):

itamarhaber (approved):

PR #1737, opened 2022-01-11
Use " instead of ` for better readability in protocol.md

enjoy-binbin:
https://redis.io/topics/protocol

image

I would like to enclose it in double quotes, like other places
image

### DIFF
--- a/topics/protocol.md
+++ b/topics/protocol.md
@@ -56,7 +56,7 @@ In RESP, the type of some data depends on the first byte:
 * For **Errors** the first byte of the reply is "-"
 * For **Integers** the first byte of the reply is ":"
 * For **Bulk Strings** the first byte of the reply is "$"
-* For **Arrays** the first byte of the reply is "`*`"
+* For **Arrays** the first byte of the reply is "*"
 
 Additionally RESP is able to represent a Null value using a special variation of Bulk Strings or Array as specified later.
 
@@ -195,7 +195,7 @@ returns elements of a list.
 
 RESP Arrays are sent using the following format:
 
-* A `*` character as the first byte, followed by the number of elements in the array as a decimal number, followed by CRLF.
+* A "*" character as the first byte, followed by the number of elements in the array as a decimal number, followed by CRLF.
 * An additional RESP type for every element of the Array.
 
 So an empty Array is just the following:
@@ -206,7 +206,7 @@ While an array of two RESP Bulk Strings "foo" and "bar" is encoded as:
 
     "*2\r\n$3\r\nfoo\r\n$3\r\nbar\r\n"
 
-As you can see after the `*<count>CRLF` part prefixing the array, the other
+As you can see after the "*<count>CRLF" part prefixing the array, the other
 data types composing the array are just concatenated one after the other.
 For example an Array of three integers is encoded as follows:
 
@@ -226,7 +226,7 @@ integers and a bulk string can be encoded as the follows:
 
 (The reply was split into multiple lines for clarity).
 
-The first line the server sent is `*5\r\n` in order to specify that five
+The first line the server sent is "*5\r\n" in order to specify that five
 replies will follow. Then every reply constituting the items of the
 Multi Bulk reply are transmitted.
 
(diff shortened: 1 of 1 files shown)


topics/protocol.md
```diff
@@ -195,7 +195,7 @@ returns elements of a list.
 
 RESP Arrays are sent using the following format:
 
-* A `*` character as the first byte, followed by the number of elements in the array as a decimal number, followed by CRLF.
+* A "*" character as the first byte, followed by the number of elements in the array as a decimal number, followed by CRLF.
```
zuiderkwast, line 198: The asterisk (`*`) has a special meaning in markdown: emphasis. It will only be converted to emphasis if it's two asterisks in a paragraph. I think it is prevented if it's in a code block (backticks). Examples:

```markdown
Hello *world

Hello *world*

Hello `*world*`
```

Result (GitHub's rendering):

> Hello *world
>
> Hello *world*
>
> Hello `*world*`

----

Backticks prevents the rendering as emphasis. (It's also possible to use backslash escapes according to the original markdown syntax: https://daringfireball.net/projects/markdown/syntax#backslash).

topics/protocol.md
```diff
@@ -235,7 +235,7 @@ specify a Null value (usually the Null Bulk String is used, but for historical
 reasons we have two formats).
 
 For instance when the `BLPOP` command times out, it returns a Null Array
-that has a count of `-1` as in the following example:
+that has a count of "-1" as in the following example:
```
zuiderkwast, line 238: Here -1 is just a number. It's not a string. I think quotes is misleading in this case. I think no formatting at all is OK. Backticks is also OK.

topics/protocol.md
```diff
@@ -306,7 +306,7 @@ The client sends the command **LLEN mylist** in order to get the length of the l
 
     S: :48293\r\n
 
-As usual we separate different parts of the protocol with newlines for simplicity, but the actual interaction is the client sending `*2\r\n$4\r\nLLEN\r\n$6\r\nmylist\r\n` as a whole.
+As usual we separate different parts of the protocol with newlines for simplicity, but the actual interaction is the client sending "*2\r\n$4\r\nLLEN\r\n$6\r\nmylist\r\n" as a whole.
```
zuiderkwast, line 309: Suggestion: Use backticks inside quotes in this and many other RESP strings for even better readability.

```suggestion
As usual we separate different parts of the protocol with newlines for simplicity, but the actual interaction is the client sending "`*2\r\n$4\r\nLLEN\r\n$6\r\nmylist\r\n`" as a whole.
```

zuiderkwast (changes requested):
I agree the quotes improves readability, but I think we should also have backticks for code style formatting, so I suggest we use both quotes and backticks (like ``"`*`"``) to make it even more clear and prevent formatting (like `*` rendered as emphasis).

topics/protocol.md
enjoy-binbin: So you think we should just follow line59?  with ``, the char will look like bold (emphasis)
```
use "`*`"
```

The picture below is how we currently see our website rendering

image

topics/protocol.md
enjoy-binbin: yes. i just wanted to quotes it
image

zuiderkwast: I think quotes is misleading, since this text refers to the number -1, not the string "-1".

topics/protocol.md
zuiderkwast: (Line 59, is the line ``For **Arrays** the first byte of the reply is "`*`"``.)

Yes, I think we can use this style also for "`+`", "`-`", "`:`", "`$`" and other places.

You don't like it?

enjoy-binbin: It is different from other chars in the original rendering

> Yes, I think we can use this style also for "+", "-", ":", "$" and other places.

but now, i can take it :)

topics/protocol.md
enjoy-binbin: right, i decided to just use the number 1

zuiderkwast (approved):
Nice!

PR #1738, opened 2022-01-11
Adding Racompass to tools.json

appsnano:
Adding Racompass to list of Redis tools.
(Racompass created by using Node.js and wrapped with Electron.)
Thanks!


itamarhaber:
Hi @appsnano - looks very impressive but there's no source available license for it, and it's only available from a store. I'm afraid we can't include it here, but let us know if this changes.

PR #1739, opened 2022-01-11
Add RESP3 notice to protocol page

itamarhaber:
ref: https://github.com/antirez/RESP3/issues/48


oranagra:
maybe missing some line break? https://redis.io/topics/protocol

PR #1744, opened 2022-01-18
Remove recommended from  radix as too few stars

liuchong:
There is too few stars on the client module `radix`, is it strange to recommend it? 🤔


itamarhaber:
Hello @liuchong 

The recommendation isn't strictly star-based and is more about the client's functionality/performance/maintenance.

From personal experience, I found radix to be simple and performant, which certainly put it in my "recommended" list :)

p.s. I'm not convinced that stars count is the sole metric for judging, well, anything.

PR #1751, opened 2022-01-25
New structure for: admin, ldb, pubsub, memory-optimization, transaction, pipelining

slorello89:
Moved admin, ldb, pubsub, memory-optimization, transaction, pipelining topics to their new structure, added appropriate headers for hugo, did some light copy editing on them, added new lines for longer lines to make diffs easier and to comply with contributing standards.


slorello89:
For some odd reason, the `git mv` for the pipelining doc did not take and it seems to think the whole thing was deleted and recreated. Any thoughts?

docs/manual/admin.md
```diff
@@ -1,11 +1,16 @@
...
+  - /topics/admin
+---
+
 
 This page contains topics related to the administration of Redis instances.
 Every topic is self contained in form of a FAQ. New topics will be created in the future.
 
-Redis setup hints
------------------
+## Redis setup hints
```
zuiderkwast, line 13: These two are equivalent in markdown. We try to avoid unnecessary changes because it makes the diff larger and messes up the git log and blame information.

docs/manual/lua-debugging.md
```diff
@@ -16,35 +27,43 @@ The Redis Lua debugger, codename LDB, has the following important features:
...
 
 A simple way to get started with the Lua debugger is to watch this video
 introduction:
 
 <iframe width="560" height="315" src="https://www.youtube.com/embed/IMvRfStaoyM" frameborder="0" allowfullscreen></iframe>
 
-**Important note:** please make sure to avoid debugging Lua scripts using your Redis production server. Use a development server instead. Also note that using the synchronous debugging mode (which is NOT the default) results into the Redis server blocking for all the time the debugging session lasts.
+> Important Note:  please make sure to avoid debugging Lua scripts using your Redis production server.
+Use a development server instead.
+Also note that using the synchronous debugging mode (which is NOT the default) results in the Redis server blocking for all the time the debugging session lasts.
```
zuiderkwast, line 39: For new text, we use this style, but please don't reflow old text. It's unnecessary changes.

docs/manual/lua-debugging.md
```diff
@@ -16,35 +27,43 @@ The Redis Lua debugger, codename LDB, has the following important features:
...
-To start a new debugging session using `redis-cli` do the following steps:
+To start a new debugging session using `redis-cli` do the following:
 
 1. Create your script in some file with your preferred editor. Let's assume you are editing your Redis Lua script located at `/tmp/script.lua`.
 2. Start a debugging session with:
 
-    ./redis-cli --ldb --eval /tmp/script.lua
+```
+./redis-cli --ldb --eval /tmp/script.lua
+```
```
zuiderkwast, line 48: Unnecessary changes. The old and new markdown are equivalent.

zuiderkwast (commented):
Is there any particular reason for moving the files?

Regarding git mv, there is actually no moving in git. It is just a delete and add another file. When we use git diff, git log and git show later, it is shown as a moved file if git is smart enough to figure this out. See the `--find-renames` option in `git help show`.

itamarhaber:
@zuiderkwast thanks for reviewing!

Moving to the new site includes both infra and content revamps. Part of the content revamp is **massively** restructuring the contents under /topics to make it approachable and usable. There's no way to do it IMO other than "refactoring" most of it.

zuiderkwast:
@itamarhaber Restructuring, fine, but I can't see how changing markdown to equivalent markdown is good for anything (other than polluting the git blame and causing merge conflicts to any open PRs); line breaks same. If there is a difference in how the markdown is rendered in the new website, fine, then please explain it in the commit message.

That said, it's up to you what you want to merge. :-)

docs/manual/admin.md
banker: Thanks for the feedback and very valid point, @zuiderkwast. It's true that these two heading types are equivalent, but the underline-style headings only support H1 and H2 elements. As part of this organizational effort, we'd like to support H3, and even H4, headings, which means that we'll also eventually have to introduce the "#"-style headings. This sets us up for consistency. Hopefully this is only a minor pain in the short term.

docs/manual/lua-debugging.md
banker: Agreed and fixed. Thanks, @zuiderkwast.

zuiderkwast:
> thanks for reviewing

@itamarhaber FYI I didn't review. I stopped after a few lines. There were too many non-essential changes so I thougt I'll review again when they're reverted.

PR #1752, opened 2022-01-25
redis.io replication, eviction, admin, and persistence edits

kaitlynmichael:
replication.md
eviction.md
admin.md
persistence.md


banker (changes requested):
Thanks for this work, @kaitlynmichael! Looks like the admin page wasn't git mv'd first. Can you fix that?

banker (approved):

PR #1756, opened 2022-01-27
Documentation for ACL V2 + ACL DRYRUN

madolson:
Reference implementation: https://github.com/redis/redis/pull/9974


commands/acl-setuser.md
```diff
@@ -17,6 +17,19 @@ set command (+set rule). Then another SETUSER call can modify the user rules:
...
+Selectors are added by wrapping the rules in parentheses and providing them just like any other rule.
+To verify that a user can execute a command, first the root permissions are checked, and then all of its selectors are checked.
+If any of these rule sets match the command, the command can be executed.
+For example:
+
+    ACL SETUSER virginia on +GET allkeys (+SET ~app1*)
+
+This sets a user with two sets of permission, one defined on the user and one defined with a selector.
+The root user permissions only allows executing the get command, but can be executed on any keys.
+The selector then grants a secondary set of permissions, access to the SET command that be executed on any key that start with app1.
```
zuiderkwast, line 30:
```suggestion
The selector then grants a secondary set of permissions: access to the `SET` command to be executed on any key that starts with "app1".
```

commands/acl-setuser.md
```diff
@@ -40,11 +53,14 @@ The following documentation is a reference manual about the capabilities of this
 
 ## List of rules
 
+Redis ACL rules are split into two categories, rules that define command permissions, "Command rules", and rules that define user state, "User management rules".
```
zuiderkwast, line 56:
```suggestion
Redis ACL rules are split into two categories: rules that define command permissions, "Command rules", and rules that define user state, "User management rules".
```

and possibly, use *emphasis* instead of "quotes" when introducing the new concepts

topics/acl.md
```diff
@@ -396,6 +413,51 @@ Example:
 In the previous section it was observed how it is possible to define commands
 ACLs based on adding/removing single commands.
 
+## Selectors
+
+Redis supports additional sets of rules that are evaluated independently of the main user permissions.
+These secondary sets of permissions are called selectors and added by wrapping a set of rules within parentheses.
+Selectors are evaluated after the main user permissions are checked and are evaluated in the order they are added to the user. 
+A command is allowed if either the root user permissions or any selectors matches a command.
```
zuiderkwast, line 421:
```suggestion
A command is allowed if either the main user permissions or any selectors matches a command.
```

topics/acl.md
```diff
@@ -396,6 +413,51 @@ Example:
...
+
+For example, consider a user with the ACL rules `+GET ~key1 (+SET ~key2)`.
+This user is able to execute `GET key1` and `SET key2 hello`, but not `GET key2` or `SET key1 world`.
+
+Unlike the root user permissions, selectors can not be modified after they are added.
+Selectors can be removed with the `clearselectors` keyword, which removes all of the added selectors.
+Note that `clearselectors` does not remove the root user permissions.
```
zuiderkwast, line 428:
```suggestion
Unlike the main user permissions, selectors cannot be modified after they are added.
Instead, selectors can be removed with the `clearselectors` keyword, which removes all of the added selectors.
Note that `clearselectors` does not remove the main user permissions.
```

topics/acl.md
```diff
@@ -396,6 +413,51 @@ Example:
...
+* R (Read): User supplied data from the key is processed, copied or returned. Note that this does not include metadata such as size information (example `STRLEN`), type information (example `TYPE`) or information about whether a value exists within a collection (example `SISMEMBER`). 
+
+Permissions can be composed together by specifying multiple characters. 
+Specifying the permission as 'RW' is considered full access and is analogous to just passing in `~<pattern>`.
+
+For a concrete example, consider a user with ACL rules `+@all ~app1:* (+@readonly ~app2:*)`.
+This user has full access on `app1:*` and readonly access on `app2:*`.
+However, some commands support reading data from one key, doing some transformation, and storing it into another key.
+One such command is the `COPY` command, which copies the data from the source key into the destination key.
+The example set of ACL rules is unable to handle a request copying data from `app2:user` into `app1:user`, since neither the root permission or the selector fully matches the command.
```
zuiderkwast, line 447: root permission => main rule set?

topics/acl.md
```diff
@@ -396,6 +413,51 @@ Example:
...
+However, using key selectors you can define a set of ACL rules that can handle this request `+@all ~app1:* %R~app2:*`.
+The first pattern is able to match `app1:user` and the second pattern is able to match `app2:user`.
+
+Which type of permission is required for a command is documented through [key specifications](/topics/key-specs#logical-operation-flags).
+The type of permission is based off the keys logical operation flags. 
+The insert, update, and delete flags map to the write key permission. 
+The access flag flag maps to the read key permission.
```
zuiderkwast, line 454:
```suggestion
The access flag maps to the read key permission.
```

topics/acl.md
```diff
@@ -396,6 +413,51 @@ Example:
...
+
+Note: Side channels are ignored when it comes to evaluating read permissions.
+Many write commands return metadata about the modified key, for example `LPUSH` returns the size of the resulting list.
+Considering any metadata to be a read would limit the utility of the write key permissions.
```
zuiderkwast, line 458:
```suggestion
Considering any metadata to be a read would limit the utility of the read key permissions.
```

???

commands/acl-dryrun.md
```diff
@@ -0,0 +1,18 @@
...
+
+@examples
+
+```
+> ACL SETUSER FOO +SET ~*
+"OK"
+> ACL DRYRUN USER SET FOO BAR
+"OK"
+> ACL DRYRUN USER GET FOO BAR
+"This user has no permissions to run the 'GET' command"
```
zuiderkwast, line 17: This example is confusing. Is FOO the username or the key or both? Is USER the username or a subcommand?

Better use some username like "virginia" that you used somewhere else.

zuiderkwast (commented):
Looks pretty good in general.

commands/acl-setuser.md
```diff
@@ -17,6 +17,19 @@ set command (+set rule). Then another SETUSER call can modify the user rules:
...
+ACL rules can also be grouped into multiple distinct sets of rules, called selectors.
+Selectors are added by wrapping the rules in parentheses and providing them just like any other rule.
+To verify that a user can execute a command, first the root permissions are checked, and then all of its selectors are checked.
+If any of these rule sets match the command, the command can be executed.
+For example:
+
+    ACL SETUSER virginia on +GET allkeys (+SET ~app1*)
+
+This sets a user with two sets of permission, one defined on the user and one defined with a selector.
+The root user permissions only allows executing the get command, but can be executed on any keys.
```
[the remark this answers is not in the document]
zuiderkwast: Btw, here is a good place to define the concepts, e.g.

> This sets a user with two sets of permission: one defined on the user (*main rules*) and one defined with a selector (*selector rules*).

commands/acl-setuser.md
```diff
@@ -17,6 +17,19 @@ set command (+set rule). Then another SETUSER call can modify the user rules:
 
 The above rule will not apply the new rule to the user virginia, so other than `SET`, the user virginia will now be able to also use the `GET` command.
 
+ACL rules can also be grouped into multiple distinct sets of rules, called selectors.
```
oranagra, line 20:
```suggestion
Starting from Redis 7.0, ACL rules can also be grouped into multiple distinct sets of rules, called selectors.
```

commands/acl-setuser.md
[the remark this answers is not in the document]
oranagra: i also had a problem with the term `root` in this context (confused with the root user in linux).
i'd be ok to keep using the term root, but we need to be more verbose on the first time we present it.
i.e.
> The root permissions (defined outside parenthesis) .....

commands/acl-setuser.md
```diff
@@ -40,11 +53,14 @@ The following documentation is a reference manual about the capabilities of this
...
-* `on`: set the user as active, it will be possible to authenticate as this user using `AUTH <username> <password>`.
-* `off`: set user as not active, it will be impossible to log as this user. Please note that if a user gets disabled (set to off) after there are connections already authenticated with such a user, the connections will continue to work as expected. To also kill the old connections you can use `CLIENT KILL` with the user option. An alternative is to delete the user with `ACL DELUSER`, that will result in all the connections authenticated as the deleted user to be disconnected.
+### Command rules
+
 * `~<pattern>`: add the specified key pattern (glob style pattern, like in the `KEYS` command), to the list of key patterns accessible by the user. You can add multiple key patterns to the same user. Example: `~objects:*`
+* `%R~<pattern>`: Add the specified read key pattern. This behaves similar to the regular key pattern but only grants permission to read from keys that match the given pattern.
+* `%W~<pattern>`: Add the specified write key pattern. This behaves similar to the regular key pattern but only grants permission to write to keys that match the given pattern.
```
oranagra, line 63: 1. we must specify that these rules are only supported from redis 7.0 and up.
2. i think we must give a link to the place that specifies what's a read, probably to the section i see you added in topics/acl (`https://redis.io/topics/key-specs#Key permissions`?)
3. i think we should change the above description of the `~<pattern>` to explicitly say it allows both read and write
4. i think it might be nice to also mention that `%RW~` is supported and is an alias for just `~`

these comments are probably applicable for topics/acl.md too

topics/acl.md
```diff
@@ -200,9 +207,16 @@ computers to read, while `ACL LIST` is more biased towards humans.
...
     6) "-@all +get"
     7) "keys"
-    8) 1) "cached:*"
+    8) "~cached:*"
     9) "channels"
-    10) 1) "*"
+    10) "&*"
+    11) "selectors"
+    12) 1) 1) "commands"
+           2) "-@all +set"
```
oranagra, line 215: commenting here, i think the doc page for ACL GETUSER must mention this change.
maybe also in the history (which is part of the json file in the redis repo)

topics/acl.md
```diff
@@ -126,6 +128,11 @@ Configure valid passwords for the user:
 
 *Note: an use that is not flagged with nopass, and has no list of valid passwords, is effectively impossible to use, because there will be no way to log in as such user.*
 
+Configure selectors for the user:
+
+* `(<rule list>)`: Create a new selector to match rules against. Selectors are evaluated after the user permissions, and are evaluated according to the order they are defined. If a command matches either the user permissions or any selector, it is allowed.
```
oranagra, line 133: add a link to the selectors section below?

topics/acl.md
```diff
@@ -396,6 +413,51 @@ Example:
...
+For example, consider a user with the ACL rules `+GET ~key1 (+SET ~key2)`.
+This user is able to execute `GET key1` and `SET key2 hello`, but not `GET key2` or `SET key1 world`.
+
+Unlike the root user permissions, selectors can not be modified after they are added.
+Selectors can be removed with the `clearselectors` keyword, which removes all of the added selectors.
+Note that `clearselectors` does not remove the root user permissions.
+
+## Key permissions
+
+Redis key patterns can also be used to define how a command is able to touch a key.
```
oranagra, line 432:
```suggestion
Starting with Redis 7.0, Redis key patterns can also be used to define how a command is able to touch a key.
```

topics/acl.md
```diff
@@ -396,6 +413,51 @@ Example:
...
+
+For a concrete example, consider a user with ACL rules `+@all ~app1:* (+@readonly ~app2:*)`.
+This user has full access on `app1:*` and readonly access on `app2:*`.
+However, some commands support reading data from one key, doing some transformation, and storing it into another key.
+One such command is the `COPY` command, which copies the data from the source key into the destination key.
+The example set of ACL rules is unable to handle a request copying data from `app2:user` into `app1:user`, since neither the root permission or the selector fully matches the command.
+However, using key selectors you can define a set of ACL rules that can handle this request `+@all ~app1:* %R~app2:*`.
+The first pattern is able to match `app1:user` and the second pattern is able to match `app2:user`.
+
+Which type of permission is required for a command is documented through [key specifications](/topics/key-specs#logical-operation-flags).
```
oranagra, line 451: i think there's a way to provide a direct link, maybe there's a better way than `https://redis.io/topics/key-specs#logical-operation-flags`. @itamarhaber

topics/acl.md
madolson: Not sure about this, seems like I should stick with one way to describe the root user permission. I would think I should update everything to main in this case.

topics/acl.md
oranagra: ohh, never mind. i see this is the right syntax.

topics/acl.md
zuiderkwast: Oh, right, i rememver this discussion in the original PR. Perhaps it can be explained even more thoroughly or with emphasis on *write* to highlight that it's not a typo.

commands/acl-setuser.md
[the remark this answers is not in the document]
zuiderkwast: Agree. Defining a term and then sticking to it will make it very clear.

I'm fine with root or any other term.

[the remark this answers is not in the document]
madolson: Okay, I'll just explicitly work to define it. We use root in the engine code.

topics/acl.md
madolson: Yeah, I think my mistake was I start by talking about read, then I pivot to talking about write. In reality it's just a conversation about read/lack of read permissions.

madolson:
@zuiderkwast Thanks for the comments! I still stuck with the "root" terminology, since it makes the most sense to me. In both cases it is defined before usage. If you still feel strongly I would be okay switch it to "main" or "primary" permissions as well.

commands/acl-getuser.md
```diff
@@ -1,30 +1,42 @@
...
-patterns. Note that command rules are returned as a string in the same
-format used with the `ACL SETUSER` command. This description of command rules
-reflects the user's effective permissions, so while it may not be identical to
-the set of rules used to configure the user, it is still functionally identical.
+Specifically, it lists the user's ACL flags, password hashes, commands, key patterns, channel patterns (Added in version 6.2) and selectors (Added in version 7.0).
+Additional information may be returned in the future if more metadata is added to the user.
+
+Command rules are always returned in the same format as the one used in the `ACL SETUSER` command.
+Before version 7.0, keys and channels were returned as an array of patterns, however in version 7.0 later they are now also returned in same format as the one used in the `ACL SETUSER` command.
+Selectors are
```
oranagra, line 8:
```suggestion
```

oranagra (approved):

topics/acl.md
oranagra: @madolson please make a PR to the redis-repo to add a history entry for ACL GETUSER

commands/acl-setuser.md
```diff
@@ -17,6 +17,18 @@ set command (+set rule). Then another SETUSER call can modify the user rules:
 
 The above rule will not apply the new rule to the user virginia, so other than `SET`, the user virginia will now be able to also use the `GET` command.
 
+Starting from Redis 7.0, ACL rules can also be grouped into multiple distinct sets of rules, called selectors.
+Selectors are added by wrapping the rules in parentheses and providing them just like any other rule.
+In order to execute a command, either the root permissions (rules defined outside of parenthesis) or any of the selectors (rules defined inside parenthesis) must match the given command.
+For example:
+
+    ACL SETUSER virginia on +GET allkeys (+SET ~app1*)
```
zuiderkwast, line 25: Is "on" a pure token, then perhaps spell it in uppercase?

commands/acl-setuser.md
```diff
@@ -57,12 +73,19 @@ This is a list of all the supported Redis ACL rules:
...
 * `#<hashedpassword>`: Add the specified hashed password to the list of user passwords. A Redis hashed password is hashed with SHA256 and translated into a hexadecimal string. Example: `#c3ab8ff13720e8ad9047dd39466b3c8974e592c2fa383d4a3960714caef0c4f2`.
 * `<password`: Like `>password` but removes the password instead of adding it.
 * `!<hashedpassword>`: Like `#<hashedpassword>` but removes the password instead of adding it.
-* reset: Remove any capability from the user. It is set to off, without passwords, unable to execute any command, unable to access any key.
+* `(<rule list>)`: (Available in Redis 7.0 and later) Create a new selector to match rules against. Selectors are evaluated after the user permissions, and are evaluated according to the order they are defined. If a command matches either the user permissions or any selector, it is allowed. See [selectors](/topics/acl#selectors) for more information.
+* `clearselectors`: (Available in Redis 7.0 and later) Delete all of the selectors attached to the user.
+* `reset`: Remove any capability from the user. It is set to off, without passwords, unable to execute any command, unable to access any key.
```
zuiderkwast, line 88: How about uppercase for the pure tokens `CLEARSELECTORS`, `RESET` here as well as and `ON`, `OFF` and `NOPASS` above?

zuiderkwast (approved):
A few final comments, then I'm happy.

commands/acl-setuser.md
madolson: No, it's one of the ACL rules. Which, in this file at least are never uppercased.

commands/acl-setuser.md
madolson: Not sure these are pure tokens, since they aren't part of the command syntax. We would probably want all keywords to uppercase then throughout the doc, which I don't really want to mess with git history in this PR.

topics/acl.md
madolson: Yes, that is why this one is still open :)

madolson: https://github.com/redis/redis/pull/10200

commands/acl-setuser.md
zuiderkwast: OK. Leave them lowercase.

zuiderkwast:
Fine, although it's VIRGINIA in DRYRUN and virginia in SETUSER. :-)

PR #1769, opened 2022-02-01
Add note about acl-pubsub-default default behaviour.

hpatro:


enjoy-binbin (commented):
ref: https://github.com/redis/redis/pull/10181

oranagra (approved):

PR #1775, opened 2022-02-04
Update doc for acl getuser

hwware:
return nil when user does not exist

### DIFF
--- a/commands/acl-getuser.md
+++ b/commands/acl-getuser.md
@@ -11,6 +11,8 @@ Selectors are listed in the order they were applied to the user, and include inf
 
 @array-reply: a list of ACL rule definitions for the user.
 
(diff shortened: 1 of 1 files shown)


commands/acl-getuser.md
```diff
@@ -11,6 +11,8 @@ Selectors are listed in the order they were applied to the user, and include inf
 
 @array-reply: a list of ACL rule definitions for the user.
 
+If `user` does not exist a nil reply is returned.
```
itamarhaber, line 14:
```suggestion
If `user` does not exist a @nil-reply is returned.
```

itamarhaber (approved):

itamarhaber:
Thanks @hwware - merging.

PR #1784, opened 2022-02-08
GETKEYSANDFLAGS sub-command, new module apis, and key-spec flags changes

oranagra:
doc changes for https://github.com/redis/redis/pull/10237

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -582,7 +582,8 @@
                 },
                 "RW": true,
                 "access": true,
-                "update": true
+                "update": true,
+                "variable_flags": true
             }
         ],
         "arguments": [
@@ -2854,6 +2855,21 @@
             "stale"
         ]
     },
+    "COMMAND GETKEYSANDFLAGS": {
+        "summary": "Extract keys given a full Redis command",
+        "since": "7.0.0",
+        "group": "server",
+        "complexity": "O(N) where N is the number of arguments to the command",
+        "acl_categories": [
+            "@slow",
+            "@connection"
+        ],
+        "arity": -4,
+        "command_flags": [
+            "loading",
+            "stale"
+        ]
+    },
     "COMMAND HELP": {
         "summary": "Show helpful text about the different subcommands",
         "since": "5.0.0",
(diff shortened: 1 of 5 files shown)


commands.json
```diff
@@ -2854,6 +2855,21 @@
             "stale"
         ]
     },
+    "COMMAND GETKEYSANDFLAGS": {
+        "summary": "Extract keys given a full Redis command",
```
itamarhaber, line 2859:
```suggestion
        "summary": "Extract keys and flags given a full Redis command",
```

commands.json
```diff
@@ -4244,7 +4260,7 @@
         "arity": 3,
         "arguments": [
             {
-                "name": "function-name",
+                "name": "library-name",
```
itamarhaber, line 4263: 👍

itamarhaber (approved):

topics/key-specs.md
```diff
@@ -180,6 +181,12 @@ The difficulty arises, for example, because the string _"STORE"_ is both a keywo
 the only commands with _incomplete_ key specifications are `SORT` and `MIGRATE`.
 We don't expect the addition of such commands in the future.
 
+### variable_flags
+
+In some commands the flags for the same key name argument can depend on other arguments.
+For example the SET command has a GET argument, without the GET argument, SET is a write-only command, but with it, it becomes read+write.
+When this flag is present, it means that the flags in the key-spec will cover all options, but depending on some arguments, some flags can be removed or changed.
```
itamarhaber, line 188:
```suggestion
In some commands, the flags for the same key name argument can depend on other arguments.
For example, consider the `SET` command and its optional  _GET_ argument.
Without the _GET_ argument, `SET` is write-only, but it becomes a read and write command with it.
When this flag is present, it means that the key specification flags cover all possible options, but the effective flags depend on other arguments.
```

commands.json
oranagra: @itamarhaber should i remind you this is an auto-generated file?
i can apply your suggestion on the source, but i'm too lazy, and i think it's clear without it..

commands.json
itamarhaber: :)

PR #1786, opened 2022-02-09
Change zrange parameters names

pchmielewski:
In #1736 there were introduced new names of parameters 'min', 'max' (now 'start', 'stop'), but command in header wasn't updated

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -15665,11 +15665,11 @@
                 "key_spec_index": 0
             },
             {
-                "name": "min",
+                "name": "start",
                 "type": "string"
             },
             {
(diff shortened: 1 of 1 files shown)


itamarhaber:
Hi @pchmielewski,

Thanks for noting and fixing - much appreciated!

I'm afraid, however, that the fix isn't effective - these "definitions" now live at https://github.com/redis/redis/blob/unstable/src/commands/zrange.json.
Would you care to close this PR and create a new one there?

itamarhaber:
I'm so forget-full!

PR #1787, opened 2022-02-10
Fix incorrect references to SYNC command in function dump doc

enjoy-binbin:


itamarhaber (approved):
Thank you @enjoy-binbin

PR #1800, opened 2022-02-22
Add parameter part for  PUBSUB SHARDNUMSUB command 

hwware:
PUBSUB SHARDNUMSUB command misses the following parameter part:

"arguments": [
{
"name": "channel",
"type": "string",
"optional": true,
"multiple": true
}


oranagra:
this json file is auto generated. no need to make PR to update it.

PR #1803, opened 2022-02-28
Clarify the importance of CLUSTER SETSLOT NODE to dst before src

zuiderkwast:
See redis/redis#7116


enjoy-binbin (approved):
good job, it also solved one of my doubts

zuiderkwast:
@enjoy-binbin Thanks. Do you think we should fix redis-cli to use this order too? Right now it sends SETSLOTS NODE to all nodes in any order.

enjoy-binbin:
@zuiderkwast yes, i think we can fix it, looks like it's easy to fix

zuiderkwast:
@enjoy-binbin See redis/redis#10381. (It also fixes another problem.)

PR #1806, opened 2022-03-02
Explicitly describe how Lua script load in cluster mode

hwware:
Through this PR, user could explicitly know in the situation when there are multiple master nodes in cluster,
the Lua script should be loaded  individually on each master node

### DIFF
--- a/topics/cluster-tutorial.md
+++ b/topics/cluster-tutorial.md
@@ -106,7 +106,10 @@ slots hold by nodes, does not require any downtime.
 
 Redis Cluster supports multiple key operations as long as all the keys involved
 into a single command execution (or whole transaction, or Lua script
-execution) all belong to the same hash slot. The user can force multiple keys
+execution) all belong to the same hash slot. For Lua script, if there are 
+multiple master nodes, and the user wants to execute the operations on multiply keys 
+which belong to different nodes in a script, the user should load and execute the script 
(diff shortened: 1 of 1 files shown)


topics/cluster-tutorial.md
```diff
@@ -106,7 +106,10 @@ slots hold by nodes, does not require any downtime.
 
 Redis Cluster supports multiple key operations as long as all the keys involved
 into a single command execution (or whole transaction, or Lua script
-execution) all belong to the same hash slot. The user can force multiple keys
+execution) all belong to the same hash slot. For Lua script, if there are 
+multiple master nodes, and the user wants to execute the operations on multiply keys 
+which belong to different nodes in a script, the user should load and execute the script 
+individually on each master node, and the results aggregated. The user can force multiple keys
```
madolson, line 112: I'm not sure this is really all that different from multi-key commands, which are already described here. You're sort of describing a work around here, running the script on each master, but the same would be true for any multi-key command.

hwware: Hi Madolson, I checked cluster doc part, in the section --Migrating to Redis Cluster, it explicitly writes that Redis Cluster DO NOT support Multiple keys operations, or transactions, or Lua scripts involving multiple keys operate in different tag. 
So I will close PR, and thanks for your explanation to me.

PR #1809, opened 2022-03-06
add sort command new acl documentation

ranshid:
1. Add notice that 'GET' and 'BY" options are not valid for cluster-mode
2. Add notice for 'GET' and 'SET' options are not valid when no
   Full key read permissions

### DIFF
--- a/commands/sort.md
+++ b/commands/sort.md
@@ -112,6 +112,7 @@ Starting from Redis 7.0, any use of `GET` or `BY` which reference external key p
 Full key read permissions can be set for the user by, for example, specifying `'%R~*'` or `'~*` with the relevant command access rules.
 You can check the `ACL SETUSER` command manual for more information on setting ACL access rules.
 If full key read permissions aren't set, the command will fail with an error.
+
 ## Storing the result of a SORT operation
 
 By default, `SORT` returns the sorted elements to the client.
--- a/commands/sort.md
+++ b/commands/sort.md
@@ -103,21 +103,14 @@ It is also possible to `!GET` the element itself using the special pattern `#`:
 SORT mylist BY weight_* GET object_* GET #
 ```
 
-## Restrictions for using external keys with Sort command
+## Restrictions for using external keys
 
-When enabling `Redis cluster-mode` there is no way to garantee the exsistence of
-the external keys on the node which the command is processed on. In case the 
-`cluster-enabled` configuration is set to `yes` any use of `GET` or `BY`
-which reference external key pattern will cause the command to fail with error.
+When enabling `Redis cluster-mode` there is no way to guarantee the existence of the external keys on the node which the command is processed on.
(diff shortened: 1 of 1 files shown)


ranshid:
@madolson - please review.
I am not sure I chose the best structure for this, so I am ready to get other suggestions.

commands/sort.md
```diff
@@ -103,6 +103,22 @@ It is also possible to `!GET` the element itself using the special pattern `#`:
 SORT mylist BY weight_* GET object_* GET #
 ```
 
+## Restrictions for using external keys
+
+When enabling `Redis cluster-mode` there is no way to guarantee the existence of the external keys on the node which the command is processed on.
+In this case, any use of `GET` or `BY` which reference external key pattern will cause the command to fail with an error.
+
+Starting from Redis 7.0, any use of `GET` or `BY` which reference external key pattern will only be allowed in case the current user running the command has full key read permissions.
+Full key read permissions can be set for the user by, for example, specifying `'%R~*'` or `'~*` with the relevant command access rules. You can check the `ACL SETUSER` command manual for more information on setting ACL access rules.
```
oranagra, line 112: nit pick, i'd suggest to break that line, one line per sentence, would make future edits easier to review (smaller diff).

madolson, line 112:
```suggestion
Full key read permissions can be set for the user by, for example, specifying `'%R~*'` or `'~*` with the relevant command access rules.
You can check the `ACL SETUSER` command manual for more information on setting ACL access rules.
```

commands/sort.md
```diff
@@ -103,6 +103,22 @@ It is also possible to `!GET` the element itself using the special pattern `#`:
...
+If full key read permissions aren't set, the command will fail with an error.
+
+    > ACL SETUSER alice on NOPASS +sort ~weight_* ~mylist
+        OK    
+    > AUTH alice
+        OK
+    > SORT mylist BY weight_*
+        (error) NOPERM this user has no permissions to access one of the keys used as arguments
+
```
madolson, line 121:
```suggestion
```

commands/sort.md
```diff
@@ -103,6 +103,15 @@ It is also possible to `!GET` the element itself using the special pattern `#`:
...
 
+## Restrictions for using external keys
+
+When enabling `Redis cluster-mode` there is no way to guarantee the existence of the external keys on the node which the command is processed on.
+In this case, any use of `GET` or `BY` which reference external key pattern will cause the command to fail with an error.
+
+Starting from Redis 7.0, any use of `GET` or `BY` which reference external key pattern will only be allowed in case the current user running the command has full key read permissions.
+Full key read permissions can be set for the user by, for example, specifying `'%R~*'` or `'~*` with the relevant command access rules.
+You can check the `ACL SETUSER` command manual for more information on setting ACL access rules.
+If full key read permissions aren't set, the command will fail with an error.
```
madolson, line 114:
```suggestion
If full key read permissions aren't set, the command will fail with an error.

```

PR #1815, opened 2022-03-14
Added cluster shard command documentation

madolson:
Ref: https://github.com/redis/redis/pull/10293

### DIFF
--- a/commands/cluster-shards.md
+++ b/commands/cluster-shards.md
@@ -3,8 +3,8 @@ A shard is defined as a collection of nodes that serve the same set of slots and
 A shard may only have a single master at a given time, but may have multiple replicas.
 It is possible for a shard to not be serving any slots while still having replicas.
 
-The command is suitable to be used by Redis Cluster client libraries implementations in order to understand the topology of the cluster.
-A client should issue this command on startup in order to retrieve the map associating cluster *hash slots* with actual node network information.
+The command is suitable to be used by Redis Cluster client libraries in order to understand the topology of the cluster.
+A client should issue this command on startup in order to retrieve the map associating cluster *hash slots* with actual node information.
 This map should be used to direct commands to the node that is likely serving the slot associated with a given command.
 In the event the command is sent to the wrong node, this command can then be used to update the topology of the cluster.
 
@@ -31,7 +31,7 @@ The current list of attributes:
 * id: The unique node id for this particular node.
 * endpoint: The preferred endpoint to reach the node, see below for more information about the possible values of this field.
 * ip: The IP address to send requests to for this node.
-* hostname (optional): The announced FQDN hostname to send requests to for this node.
+* hostname (optional): The announced hostname to send requests to for this node.
 * port (optional): The TCP port number of the node.
 * tls-port (optional): The TLS port of the node.
 * role: The replication role of this node.
@@ -39,7 +39,7 @@ The current list of attributes:
 * health: Either `online`, `failed`, or `loading`. This information should be used to determine which nodes should be sent traffic. The `loading` health state should be used to know that a node is not currently eligible to serve traffic, but may be eligible in the future. 
 
 The endpoint, along with the port, defines the location that clients should use to send requests for a given slot.
-A NULL value for the endpoint indicates the node has an unknown endpoint and the client should connect to the same endpoint it used to send the `CLUSTER SLOTS` command but with the port returned from the command.
+A NULL value for the endpoint indicates the node has an unknown endpoint and the client should connect to the same endpoint it used to send the `CLUSTER SHARDS` command but with the port returned from the command.
 This unknown endpoint configuration is useful when the Redis nodes are behind a load balancer that Redis doesn't know the endpoint of.
 Which endpoint is set is determined by the `cluster-preferred-endpoint-type` config.
 
--- a/commands/cluster-shards.md
+++ b/commands/cluster-shards.md
@@ -1,17 +1,20 @@
 `CLUSTER SHARDS` returns details about the shards of the cluster.
 A shard is defined as a collection of nodes that serve the same set of slots and that replicate from each other.
-A shard may only have a single master at a given time, but may have multiple replicas.
+A shard may only have a single master at a given time, but may have multiple or no replicas.
 It is possible for a shard to not be serving any slots while still having replicas.
+
+This command replaces the `CLUSTER SLOTS` command, by providing a more efficient and extensible representation of the cluster. 
 
 The command is suitable to be used by Redis Cluster client libraries in order to understand the topology of the cluster.
 A client should issue this command on startup in order to retrieve the map associating cluster *hash slots* with actual node information.
 This map should be used to direct commands to the node that is likely serving the slot associated with a given command.
-In the event the command is sent to the wrong node, this command can then be used to update the topology of the cluster.
+In the event the command is sent to the wrong node, in that it received a '-MOVED' redirect, this command can then be used to update the topology of the cluster.
 
 The command returns an array of shards, with each shard containing two fields, 'slots' and 'nodes'. 
 
-The 'slots' field is a list of slot ranges served by this shard, stored as pair of integers storing the start and end slots of the ranges.
-For example, if a node owns the slots 1, 2, 3, 5, 7, 8, 9, the slots field would store the ranges like:
+The 'slots' field is a list of slot ranges served by this shard, stored as pair of integers representing the inclusive start and end slots of the ranges.
+For example, if a node owns the slots 1, 2, 3, 5, 7, 8 and 9, the slots ranges would be stored as [1-3], [5-5], [7-9].
+The slots field would therefor be represented by the following list of integers.
 
 ```
 1) 1) "slots"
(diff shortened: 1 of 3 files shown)


commands/cluster-shards.md
```diff
@@ -0,0 +1,150 @@
...
+   2) 1) (integer) 1
+      2) (integer) 3
+      3) (integer) 5
+      4) (integer) 5
+      5) (integer) 7
+      6) (integer) 9
+```
+
+The 'nodes' field contains a list of all nodes within the shard.
+Each individual node is a map of attributes that describe the node.
```
zuiderkwast, line 27: It's slightly confusing that it's described as a map, but the examples are in RESP2. Some mention of how RESP3 maps fall back to RESP2 could be useful here, possibly with a link to the protocol page.

commands/cluster-shards.md
```diff
@@ -0,0 +1,150 @@
...
+A shard is defined as a collection of nodes that serve the same set of slots and that replicate from each other.
+A shard may only have a single master at a given time, but may have multiple replicas.
+It is possible for a shard to not be serving any slots while still having replicas.
+
+The command is suitable to be used by Redis Cluster client libraries in order to understand the topology of the cluster.
+A client should issue this command on startup in order to retrieve the map associating cluster *hash slots* with actual node information.
+This map should be used to direct commands to the node that is likely serving the slot associated with a given command.
+In the event the command is sent to the wrong node, this command can then be used to update the topology of the cluster.
+
+The command returns an array of shards, with each shard containing two fields, 'slots' and 'nodes'.
```
zuiderkwast, line 11: A shard is a map (implicitly understood). See next comment.

topics/cluster-spec.md
```diff
@@ -390,7 +390,7 @@ be issued it can compute the hash slot of the target key and have a
 greater chance of choosing the right node.
 
 An alternative is to just refresh the whole client-side cluster layout
-using the `CLUSTER SLOTS` commands
+using the `CLUSTER SHARDS`, or the deprecated `CLUSTER SLOTS`, command
```
zuiderkwast, line 393: If we are deprecating it, shouldn't that be mentioned on the `CLUSTER SLOTS` page with a link to `CLUSTER SHARDS`?

commands/cluster-shards.md
```diff
@@ -0,0 +1,150 @@
+`CLUSTER SHARDS` returns details about the shards of the cluster.
+A shard is defined as a collection of nodes that serve the same set of slots and that replicate from each other.
+A shard may only have a single master at a given time, but may have multiple replicas.
+It is possible for a shard to not be serving any slots while still having replicas.
+
+The command is suitable to be used by Redis Cluster client libraries in order to understand the topology of the cluster.
+A client should issue this command on startup in order to retrieve the map associating cluster *hash slots* with actual node information.
+This map should be used to direct commands to the node that is likely serving the slot associated with a given command.
+In the event the command is sent to the wrong node, this command can then be used to update the topology of the cluster.
```
zuiderkwast, line 9: At the comma, mention that the client received a -MOVED redirect.

commands/cluster-shards.md
```diff
@@ -0,0 +1,150 @@
...
+Each individual node is a map of attributes that describe the node. 
+Some attributes are optional and more attributes may be added in the future. 
+The current list of attributes:
+
+* id: The unique node id for this particular node.
+* endpoint: The preferred endpoint to reach the node, see below for more information about the possible values of this field.
+* ip: The IP address to send requests to for this node.
+* hostname (optional): The announced hostname to send requests to for this node.
+* port (optional): The TCP port number of the node.
+* tls-port (optional): The TLS port of the node.
```
zuiderkwast, line 36: Mention (here or below the list) that at least one of the ports is present.

Maybe mention that `port` (non-TLS) is present if TLS is not required.

commands/cluster-shards.md
```diff
@@ -0,0 +1,150 @@
+`CLUSTER SHARDS` returns details about the shards of the cluster.
+A shard is defined as a collection of nodes that serve the same set of slots and that replicate from each other.
+A shard may only have a single master at a given time, but may have multiple replicas.
```
PingXie, line 3: There is also the case where a shard has no replica but just primary (master). I think it is worth calling out explicitly as well.

commands/cluster-shards.md
```diff
@@ -0,0 +1,150 @@
...
+It is possible for a shard to not be serving any slots while still having replicas.
+
+The command is suitable to be used by Redis Cluster client libraries in order to understand the topology of the cluster.
+A client should issue this command on startup in order to retrieve the map associating cluster *hash slots* with actual node information.
+This map should be used to direct commands to the node that is likely serving the slot associated with a given command.
+In the event the command is sent to the wrong node, this command can then be used to update the topology of the cluster.
+
+The command returns an array of shards, with each shard containing two fields, 'slots' and 'nodes'. 
+
+The 'slots' field is a list of slot ranges served by this shard, stored as pair of integers storing the start and end slots of the ranges.
```
PingXie, line 13: Clarify explicitly that both of range start and end are inclusive.

topics/cluster-spec.md
PingXie: Update commands.c to mark `CLUSTER SLOTS` deprecated as well?

Also what about `CLUSTER NODES`? Should it be deprecated too?

commands/cluster-shards.md
PingXie: Should we consider the pros/cons of using the native RESP 3 map too? I am assuming it is a strategic goal to increase the RESP 3 adoption over time. If backcompat is a concern, one option is to enable the native map only for clients understanding RESP 3 and fallback to RESP 2 otherwise.

madolson: We use the terminology fairly often, I don't think we should document everywhere that a map looks different between RESP 3 and RESP 2. (Like https://github.com/redis/redis-doc/blob/80a55ab6487f4096e200e3871c083d8468bfe7b3/commands/function-stats.md#L7)

commands/cluster-shards.md
madolson: I didn't follow this comment.

topics/cluster-spec.md
madolson: `CLUSTER NODES` is not deprecated. The redis PR includes a deprecation node of `CLUSTER SLOTS`. It looks like we need to manually deprecate commands, so I can add that here.

madolson: I guess deprecation notes happen automatically. I did add a note in cluster shards why to use this command instead of `CLUSTER SLOTS`.

commands/cluster-shards.md
madolson: Tweaked wording, let me know if that seems reasonable.

commands/cluster-shards.md
zuiderkwast: If you say so. The examples make it clear anyway.

commands/cluster-shards.md
zuiderkwast: "each shard containing two fields, 'slots' and 'nodes'" -- we never stated that it's a map, so that's implicit. In RESP2 it's an array with 4 elements. Like the other comment about RESP2 and RESP3, we don't need to clarify this IIUC.

topics/cluster-spec.md
zuiderkwast: When CLUSTER SLOTS is flagged as deprecated, it's good to point users to CLUSTER SHARDS from there. Otherwise, users may not find out as easily that CLUSTER SHARDS is its replacement. That's what I meant.

madolson: What I was saying is that I think there is code that automatically puts a link at the top that points to `CLUSTER SHARDS`. I was looking at `CLUSTER SLAVES`, which has this: https://redis.io/commands/cluster-slaves.

zuiderkwast: Gotcha

zuiderkwast (approved):

commands/cluster-shards.md
```diff
@@ -1,17 +1,20 @@
...
+A shard may only have a single master at a given time, but may have multiple or no replicas.
 It is possible for a shard to not be serving any slots while still having replicas.
 
+This command replaces the `CLUSTER SLOTS` command, by providing a more efficient and extensible representation of the cluster. 
+
 The command is suitable to be used by Redis Cluster client libraries in order to understand the topology of the cluster.
 A client should issue this command on startup in order to retrieve the map associating cluster *hash slots* with actual node information.
 This map should be used to direct commands to the node that is likely serving the slot associated with a given command.
-In the event the command is sent to the wrong node, this command can then be used to update the topology of the cluster.
+In the event the command is sent to the wrong node, in that it received a '-MOVED' redirect, this command can then be used to update the topology of the cluster.
```
PingXie, line 11: received --> receives?

commands/cluster-shards.md
```diff
@@ -32,8 +35,8 @@ The current list of attributes:
 * endpoint: The preferred endpoint to reach the node, see below for more information about the possible values of this field.
 * ip: The IP address to send requests to for this node.
 * hostname (optional): The announced hostname to send requests to for this node.
-* port (optional): The TCP port number of the node.
-* tls-port (optional): The TLS port of the node.
+* port (optional): The TCP (non-TLS) port of the node. At least one of port or tls-port will be present.
```
PingXie, line 38: not related to this PR but I think that we should add some tests to validate the contract defined here.

PR #1821, opened 2022-03-21
Fix typo in eval-intro.md

arun1587:
fix a minor typo in script cache section


enjoy-binbin (approved):
thanks for the fix
ps. there is no need to open an issue, if you have a fix and you will submit the PR. The PR can describe things more clearly

PR #1832, opened 2022-03-29
Added pub-sub alias

rrelledge:
The Pub/Sub link on the [About](https://redis.io/docs/about/) page redirected to /docs/manual/pub-sub instead of /docs/manual/pubsub, which caused a 404. This new alias should fix the 404.

### DIFF
--- a/docs/manual/pubsub.md
+++ b/docs/manual/pubsub.md
@@ -5,6 +5,7 @@ weight: 1
 description: How to use pub/sub channels in Redis
 aliases:
   - /topics/pubsub
+  - /docs/manual/pub-sub
 ---
 
 `SUBSCRIBE`, `UNSUBSCRIBE` and `PUBLISH`


zuiderkwast:
Good idea! We should have redirects for all the old URIs. There are many broken links now.

zuiderkwast:
I'm not sure the new-structure branch is used anymore. Perhaps you know this better than I do?

rrelledge:
> I'm not sure the new-structure branch is used anymore. Perhaps you know this better than I do?

@zuiderkwast No, I think you are right. Back when I opened this PR, I think master might have been slightly out-of-sync with the new-structure branch, so I wasn't sure. I can try to change the destination branch for this PR to master. But if that might cause issues, we can close the PR without merging and open a new one that points to correct branch.

rrelledge:
@zuiderkwast Done. It looks like changing the destination worked okay to me.

zuiderkwast (approved):

rrelledge:
@zuiderkwast By the way, I can't merge this PR myself since I don't have write access.

PR #1852, opened 2022-04-05
Update function docs according to recent changes

MeirShpilraien:
Update function docs according to recent changes on : https://github.com/redis/redis/pull/10500

### DIFF
--- a/topics/functions-intro.md
+++ b/topics/functions-intro.md
@@ -121,7 +121,7 @@ redis> FCALL knockknock 0
 "Who's there?"
 ```
 
-Notice that the `FUNCTION LOAD` command returns the name of the loaded library, this name can later be used to delete the library or to get information about the library.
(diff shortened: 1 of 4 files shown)


topics/functions-intro.md
```diff
@@ -116,12 +116,14 @@ We can load our library and use `FCALL` to call the registered function:
...
 ```
 redis> FUNCTION LOAD "#!lua name=mylib\nredis.register_function('knockknock', function() return 'Who\\'s there?' end)"
-OK
+mylib
 redis> FCALL knockknock 0
 "Who's there?"
 ```
 
-Note that we've provided `FCALL` with two arguments: the function's registered name and the numeric value `0`. This numeric value indicates the number of key names that follow it (the same way `EVAL` and `EVALSHA` work).
+Notice that the `FUNCTION LOAD` command returns the name of the loaded library, this name can later be used to delete the library or to get information about the library.
```
oranagra, line 124: maybe refer to the exact commands DELETE and LIST?

oranagra (approved):

PR #1860, opened 2022-04-08
Update correct link in Update _index.md

insanePi:


itamarhaber (commented):
I'll be closing the PR because it is against the new-structure branch and because the issue is already fixed. Nonetheless, thank you @insanePi!

PR #1862, opened 2022-04-10
The meaning of "-inf +inf"

hgs-study:


hgs-study:
Hello, I'm hyun, a Korean developer. I don't live in an English-speaking world, so when I looked at the docs, the sentence "inf" didn't make sense at once, so I asked Google. I'm asking pr because it would be easier to understand if a simple comment was added. thank you.

enjoy-binbin:
Hi, i am also not live in english-speaking world. (i am from China)
it is good that you asked the Google and got your answer. it is a nice habit

it might be like null / NULL / None / nil, it's something customary? (i am not sure about the wording)
unfortunately, i think we will not take the comment. (Ps you can see the example is in `cli` block, it is a special blcok that we we can run the real redis commands on it)

anyway, actually you can the `inf` meaning in https://redis.io/commands/zrangebyscore/#exclusive-intervals-and-infinity

enjoy-binbin (commented):
Hi, i am also not live in english-speaking world. (i am from China)
it is good that you asked the Google and got your answer. it is a nice habit

it might be like null / NULL / None / nil, it's something customary? (i am not sure about the wording)
unfortunately, i think we will not take the comment. (Ps you can see the example is in `cli` block, it is a special blcok that we we can run the real redis commands on it)

anyway, actually you can the `inf` meaning in https://redis.io/commands/zrangebyscore/#exclusive-intervals-and-infinity

commands/zcount.md
```diff
@@ -16,6 +16,6 @@ Note: the command has a complexity of just O(log(N)) because it uses elements ra
 ZADD myzset 1 "one"
 ZADD myzset 2 "two"
 ZADD myzset 3 "three"
-ZCOUNT myzset -inf +inf
+ZCOUNT myzset -inf +inf  // -inf +inf : The meaning of this sentence is infinity.
```
zuiderkwast, line 19: You can't use `//` as a comment in redis-cli, so this is an error if the example is pasted in redis-cli.

Please add this description of negative infinity and positive infinity somewhere else in this document, together with the description of min and max.

hgs-study: As you advised in the review, I added the contents of 'inf' and `//` to another place without using it. Please check. Thank you.

commands/zcount.md
```diff
@@ -11,11 +11,12 @@ Note: the command has a complexity of just O(log(N)) because it uses elements ra
 @integer-reply: the number of elements in the specified score range.
 
 @examples
+- `-inf +inf` : 'inf' stands for infinity, meaning from least negative infinity to maximum positive infinity.
```
zuiderkwast, line 14: I don't think this explanation should be under "Examples". It's better to explain it in the main documentation section. We should also explain exclusive ranges with `(`.

We already have "The min and max arguments have the same semantic as described for ZRANGEBYSCORE.", but ZRANGEBYSCORE is deprecated and replaced by ZRANGE with the BYSCORE argument. So instead, I think we should replace "The min and max arguments have the same semantic as described for ZRANGEBYSCORE." with the same explanation that we have in the ZRANGE docs:

> **start** and **stop** can be **-inf** and **+inf**, denoting the negative and positive infinities, respectively. This means that you are not required to know the highest or lowest score in the sorted set to get all elements from or up to a certain score.
>
> By default, the score intervals specified by **start** and **stop** are closed (inclusive). It is possible to specify an open interval (exclusive) by prefixing the score with the character `(`.

commands/zcount.md
```diff
@@ -11,11 +11,12 @@ Note: the command has a complexity of just O(log(N)) because it uses elements ra
...
 
 ```cli
 ZADD myzset 1 "one"
 ZADD myzset 2 "two"
 ZADD myzset 3 "three"
-ZCOUNT myzset -inf +inf
+ZCOUNT myzset -inf +inf
```
zuiderkwast, line 20: Don't change the whitespace here. (Suggesting self-review so you can see such mistakes.)

PR #1870, opened 2022-04-12
Fix broken link to redis docs on community page

ioreka:
Ironically a markdown error is currently providing an incorrectly-rendered link to redis docs :')


enjoy-binbin (commented):
thanks, ref: https://github.com/redis/redis-doc/pull/1833 (duplicated)

PR #1885, opened 2022-04-22
Add coredis library to python clients

alisaifee:
# Details 
[coredis](https://github.com/alisaifee/coredis) was forked from [aredis](https://github.com/NoneGG/aredis) as the latter had become unmaintained since 2020.

- coredis [2.x](https://github.com/alisaifee/coredis/tree/2.x) refreshed aredis with support for python 3.10 and brought in some outstanding pull requests/bug fixes.
- coredis [3.x](https://coredis.readthedocs.org/en/latest) has diverged from aredis and is no longer API compatible. It adds support for redis 7, type annotations, RESP3 support among others.

As a follow up I would also recommend marking aredis as unmaintained.

### DIFF
--- a/clients/python/github.com/alisaifee/coredis.json
+++ b/clients/python/github.com/alisaifee/coredis.json
@@ -1,0 +1,4 @@
+{
+  "name": "coredis",
+  "description": "Async redis client with support for redis server, cluster & sentinel"
+}


zuiderkwast:
Thanks!

Please add the follow up and mark aredis as unmaintained.

I think it would be good to add some of this to the descriptions so users can know a bit more before following the links:

> [coredis](https://github.com/alisaifee/coredis) was forked from [aredis](https://github.com/NoneGG/aredis) as the latter had become unmaintained since 2020.
> 
> * coredis [2.x](https://github.com/alisaifee/coredis/tree/2.x) refreshed aredis with support for python 3.10 and brought in some outstanding pull requests/bug fixes.
> 
> * coredis [3.x](https://coredis.readthedocs.org/en/latest) has diverged from aredis and is no longer API compatible. It adds support for redis 7, type annotations, RESP3 support among others.

alisaifee:
@zuiderkwast thanks! What would  be the recommended way of marking a client as unmaintained?
The various options for async (and otherwise) clients for python actually needs a bit of a cleanup as:
1. Aioredis is now absorbed by redis-py
2. redis-py-cluster is now absorbed by redis-py
3. aredis is unmaintained 
4. asyncio-redis has not had an update since 2020

zuiderkwast:
> What would be the recommended way of marking a client as unmaintained?

It is already implicitly marked as described on the clients page: "Repositories with some activity within the latest six months are marked with a (greed dot)." so if a dot is missing it has no recent activity.

Apart from that, there is no recommended way. We could use the description field and we could also just delete obsolete clients from the list.

zuiderkwast:
I don't know why it doesn't appear on the page. @itamarhaber do you know?

alisaifee:
> I don't know why it doesn't appear on the page. @itamarhaber do you know?

A wild guess based on symptoms is caching. I noticed that clients removed in #1984 are indeed removed, however the content update for existing clients is also not reflect (e.g. redis-py-cluster's description is still outdated)

zuiderkwast:
@banker The clients page doesn't seem to get updated. Do you think it can be a caching issue? I guess someone needs to look into it.

PR #1902, opened 2022-04-28
Update install-redis-on-linux.md

kirkhang:
There's an extra space

### DIFF
--- a/docs/getting-started/installation/install-redis-on-linux.md
+++ b/docs/getting-started/installation/install-redis-on-linux.md
@@ -30,6 +30,6 @@ The [Snapcraft store](https://snapcraft.io/store) provides [Redis installation p
 sudo yum install epel-release
 sudo yum install snapd
 sudo systemctl enable --now snapd.socket
-sudo ln -s /var/lib/snapd/snap /snap
+sudo ln -s /var/lib/snapd/snap/snap
 sudo snap install redis
 {{< / highlight  >}}


enjoy-binbin (approved):

docs/getting-started/installation/install-redis-on-linux.md
```diff
@@ -30,6 +30,6 @@ The [Snapcraft store](https://snapcraft.io/store) provides [Redis installation p
 sudo yum install epel-release
 sudo yum install snapd
 sudo systemctl enable --now snapd.socket
-sudo ln -s /var/lib/snapd/snap /snap
+sudo ln -s /var/lib/snapd/snap/snap
 sudo snap install redis
```
zuiderkwast, line 34: This line doesn't make much sense. Is there a file at `/var/lib/snapd/snap/snap` or is it creating a symlink to a nonexisting file? Have you verified this on CentOS?

I think it's better to delete this line. The next command `sudo snap install redis` is using snap from the $PATH and not using `./snap` anyway.

I my system (Debian) I have `/var/lib/snapd/snaps/`, but I don't have `/var/lib/snapd/snap/snap` nor `/var/lib/snapd/snap`. The snap binary is at `/usr/bin/snap`.

zuiderkwast (changes requested):

docs/getting-started/installation/install-redis-on-linux.md
yossigo: @zuiderkwast I agree, not sure where this came from but this page is wrong, what it really describes is how to install snap on Red Hat but I don't think it should be here.

yossigo:
Obsoleted by #1990

PR #1908, opened 2022-04-30
Fix the example in the FCALL section

rafaelvargas:
Fixes the example shown in the FCALL section. Corrects the `FUNCTION LOAD` command invocation.


enjoy-binbin (approved):
thanks, LGTM. see #1852 and https://redis.io/commands/function-load/#examples

PR #1923, opened 2022-05-08
Add Medis to the tools page

luin:
Hey there 👋

Just want to list Medis on the tools page so users can find and evaluate whether it fits they needs.

BTW I found that the instructions for building the website on README seem outdated so LMK if there is anything missing in this PR. Thanks for the great work as always ❤️.

### DIFF
--- a/tools/gui/github.com/luin/medis.json
+++ b/tools/gui/github.com/luin/medis.json
@@ -1,0 +1,6 @@
+{
+  "name": "Medis",
+  "description": "Medis is a beautiful, easy-to-use Mac database management application for Redis.",
+  "homepage": "https://github.com/luin/medis",
+  "twitter": ["medisclient"]
+}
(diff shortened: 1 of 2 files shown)


itamarhaber:
Hi @luin and thank you for the PR!

Regrettably, Medis is not open-source so we can't accept it at this time. As for the outdated instructions, thanks for the heads up - I'll see what I can do about it.

luin:
@itamarhaber Thanks for the quick reply. That makes sense. I updated the PR to link to Medis 1 then, which is open-sourced and work on both Windows and Mac. LMK if that conforms to rules.

itamarhaber:
Totally :) thanks!

luin:
Hey @itamarhaber,
Just a quick follow-up: it looks like Medis is still not on https://redis.io/docs/tools/#gui. Is there anything I missed in this PR?

PR #1926, opened 2022-05-09
fix incorrect link

zhugezy:
don't know if it's a correct fix ;)


zuiderkwast (approved):
The link works: redis.io/docs/manual/programmability/functions-intro/#function-flags

I think it is correct. I'm not merging anything as long as the CI jobs are failing though. (Not your fault.)

itamarhaber (approved):
Thank you @zhugezy - merging!

PR #1953, opened 2022-05-21
Add documentation documenting read-only scripts

madolson:
Based on discussion from https://github.com/redis/redis/pull/10728#issuecomment-1127313929, want to better define when to use `EVAL` et. al vs `EVAL_RO` et al.

PR #1954, opened 2022-05-22
script flags update

oranagra:
updating docs with regards to these issues:
* https://github.com/redis/redis/pull/10699
* https://github.com/redis/redis-doc/issues/1948
* https://github.com/redis/redis/pull/10744

### DIFF
--- a/docs/manual/programmability/lua-api.md
+++ b/docs/manual/programmability/lua-api.md
@@ -467,16 +467,19 @@ You can use the following flags and instruct the server to treat the scripts' ex
     However, note that the server will return an error if the script attempts to call a write command.
     Also note that currently `PUBLISH`, `SPUBLISH` and `PFCOUNT` are also considered write commands in scripts, because they could attempt to propagate commands to replicas and AOF file.
 
+    For more information please refer to [Read-only scripts](/docs/manual/programmability/#read-only_scripts)
+
 * `allow-oom`: use this flag to allow a script to execute when the server is out of memory (OOM).
 
-    Unless used, Redis will deny the execution of scripts when in an OOM state.
+    Unless used, Redis will deny the execution of flagged scripts (Functions and Eval scripts with [shebang](/topics/eval-intro#eval-flags)) when in an OOM state.
     Furthermore, when you use this flag, the script can call any Redis command, including commands that aren't usually allowed in this state.
     Specifying `no-writes` or using `FCALL_RO` / `EVAL_RO` also implies the script can run in OOM state (without specifying `allow-oom`)
 
-* `allow-stale`: a flag that enables running the script against a stale replica when the `replica-serve-stale-data` config is set to `no` .
+* `allow-stale`: a flag that enables running the flagged scripts (Functions and Eval scripts with [shebang](/topics/eval-intro#eval-flags)) against a stale replica when the `replica-serve-stale-data` config is set to `no` .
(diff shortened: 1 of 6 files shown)


docs/manual/programmability/lua-api.md
```diff
@@ -454,25 +454,29 @@ You can use the following flags and instruct the server to treat the scripts' ex
...
     Using this flag allows executing the script:
-    1. With `FCALL_RO` / `EVAL_RO` against masters and read-only replicas.
-    2. Even if there's a disk error (Redis is unable to persist so it rejects writes).
+    1. With `FCALL_RO` / `EVAL_RO`
+    2. On read-only replicas.
+    3. Even if there's a disk error (Redis is unable to persist so it rejects writes).
+    4. When over the memory limit since it implies the script doesn't increase memory consumption (see `allow-oom` below)
 
     However, note that the server will return an error if the script attempts to call a write command.
+    Also note that currently `PUBLISH`, `SPUBLISH` and `PFCOUNT` are also considered write commands in scripts, because they could attempt to propagate commands to replicas and AOF file.
```
MeirShpilraien, line 468: Should we mentioned that its on 7.0.1 and above?

docs/manual/programmability/lua-api.md
```diff
@@ -454,25 +454,29 @@ You can use the following flags and instruct the server to treat the scripts' ex
...
     Furthermore, when you use this flag, the script can call any Redis command, including commands that aren't usually allowed in this state.
+    Specifying `no-writes` or using `FCALL_RO` / `EVAL_RO` also implies the script can run in OOM state (without specifying `allow-oom`)
 
-* `allow-stale`: a flag that enables running the script against a stale replica.
+* `allow-stale`: a flag that enables running the script against a stale replica when the `replica-serve-stale-data` config is set to `no` .
 
-    By default, Redis prevents data consistency problems from using old data by having stale replicas return a runtime error.
-    In cases where the consistency is a lesser concern, this flag allows stale Redis replicas to run the script.
+    Redis can be set to prevent data consistency problems from using old data by having stale replicas return a runtime error.
+    In cases where the consistency is a lesser concern, this flag allows stale Redis replicas to run the script anyway.
```
MeirShpilraien, line 479: Should we mention that in such case, commands that are not allowed while stale will be deny with error?

MeirShpilraien (commented):
:+1: 2 small comments.

docs/manual/programmability/lua-api.md
oranagra: i don't know.. this whole section is about a new feature of 7.0 (flags). so if we document that, then it's about a behavior change between 7.0.0 and 7.0.1.
but then, maybe we should document that in despite what the text says, in 7.0.0 `no-writes` did not imply `allow-oom`, and that you had to add it explicitly?
i think this could cause a lot of confusion (the text will be hard to follow).

maybe i'll just add some `history` section and keep it separate from the rest of the text.

oranagra: actually, i now recall that #1953 adds a new section specific to read-only scripts (currently located in the wrong file IMHO).
i guess we should squash these two PRs together, or merge the other and then modify the text again here.

MeirShpilraien (approved):

itamarhaber (approved):

PR #1955, opened 2022-05-22
Remove extra title in key-specs page

enjoy-binbin:
We do have a `title: "Command key specifications"` in the front.
So this one will render an extra title, remove it.


enjoy-binbin:
see https://redis.io/docs/reference/key-specs/

image

PR #1964, opened 2022-05-27
Update command-list.md

ofirluzon:
Fix ACL categories link

### DIFF
--- a/commands/command-list.md
+++ b/commands/command-list.md
@@ -3,7 +3,7 @@ Return an array of the server's command names.
 You can use the optional _FILTERBY_ modifier to apply one of the following filters:
 
  - **MODULE module-name**: get the commands that belong to the module specified by _module-name_.
- - **ACLCAT category**: get the commands in the [ACL category](/topics/acl) specified by _category_.
+ - **ACLCAT category**: get the commands in the [ACL category](/manual/security/acl/#command-categories) specified by _category_.
  - **PATTERN pattern**: get the commands that match the given glob-like _pattern_.
 
 @return
--- a/commands/command-list.md
+++ b/commands/command-list.md
@@ -3,7 +3,7 @@ Return an array of the server's command names.
 You can use the optional _FILTERBY_ modifier to apply one of the following filters:
 
  - **MODULE module-name**: get the commands that belong to the module specified by _module-name_.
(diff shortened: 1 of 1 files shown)


enjoy-binbin (approved):
i did not actually test it, but LGTM, thanks

commands/command-list.md
```diff
@@ -3,7 +3,7 @@ Return an array of the server's command names.
 You can use the optional _FILTERBY_ modifier to apply one of the following filters:
 
  - **MODULE module-name**: get the commands that belong to the module specified by _module-name_.
- - **ACLCAT category**: get the commands in the [ACL category](/topics/acl) specified by _category_.
+ - **ACLCAT category**: get the commands in the [ACL category](/manual/security/acl/#command-categories) specified by _category_.
```
itamarhaber, line 6: Thanks @ofirluzon (& @enjoy-binbin).

I've enabled the preview for this PR (soon to be automatic for all 🤞) so you can check it out. Currently, it doesn't resolve the issue.

I can think of two ways to go about this:

1. Fix the link as below, and/or
2. Add an entry in the website's redirects file for backward compatibility (@nermiller WDUT?)

```suggestion
 - **ACLCAT category**: get the commands in the [ACL category](/docs/manual/security/acl/#command-categories) specified by _category_.
```

nermiller: Yes, I think we need to update the path as specified here. Let's discuss redirects.

zuiderkwast: I think we need both 1 and 2.

oranagra: i don't understand what's the problem, or maybe it was already fixed.
the page at https://redis.io/commands/command-list/ works nicely to me.

zuiderkwast: Currently there are two redirects happening. The link "ACL category" on the page points to `/topics/acl`, which returns a 301-redirect to `/topics/acl/` which returns a HTML page with the tags `<link rel=canonical href=/docs/manual/security/acl/>` and `<meta http-equiv=refresh content="0; url=/docs/manual/security/acl/">` which effectively is a HTML-layer redirect to the actual page. It works but it adds latency on a slow connection. However, I also don't like the idea of manually reviewing all these updated links.

nermiller: @ofirluzon , please accept @itamarhaber's suggestion. We will add a PR for the redirect. Thanks!

PR #1965, opened 2022-05-28
Add hdt3213/godis into libraries/go

HDT3213:
Add hdt3213/godis into libraries/go


zuiderkwast (approved):
Looks good. It has 1.9k stars so I think we should add it.

HDT3213:
Thank you very much. We will continue to improve godis

PR #1810, opened 2022-03-07
Added 'm' to example String in notifications

aangiel:

PR #1968, opened 2022-05-31
Repeaiting: added 'm' in sample for every possible event

aangiel:
Repeating case the original PR (#1810) became too old


enjoy-binbin (approved):

docs/manual/keyspace-notifications.md
```diff
@@ -82,7 +82,7 @@ will be delivered regardless of the rest of the string.
 For instance to enable just Key-space events for lists, the configuration
 parameter must be set to `Kl`, and so forth.
 
-The string `KEA` can be used to enable every possible event.
+The string `KEAm` can be used to enable every possible event.
```
oranagra, line 85: now that we have a notification for new keys (https://github.com/redis/redis/pull/10512), this will get outdated again.
how about keeping the `KEA` and instead change the text to so that it doesn't say "every possible event"?

PR #1970, opened 2022-06-01
fix BITFIELD_RO command synopsis

ranshid:
Current documentation only states a single GET tocken is needed
which is not true.
marking the command with multiple_token

### DIFF
--- a/commands.json
+++ b/commands.json
@@ -736,6 +736,7 @@
                 "type": "block",
                 "token": "GET",
                 "multiple": true,
+                "multiple_token": true,
                 "arguments": [
                     {
                         "name": "encoding",


ranshid:
@oranagra do you agree we can merge this fix (not urgent in any way...)

oranagra:
The fix seems right, and indeed bitfield doesn't seem to have this problem since these options are in a group and the multiple is applied to it. 

However this file is now auto generated from the json files in the redis project, so please submit the fix there.

ranshid:
> The fix seems right, and indeed bitfield doesn't seem to have this problem since these options are in a group and the multiple is applied to it.
> 
> However this file is now auto generated from the json files in the redis project, so please submit the fix there.

@oranagra https://github.com/redis/redis/pull/10820/
closing this PR then

PR #1971, opened 2022-06-01
Spellcheck

slorello89:
@itamarhaber 
This adds a GitHub action to perform spell-checking against any new PR to master for an update to the docs. Spellchecker is a bit different than the old aspell spell checker called out in the README (I'm assuming that's gone as I don't see the make file anymore?) This also does some light grammatical highlighting as well, e.g. it checks for correct usage of indefinite articles, and detects accidental word duplication (those can be disabled if we think they're dumb but I think they're useful at least).

I took the liberty of generating and curating a list of exceptions needed for the site (there were a few legitimate errors that I went through and corrected)


slorello89:
closing while I debug the new action.

PR #1989, opened 2022-06-08
XADD redis entry clarfication

kamran-redis:
Believe list describes stream entry better and is consistent with behaviour


itamarhaber (approved):
Thanks @kamran-redis - merging.

PR #1991, opened 2022-06-08
7.0.1 auto-gen doc updates

oranagra:


docs/reference/modules/modules-api-ref.md
```diff
@@ -5900,7 +5901,7 @@ filter applies in all execution paths including:
 
 1. Invocation by a client.
 2. Invocation through [`RedisModule_Call()`](#RedisModule_Call) by any module.
-3. Invocation through Lua `redis.call()`.
+3. Invocation through Lua `redis.`call()``.
```
oranagra, line 5904: @itamarhaber @slorello89 there's some issue with with the auto-generated file after the edit here:
https://github.com/redis/redis/pull/10812
meanwhile, i'm reverting it manually, please try to fix.
```suggestion
3. Invocation through Lua `redis.call()`.
```

PR #2004, opened 2022-06-20
Add Redis GUI intellij Idea by eltonsandre

eltonsandre:


itamarhaber:
Hello @eltonsandre 

I'm afraid we can't accept this request as it appears related to a closed-source commercial product.

PR #2024, opened 2022-07-01
Fixes redis/redis-doc#2022 ACL LIST to ACL GETUSER

noahehall:
- *ACL LIST* designed for computers
- *ACL GETUSER* is more human readable


itamarhaber (approved):
Thanks @noahehall - merging.

PR #2025, opened 2022-07-03
Remote LFU from volatile-ttl policy description

enjoy-binbin:


itamarhaber (approved):
Good catch @enjoy-binbin - thanks!

PR #2035, opened 2022-07-10
Add integer type to the return of LCS LEN

enjoy-binbin:
It is better to add type display, also same as LCS IDX below


itamarhaber (approved):
Agreed - thanks!

PR #2036, opened 2022-07-10
Fix typo in RESP protocol example

zeroflaw:
While writing a RESP parser in Rust, I copied some of the examples from the docs to use in my tests. 

When adding validation to my code I noticed a bulk string had the incorrect length.


itamarhaber (approved):
Thanks @zeroflaw - merging!

/cc @banker @rrelledge next time we touch examples to remove foobars please note these technical tidbits ;)

PR #2048, opened 2022-07-19
Fixes command arguments doc

itamarhaber:


docs/reference/command-arguments.md
```diff
@@ -14,8 +14,7 @@ This key stores an array that describes the command's arguments.
 Every element in the _arguments_ array is a map with the following fields:
 
 * **name:** the argument's name, always present.
-  The name of an argument is given for identification purposes alone.
-  It isn't displayed during the command's syntax rendering.
+  The name is displayed during the command's syntax rendering as a placeholder for user input.
```
guybe7, line 17: unless it's a "container" argument (not sure about the name) - i.e. "oneof" or "block"

guybe7:
```
For arguments types other than _oneof_ and _block_, this is a string that describes the value in the command's syntax.
``` 
(from below)

PR #2056, opened 2022-07-26
kiss to diss

Invidia19:


itamarhaber:
Hi @Invidia19 - thanks for the suggestion, but I'm not sure it is needed as the current works well enough, don't you think?

Invidia19:
Yeah i agree, right now the existing one works just fine.

itamarhaber:
Excellent - so let's keep it simply simple. Closing :)

PR #2063, opened 2022-08-03
Fix multi dimensional indexes miscalculation

lucacillario:
Fixed multi dimensional indexes miscalculation.


itamarhaber (approved):
Off-by-one are the hardest - thank you so much @lucacillario. you have a keen eye.

PR #2064, opened 2022-08-03
Fix `ssubscribe` examples

j178:


itamarhaber (approved):
Very true - thanks @j178, merging.

itamarhaber:
This fixes the break in 7.0.1 (https://github.com/redis/redis/pull/10792)

PR #2071, opened 2022-08-06
Correct docs

manojchandrashekar:
Incorrect name for the library. (Python --> Node.js)

### DIFF
--- a/libraries/typescript/github.com/redis/redis-om-node.json
+++ b/libraries/typescript/github.com/redis/redis-om-node.json
@@ -1,5 +1,5 @@
 {
     "name": "Redis OM for Node.js",
-    "description": "Object mapping, and more, for Redis and Node.js.",
+    "description": "Object mapping, and more, for Redis and Node.js. Written in TypeScript.",
     "recommended": true
 }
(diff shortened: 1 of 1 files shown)


libraries/typescript/github.com/redis/redis-om-node.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "Redis OM for Python",
+    "name": "Redis OM for Node.js",
     "description": "Object mapping, and more, for Redis and Node.js.",
```
enjoy-binbin, line 3: may or may not, just a thought
```suggestion
    "description": "Object mapping, and more, for Redis and Node.js. Written in TypeScript.",
```

enjoy-binbin (approved):
thanks, LGTM

libraries/typescript/github.com/redis/redis-om-node.json
manojchandrashekar: Hey, I am not sure if it matters that it's written in TypeScript. I don't think it is needed to be specified.

Please overwrite if you think it is needed. :)

itamarhaber (approved):

PR #2081, opened 2022-08-12
updated install-redis-on-windows

KTK27YT:
updated install-redis-on-windows remove the link to follow install on linux. Since the link is dead and results in a 404 error


KTK27YT:
I followed the steps below and it worked,moreover the link doesn't work it leads to a 404 error

KTK27YT:
Resolves issue code #2112

itamarhaber:
Hi @KTK27YT - this was fixed by providing the correct link (instead of removing it altogether) via #2097.

Thanks for the PR, closing as it is the wrong fix.

PR #2093, opened 2022-08-19
[BUG FIX] Correct DUMP command example

### LINKED ISSUE #2076: DUMP: Example returns a utf-8 codec error.
The example on the page returns an error: `(error) 'utf-8' codec can't decode byte 0xc0 in position 1: invalid start byte`. 

It doesn't look like that's intended.

Screenshot from https://redis.io/commands/dump/:
image

nermiller:
Fixes #2076, corrects example output

### DIFF
--- a/commands/dump.md
+++ b/commands/dump.md
@@ -27,8 +27,9 @@ If `key` does not exist a nil bulk reply is returned.
 
 @examples
 
-```cli
-SET mykey 10
-DUMP mykey
-
-```
+{{< highlight bash >}}
+> SET mykey 10
+OK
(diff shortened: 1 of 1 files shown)


itamarhaber:
@nermiller the correct fix is to resolve https://github.com/redis/redis-clinterwebz/issues/12

nermiller:
> @nermiller the correct fix is to resolve [redis/redis-clinterwebz#12](https://github.com/redis/redis-clinterwebz/issues/12)

Agree. Consulted with @slorello89 and this is a temporary fix.

PR #2100, opened 2022-08-25
Add note about LASTSAVE without previous BGSAVE

MichaHoffmann:
A small note about an edge case that can lead the described procedure to hang indefinitely if BGSAVE returned the same second that redis started.

### DIFF
--- a/commands/lastsave.md
+++ b/commands/lastsave.md
@@ -1,7 +1,7 @@
 Return the UNIX TIME of the last DB save executed with success.
 A client may check if a `BGSAVE` command succeeded reading the `LASTSAVE` value,
 then issuing a `BGSAVE` command and checking at regular intervals every N
-seconds if `LASTSAVE` changed.
+seconds if `LASTSAVE` changed. Redis considers the database saved successfully at startup.
 
 @return
 


enjoy-binbin (commented):
the note looks good, but i guess we also need to mention SAVE if we mention BGSAVE
like SAVE will also update the lastsave value

anyway, if we need the note, i think we should drop the BGSAVE,
and just mention the lastsave default value like: (At startup we consider the DB saved.)

MichaHoffmann:
> the note looks good, but i guess we also need to mention SAVE if we mention BGSAVE like SAVE will also update the lastsave value
> 
> anyway, if we need the note, i think we should drop the BGSAVE, and just mention the lastsave default value like: (At startup we consider the DB saved.)

Thanks! good points, i adapted the note!

commands/lastsave.md
```diff
@@ -1,7 +1,7 @@
 Return the UNIX TIME of the last DB save executed with success.
 A client may check if a `BGSAVE` command succeeded reading the `LASTSAVE` value,
 then issuing a `BGSAVE` command and checking at regular intervals every N
-seconds if `LASTSAVE` changed.
+seconds if `LASTSAVE` changed. Redis considers the database saved successfully at startup.
```
enjoy-binbin, line 4: Let's also mention this default value? Otherwise the sentence looks weird and lacks context
something like your first commit: will return the UNIX TIME that the redis server started.

enjoy-binbin:
yep, i think we can update it and merge it

commands/lastsave.md
nermiller, line 4:
``` suggestion
seconds if `LASTSAVE` changed. Redis considers the database saved successfully at startup. `LASTSAVE` returns the UNIX TIME of the server start.
```

nermiller (approved):
@enjoy-binbin how's this?

commands/lastsave.md
enjoy-binbin: look goods, i am not very good at english, how about something like this?

Redis considers the database saved successfully at startup and `LASTSAVE` will return the UNIX TIME of the server start by default.

nermiller: Updated!

enjoy-binbin: thanks for picking this up, LGTM

MichaHoffmann:
Thanks! I kinda forgot about this for a bit! Thanks for picking it up!

PR #2101, opened 2022-08-26
fix the description about aof-write in latency monitor

uvletter:


chenyang8094 (approved):
LGTM
@oranagra Please have a look.

oranagra (approved):
summary of changes:
these 4 events are measuring the same thing, in different states,
they all measure `write` (not `fsync` which was stated).
and the `pending-fsync` one is while there's a pending `fsync` (not `write` which was stated).

in essence, fsync and write were swapped.

PR #2104, opened 2022-08-28
Adds a note about ping errors

itamarhaber:


oranagra (approved):
LGTM.
could be nice to mention the change in redis 7, but it's hard if we don't wanna mention the exact errors (`-LOADING`), since we don't do that for any other command.

PR #2107, opened 2022-08-31
fix a markdown link syntax issue

atuttle:
Ran into this bit of raw markdown while reading the docs this morning:

image


enjoy-binbin (approved):
thanks for the fix

itamarhaber (approved):
Thanks for spotting, reporting and fixing this @atuttle - merging.

PR #2111, opened 2022-09-02
remove stale info about config get requirepass

andyxning:
currently `config get requirepass` still returns raw password instead hashed one.


enjoy-binbin (approved):
thanks, i tested it in 6.0 and 7.0, you are right, this looks like an outdated document
let's get another approval in case we're missing something

itamarhaber:
Thanks @andyxning - good find, let's merge this fix.

PR #2120, opened 2022-09-13
Adds RESP3 to protocol specification

### LINKED ISSUE #1920: RESP protocol spec -- RESP3 missing
There was a link to the RESP3 documentation added earlier, but it seems to have been lost when the new website was activated.

Additionally, this makes me wonder: Are there more things that have been lost in the switch?

itamarhaber:
Fixes #1920


docs/reference/protocol-spec.md
```diff
@@ -218,28 +255,34 @@ integers and a bulk string can be encoded as follows:
...
+The encoding of a Null Array is that of an Array with the length of -1, i.e.:
 
-Nested arrays are possible in RESP. For example a nested array of two arrays
-is encoded as follows:
+    *-1\r\n
+
+A client should return its caller a null object when Redis replies with a Null Array, rather than an empty array.
+This is necessary to distinguish between an empty list and a different condition (for instance, the timeout condition of the `BLPOP` command).
+
+Nested arrays are possible in RESP.
```
leibale, line 284: Arrays can contain any nested data type, not just arrays

docs/reference/protocol-spec.md
```diff
@@ -268,25 +313,278 @@ Null element:
...
+Due to historical reasons, RESP2 features two specially-crafted values for representing null values of Bulk Strings and Arrays.
+This duality has always been a redundancy that added zero semantical value to the protocol itself.
+
+The Null type, introduced in RESP3, aims to fix this wrong.
+{{% /alert %}}}}
+
+<a name="boolean-reply">
+
+### Booleans
+RESP Booleans encode true and false values in the following ways:
```
leibale, line 341: why not use the same format as other types?

Booleans are encoded as follows:

    #<boolean>\r\n

* The "number sight" character (`#`) as the first byte.
* `t` or `f` (`true` or `false`).
* The CRLF terminator.

docs/reference/protocol-spec.md
```diff
@@ -268,25 +313,278 @@ Null element:
...
-## Send commands to a Redis server
+<a name="set-reply"></a>
 
-Now that you are familiar with the RESP serialization format, you can use it to help write a Redis client library. We can further specify
-how the interaction between the client and the server works:
+### Sets
+Sets are somewhat like [Arrays](#arrays) but are unordered and should only contain unique elements.
 
-* A client sends the Redis server a RESP Array consisting of only Bulk Strings.
+RESP Set's encoding is:
```
nermiller, line 496:
```suggestion
RESP set's encoding is:
```

docs/reference/protocol-spec.md
```diff
@@ -268,25 +313,278 @@ Null element:
...
+
+    ~<number-of-elements>\r\n<element-1>...<element-n>
+
+* A tilde (`~`) as the first byte.
+* The number of elements in the Set as a decimal number.
```
nermiller, line 501:
```suggestion
* The number of elements in the set as a decimal number.
```

docs/reference/protocol-spec.md
```diff
@@ -268,25 +313,278 @@ Null element:
...
+* The CRLF terminator.
+* An additional RESP type for every element of the Set.
+
+Clients should return the native set type if it is available in their programming language.
+Alternatively, in the absence of a native set type, an array coupled with type information can be used (in C, for example).
+
+<a name="push-event"></a>
+
+### Pushes
+RESP's Pushes contain out-of-band data.
```
nermiller, line 511:
```suggestion
RESP's pushes contain out-of-band data.
```

docs/reference/protocol-spec.md
```diff
@@ -268,25 +313,278 @@ Null element:
...
+They are an exception to the protocol's request-response model and provide a generic _push mode_ for connections.
+
+Push events are encoded similarly to [Arrays](#arrays), differing only in their first byte:
+
+    ><number-of-elements>\r\n<element-1>...<element-n>
+
+* A greater-than sign (`>`) as the first byte.
+* The number of elements as a decimal number.
+* The CRLF terminator.
+* An additional RESP type for every element of the Push Event.
```
nermiller, line 521:
```suggestion
* An additional RESP type for every element of the push event.
```

docs/reference/protocol-spec.md
```diff
@@ -268,25 +313,278 @@ Null element:
...
+Clients should return the native set type if it is available in their programming language.
+Alternatively, in the absence of a native set type, an array coupled with type information can be used (in C, for example).
+
+<a name="push-event"></a>
+
+### Pushes
+RESP's Pushes contain out-of-band data.
+They are an exception to the protocol's request-response model and provide a generic _push mode_ for connections.
+
+Push events are encoded similarly to [Arrays](#arrays), differing only in their first byte:
```
nermiller, line 514:
```suggestion
Push events are encoded similarly to [arrays](#arrays), differing only in their first byte:
```

docs/reference/protocol-spec.md
```diff
@@ -268,25 +313,278 @@ Null element:
...
+
+    ><number-of-elements>\r\n<element-1>...<element-n>
+
+* A greater-than sign (`>`) as the first byte.
+* The number of elements as a decimal number.
+* The CRLF terminator.
+* An additional RESP type for every element of the Push Event.
+
+Pushed data may precede or follow any of RESP's data types but never inside them.
+That means a client won't find Push data in the middle of a Map reply, for example.
```
nermiller, line 524:
```suggestion
That means a client won't find push data in the middle of a map reply, for example.
```

docs/reference/protocol-spec.md
```diff
@@ -268,25 +313,278 @@ Null element:
...
+It also means that pushed data may appear before or after a command's reply, as well as by itself (without calling any command).
+
+Clients should react to Pushes by invoking a callback that implements their handling of the pushed data.
```
nermiller, line 527:
```suggestion
Clients should react to pushes by invoking a callback that implements their handling of the pushed data.
```

docs/reference/protocol-spec.md
```diff
@@ -268,25 +313,278 @@ Null element:
...
+* **id**: the connection's identifier (ID).
+* **mode**: "standalone", "sentinel" or "cluster".
+* **role**: "master" or "replica".
+* **modules**: list of loaded modules as an Array of Bulk Strings.
+
+## Sending commands to a Redis server
+Now that you are familiar with the RESP serialization format, you can use it to help write a Redis client library.
+We can further specify how the interaction between the client and the server works:
+
+* A client sends the Redis server a RESP [Array](#arrays) consisting of only Bulk Strings.
```
nermiller, line 581:
```suggestion
* A client sends the Redis server a RESP [array](#arrays) consisting of only bulk strings.
```

docs/reference/protocol-spec.md
```diff
@@ -268,25 +313,278 @@ Null element:
...
 * A Redis server replies to clients, sending any valid RESP data type as a reply.
 
-So for example a typical interaction could be the following.
+So, for example, a typical interaction could be the following.
 
-The client sends the command **LLEN mylist** in order to get the length of the list stored at key *mylist*. Then the server replies with an Integer reply as in the following example (C: is the client, S: the server).
+The client sends the command `LLEN mylist` to get the length of the list stored at the key _mylist_.
+Then the server replies with an [Integer](#integers) reply as in the following example (`C:` is the client, `S:` the server).
```
nermiller, line 587:
```suggestion
Then the server replies with an [integer](#integers) reply as in the following example (`C:` is the client, `S:` the server).
```

docs/reference/protocol-spec.md
```diff
@@ -299,51 +597,44 @@ The client sends the command **LLEN mylist** in order to get the length of the l
...
+Pipelining is supported, so multiple commands can be sent with a single write operation by the client.
+The client can skip reading replies and continue to send the commands one after the other.
 All the replies can be read at the end.
 
 For more information, see [Pipelining](/topics/pipelining).
 
 ## Inline commands
+Sometimes you may need to send a command to the Redis server but only have `telnet` available.
+While the Redis protocol is simple to implement, it is not ideal for interactive sessions, and `redis-cli` may not always be available.
+For this reason, Redis also accepts commands in the **inline command** format.
```
nermiller, line 610:
```suggestion
For this reason, Redis also accepts commands in the _inline command_ format.
```

docs/reference/protocol-spec.md
```diff
@@ -299,51 +597,44 @@ The client sends the command **LLEN mylist** in order to get the length of the l
...
 
-While the Redis protocol is human readable and easy to implement, it can
-be implemented with a performance similar to that of a binary protocol.
+While the Redis protocol is human-readable and easy to implement, its implementation can exhibit performance similar to that of a binary protocol.
 
-RESP uses prefixed lengths to transfer bulk data, so there is
-never a need to scan the payload for special characters, like with JSON, nor to quote the payload that needs to be sent to the
-server.
+RESP uses prefixed lengths to transfer bulk data.
+That makes scanning the payload for special characters unnecessary (unlike parsing JSON , for example).
```
nermiller, line 630:
```suggestion
That makes scanning the payload for special characters unnecessary (unlike parsing JSON, for example).
```

docs/reference/protocol-spec.md
```diff
@@ -299,51 +597,44 @@ The client sends the command **LLEN mylist** in order to get the length of the l
...
+For the same reason, quoting and escaping the payload isn't needed.
 
-The Bulk and Multi Bulk lengths can be processed with code that performs
-a single operation per character while at the same time scanning for the
-CR character, like the following C code:
+Reading the length of Aggregate types (for example, Bulk Strings or Arrays) can be processed with code that performs a single operation per character while at the same time scanning for the CR character.
```
nermiller, line 633:
```suggestion
Reading the length of aggregate types (for example, bulk strings or arrays) can be processed with code that performs a single operation per character while at the same time scanning for the CR character.
```

docs/reference/protocol-spec.md
```diff
@@ -299,51 +597,44 @@ The client sends the command **LLEN mylist** in order to get the length of the l
...
 
-```
+This C code example illustrates the above:
```
nermiller, line 635:
```suggestion
Example (in C):
```

nermiller (approved):
LGTM with some minor edits

zuiderkwast (approved):
Nice! If it takes a long time to handle all the review comments, I suggest merging it as is and handling the comments afterwards in a separate PR. This specification is missing since forever and it'd be better to have something in the meantime IMO.

nermiller:
@itamarhaber since this is an important change, mind resolving the conflicts and merging? We can take care of the feedback later, as suggested by @zuiderkwast.

docs/reference/protocol-spec.md
```diff
@@ -1,214 +1,251 @@
...
 aliases:
     - /topics/protocol
 ---
 
-Redis clients use a protocol called **RESP** (REdis Serialization Protocol) to communicate with the Redis server. While the protocol was designed specifically for Redis, it can be used for other client-server software projects.
+Redis clients use a protocol called **RESP** (REdis Serialization Protocol) to communicate with the Redis server.
+While the protocol was designed specifically for Redis, it can be used for other client-server software projects.
 
-RESP is a compromise between the following things:
+RESP is a compromise among the following considerations:
```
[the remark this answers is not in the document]
itamarhaber: I prefer leaving the "compromise" there - it explains (some of) the decisions that were made.

docs/reference/protocol-spec.md
```diff
@@ -1,214 +1,251 @@
...
-    "-Error message\r\n"
+    -Error message\r\n
 
-Error replies are only sent when something goes wrong, for instance if
-you try to perform an operation against the wrong data type, or if the command
-does not exist. The client should raise an exception when it receives an Error reply.
+Redis replies with an Error only when something goes wrong.
+For example, if you try to perform an operation against the wrong data type, or if the command
+does not exist.
+The client should raise an exception when it receives an Error reply.
```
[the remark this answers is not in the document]
itamarhaber: This is a recommendation to client authors, not a requirement.

docs/reference/protocol-spec.md
```diff
@@ -218,28 +255,34 @@ integers and a bulk string can be encoded as follows:
...
 
-Null Arrays exist as well and are an alternative way to
-specify a Null value (usually the Null Bulk String is used, but for historical
-reasons we have two formats).
+<a name="nil-array-reply"></a>
 
-For instance, when the `BLPOP` command times out, it returns a Null Array
-that has a count of `-1` as in the following example:
+#### Null Arrays
+Whereas RESP3 has a dedicated data type for [null values](#nulls), RESP2 has no such type. Instead, due to historical reasons, the representation of null values in RESP2 is via predetermined forms of the [Bulk Strings](#bulk-strings) and Arrays types.
```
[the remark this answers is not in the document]
itamarhaber: You forgot to link the arrays like you did above :)

docs/reference/protocol-spec.md
```diff
@@ -1,214 +1,251 @@
...
+While RESP is technically non-TCP specific, the protocol is used exclusively with TCP connections (or equivalent stream-oriented connections like Unix sockets) in the context of Redis.
 
-RESP can represent a Null value using a special variation of Bulk Strings or Array as specified later.
+## Request-Response model
+The Redis server accepts commands composed of different arguments.
+Then, the server processes the command and sends the reply back to the client.
+
+This is the simplest model possible; however, there are some exceptions:
+
+* Redis supports [pipelining](#multiple-commands-and-pipelining).
```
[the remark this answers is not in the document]
itamarhaber: Agreed, but this is a list of exceptions the simple req->rep model, and pipelining is an exception, no?

docs/reference/protocol-spec.md
```diff
@@ -1,214 +1,251 @@
...
 <a name="error-reply"></a>
 
-## RESP Errors
-
-RESP has a specific data type for errors. They are similar to
-RESP Simple Strings, but the first character is a minus '-' character instead
-of a plus. The real difference between Simple Strings and Errors in RESP is that clients treat errors
-as exceptions, and the string that composes
-the Error type is the error message itself.
+### Errors
```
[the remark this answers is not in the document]
itamarhaber: So, following the same logic, we should have "Simple Error" and "Bulk Error" instead of "Error" and "Blob Error" to match strings. That's a nice consistency, but I wonder what would break if change names this far into the game.

docs/reference/protocol-spec.md
```diff
@@ -1,214 +1,251 @@
...
+Many Redis commands return RESP Integers, including `INCR`, `LLEN`, and `LASTSAVE`.
+An integer, by itself, has no special meaning other than in the context of the command that returned it.
+For example, it is an incremental number for `INCR`, a UNIX timestamp for `LASTSAVE`, and so forth.
+However, the returned integer is guaranteed to be in the range of a signed 64-bit integer.
 
-There is no special meaning for the returned integer. It is just an
-incremental number for `INCR`, a UNIX time for `LASTSAVE`, and so forth. However,
-the returned integer is guaranteed to be in the range of a signed 64-bit integer.
+In some cases, Integers can represent true and false Boolean values.
+For instance, commands such as `EXISTS` and `SISMEMBER` return 1 for true and 0 for false.
```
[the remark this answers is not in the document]
itamarhaber: Thanks - removed.

docs/reference/protocol-spec.md
```diff
@@ -1,214 +1,251 @@
...
-`EXISTS`, `INCR`, `INCRBY`, `DECR`, `DECRBY`, `DBSIZE`, `LASTSAVE`,
-`RENAMENX`, `MOVE`, `LLEN`, `SADD`, `SREM`, `SISMEMBER`, `SCARD`.
-
-<a name="nil-reply"></a>
 <a name="bulk-string-reply"></a>
 
-## RESP Bulk Strings
+### Bulk Strings
+A Bulk String represents a single binary-safe string.
+The string can be of any size, but Redis limits it by default up to 512MB (see the `proto-max-bulk-len` configuration directive).
```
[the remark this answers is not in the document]
itamarhaber: :+1:

docs/reference/protocol-spec.md
```diff
@@ -1,214 +1,251 @@
...
-use RESP Arrays as their replies. An example is the `LRANGE` command that
-returns elements of a list.
+RESP Arrays' encoding uses the following format:
 
-RESP Arrays are sent using the following format:
+    *<number-of-elements>\r\n<element-1>...<element-n>
 
-* A `*` character as the first byte, followed by the number of elements in the array as a decimal number, followed by CRLF.
+* An asterisk (`*`) as the first byte.
+* The number of elements in the Array as a decimal number.
```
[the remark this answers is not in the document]
itamarhaber: Pushed it even further into "as the string representation of an integer."

docs/reference/protocol-spec.md
itamarhaber: All aggregates can do that, but I understand your point and have added some words to that effect.

docs/reference/protocol-spec.md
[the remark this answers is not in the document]
zuiderkwast: In [RESP3.md](https://github.com/redis/redis-specifications/blob/master/protocol/RESP3.md) it is "Simple Error" and "Blob Error".

I don't think it's a big deal to rename them or even mention both names (e.g. "bulk error AKA blob error"), but people will most certainly use both the old and the new terms forever, which is fine IMO as long as none of the terms are ambiguous.

docs/reference/protocol-spec.md
itamarhaber: I did it to break out of the routine. I guess that's unacceptable. Fixed.

itamarhaber:
@zuiderkwast this isn't ready to merge, please wait.

docs/reference/protocol-spec.md
```diff
@@ -1,214 +1,252 @@
...
+
+* Redis requests can be [pipelined](#multiple-commands-and-pipelining).
+  Pipelining enables clients to send multiple commands at once and wait for replies later.
+* When a RESP2 connection subscribes to a [Pub/Sub](/docs/manual/pubsub) channel, the protocol changes semantics and becomes a *push* protocol.
+  The client no longer requires sending commands because the server will automatically send new messages to the client (for the channels the client is subscribed to) as soon as they are received.
+* The `MONITOR` command.
+  Invoking the `MONITOR` command switches the connection to an ad-hoc push mode.
+  The protocol of this mode is not specified but is obvious to parse.
+* [Protected mode](/docs/management/security/#protected-mode).
+  Connections opened from a non-loopback address to a Redis while in protected mode are automatically sent a `-DENIED` reply and terminated by the server.
```
leibale, line 68: Please make it clearer that the DENIED error will be emitted without issuing any command

itamarhaber: Better?

docs/reference/protocol-spec.md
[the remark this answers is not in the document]
itamarhaber: Fixed as "simple" and "bulk"

docs/reference/protocol-spec.md
```diff
@@ -365,11 +364,20 @@ Here's the encoding of the number 1.23:
 
     ,1.23\r\n
 
-The floating point number must start with a digit, even if it is zero (0).
-Exponential notation is not supported.
+A floating point number may be made up of up to three parts:
+
+    <integral-part>[.<fractional-part>[<exponent-part]]
```
zuiderkwast, line 369: There can be an exponent part without a fractional part.

```suggestion
    <integral-part>[.<fractional-part>][<exponent-part]
```

docs/reference/protocol-spec.md
```diff
@@ -351,8 +351,7 @@ RESP booleans are encoded as follows:
 <a name="double-reply"></a>
 
 ### Doubles
-The Double RESP type encodes a double-precision floating point value.
-
+The Double RESP type encodes a double-precision floating point valu
```
zuiderkwast: Oops?

docs/reference/protocol-spec.md
```diff
@@ -365,11 +364,20 @@ Here's the encoding of the number 1.23:
...
+
+The floating point number always begins with an integral.
+The integral part is made of one or more decimal digits.
+
+A fractional part may follow the integral.
+The fractional part begins with a dot (`.`), followed by one or more decimal digits.
 
-However, the decimal part is optional.
-The integer value of ten (10) can, therefore, be RESP-encoded both as an integer as well as a double:
+An exponent part may follow the fractional.
```
zuiderkwast, line 377: Perhaps rephrase slightly? The exponent part can also follow the integral part, if there are no decimal fractions.

zuiderkwast (commented):
Just a few comments on the last commit about floats.

docs/reference/protocol-spec.md
```diff
@@ -678,3 +678,9 @@ Then, the bulk data can be read with a single read operation that doesn't inspec
 Finally, the remaining CR and LF characters are discarded without additional processing.
 
 While comparable in performance to a binary protocol, the Redis protocol is significantly more straightforward to implement in most high-level languages, reducing the number of bugs in client software.
+
+## Tips for Redis client authors
+
+* For testing purposes, use [Lua's type conversions](/topics/lua-api#lua-to-resp3-type-conversion) to have Redis reply with any RESP2/RESP3 needed.
+  As an example, a RESP3 double can be generated like so:
+    EVAL "return { double = 6.379e-5 }" 0
```
itamarhaber: @mgravell @zuiderkwast @leibale make this beefier if you can :)

docs/reference/protocol-spec.md
itamarhaber: 👍 via 6280997

mgravell:
@itamarhaber first time I've found that one set of conversions (`lua-to-resp3-type-conversion`); very useful! but: is there some server version context needed in there? testing shows a few oddities:

- `return true` and `return false` gives me *integer* on both RESP2 and RESP3; I'm testing on a v6 server - is this a v7 change?
- `return { double = 42 }` gives me *double* on RESP3, but gives me *bulk string* on RESP2; this seems odd, since in RESP3.md, *double* is defined as a variant of simple-string, not bulk-string (i.e. no length prefix) - intentional?

itamarhaber:
@mgravell 

> first time I've found...

TBAH, same here 😆 

> `return true`, `return false`... is this a v7 change?

This is a conscious decision made to minimize immediate script breakage.
Once we feel most have moved to RESP3, we can debate making the breaking change.
It exists both in v6 as well as in v7.
One has to both use a RESP3 connection && instruct Lua to return RESP3 like so:

    EVAL "redis.setresp(3) return true" 0

> `return { double = 42 }`...`since in RESP3.md`...

I think this is the same thing, here's a comparison between old and new:

```
127.0.0.1:6379> HELLO 2
...
127.0.0.1:6379> ZADD z 42.1 foo
(integer) 1
127.0.0.1:6379> ZRANGE z 0 -1  WITHSCORES
1) "foo"
2) "42.100000000000001"
127.0.0.1:6379> EVAL "return redis.call('ZRANGE', 'z', 0, -1, 'WITHSCORES')" 0
1) "foo"
2) "42.100000000000001"

# No one should really do this because why?
127.0.0.1:6379> EVAL "redis.setresp(3) return redis.call('ZRANGE', 'z', 0, -1, 'WITHSCORES')" 0
1) 1) "foo"
   2) "42.100000000000001"

127.0.0.1:6379> HELLO 3
...
127.0.0.1:6379> ZRANGE z 0 -1  WITHSCORES
1) 1) "foo"
   2) (double) 42.100000000000001
127.0.0.1:6379> EVAL "return redis.call('ZRANGE', 'z', 0, -1, 'WITHSCORES')" 0
1) "foo"
2) "42.100000000000001"
127.0.0.1:6379> EVAL "redis.setresp(3) return redis.call('ZRANGE', 'z', 0, -1, 'WITHSCORES')" 0
1) 1) "foo"
   2) (double) 42.100000000000001
```





I'm not sure I understand this one, so I'll dump what I know:

* RESP2 means no doubles, so always bulk strings (e.g. `ZRANGE ... WITHSCORES`) forcing the client author to divine it from the docs.
* RESP3 has doubles, and their encoding is spec-ed even more rigorously by my last commit here. They are, indeed, "simple" in the sense of not being length-prefixed.
* When you `return 42.1` with the defaults, the pre-v6 rules would give you `(integer 42)`. To force a stringifyied double, you'd had to `return tostring(42.1)`. This remains the same.
* You can, in v6+, switch to RESP3 at the connection level. At this point, commands may return RESP3 values, but from the POV of the Lua sandbox you're still in RESP2.
* If you `redis.setresp(3)`, the script will both

docs/reference/protocol-spec.md
```diff
@@ -354,29 +362,19 @@ RESP booleans are encoded as follows:
 The Double RESP type encodes a double-precision floating point value.
 Doubles are encoded as follows:
 
-    ,<floating-point-number>\r\n
+    ,[<sign>]<integral>[.<fractional>][<E|e>[sign]<exponent>]\r\n
```
leibale, line 365:
```
,[+|-]<integral>[.<fractional>][<E|e>[+|-]<exponent>]\r\n
```

docs/reference/protocol-spec.md
```diff
@@ -85,13 +86,23 @@ Redis generally uses RESP as a [request-response](#request-response-model) proto
...
 Subsequent bytes constitute the type's contents.
-The `\r\n` (CRLF) is the protocol's _terminator_, which always separates its parts.
+
+We categorize every RESP data type as either _simple_, _bulk_ or _aggregate_.
+
+Simple types are similar to scalars in programming languages that represent plain literal values. Booleans and Integers are such examples.
+
+RESP strings are either _simple_ or _bulk_.
+Simple strings never contain carriage return (`\r`) or line feed (`\n`) characters.
+Bulk strings can contain any ASCII character and may also be referred to as _binary_ or _blob_.
```
zuiderkwast, line 102: Apart from ASCII characters (bytes in the range 0-127), it may also contain the bytes values 128-255. I think it would be better to talk about binary data than to talk about character encodings here.

docs/reference/protocol-spec.md
```diff
@@ -85,13 +86,23 @@ Redis generally uses RESP as a [request-response](#request-response-model) proto
 * The server replies with a RESP type.
   The reply's type is determined by the command's implementation and possibly by the client's protocol version.
 
-We categorize every RESP data type as either _simple_ or _aggregate_.
-Simple types are similar to scalars in programming languages that represent plain literal values, for example, Booleans.
-Aggregates, such as Arrays and Maps, can have varying numbers of sub-elements and nesting levels.
+RESP is **always** ASCII-encoded (UTF-8/Latin-1).
```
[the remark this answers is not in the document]
itamarhaber: I picked it up from #2203 :) Too much?

docs/reference/protocol-spec.md
itamarhaber: 👍 So what is the proper name for 8-bit ASCII? "Extended"? "winblows-1252"? "ISO 8859-1"?

docs/reference/protocol-spec.md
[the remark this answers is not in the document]
zuiderkwast: Well, in that case I think it can be explained whenever the spec mentions a character, it is a byte in the ASCII interpretation (e.g. \r and \n are the byte values 13 and 10), but anything else is binary data. (UTF-8 and Latin-1 only matters for byte values > 127, but those are never used in the protocol, apart from as binary data.)

docs/reference/protocol-spec.md
PerMildner: ASCII is 7-bit, never 8-bit. Uninterpreted octets (a.k.a. bytes) are not characters (neither ASCII nor Latin 1 nor UTF-8). So, I agree with @zuiderkwast, it is better to never mention "character" (unless a specific encoding is also mentioned), and for raw data it is better to talk about binary bytes.

Of course, for convenience a binary protocol can still use sequences of binary bytes whose values happen to form meaningful English words when interpreted as 7-bit ASCII but the protocol is then still binary, and the bytes are still raw bytes.

docs/reference/protocol-spec.md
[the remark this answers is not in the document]
zuiderkwast: Idea: In the beginning of the document, for example here, describe when RESP is concerned with characters and when it isn't.


```suggestion
Whenever this specification mentions the word "character", it is to be understood as the ASCII interpretation of a single byte. In other situations, such as in the contents of bulk strings, RESP is not concerned with characters and character encodings. The payload of bulk strings can contain any binary data, any sequence of bytes, including null bytes.
```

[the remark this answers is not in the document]
jobol: I propose: "RESP is a **BINARY** protocol using control sequences encoded using standard **ASCII** (Example: the character A is encoded with the binary byte of value 65, the characters <CR>, <LF>, <SP> with 13, 10 and 32 respectively.)"

docs/reference/protocol-spec.md
jobol: Telling that bulk strings should either (choose the right one):
- neither contain <CR> or <LF>
- not contain the sequence <CR><LF>
is enough, IMHO

jobol (commented):
I have not read all the text but I think that the parts related to ASCII can be improved

docs/reference/protocol-spec.md
leibale: @zuiderkwast I think he meant simple string.
@zuiderkwast some decoders might check only for CR (and assume there is LF right after it), I think "neither CR nor LF" is safer.

PR #2129, opened 2022-09-18
Remove duplicated explanation about backing up AOF persistence

hayashier:
There is duplicated explanation about backing up AOF persistence.
https://github.com/redis/redis-doc/blob/master/docs/manual/persistence.md?plain=1#L334-L377

This CR is for removing redundant explanation.


itamarhaber (approved):

itamarhaber:
Hi @hayashier 

Thanks for noticing and fixing this copy-pasta - merging!

PR #2131, opened 2022-09-21
Fix CLUSTER SHARDS output example

enjoy-binbin:
Replace output example with the latest output of
the current CLUSTER SHARDS. The cluster is created with redis utils/create-cluster


itamarhaber (approved):
LGTM - thanks as usual @enjoy-binbin :)

PR #2138, opened 2022-09-28
Fix typo

n7olkachev:


itamarhaber:
Howdy @n7olkachev and thanks for proposing these changes - I totally agree w/ both.

However, the right place to change these is actually over at https://github.com/redis/redis/tree/unstable/src/commands in each command's respective JSON file. Would you like to close this PR and replace it with one against the upstream repo?

n7olkachev:
@itamarhaber done https://github.com/redis/redis/pull/11370

PR #2148, opened 2022-10-05
keep description in update with latest option in bitop command

andyxning:
keep description in update with latest option in `bitop` command


itamarhaber (approved):
Thanks for noticing and fixing @andyxning - merging.

PR #2150, opened 2022-10-06
Clarify description of STREAMS in XREAD

andyxning:
correct description of `STREAMS` in `XREAD`


itamarhaber:
Hello @andyxning and thanks for opening this request.

I'm afraid that I disagree, however, as the `STREAMS` options must be the command's last option (followed by stream keys and IDs, of course). What is your reasoning for proposing this change?

andyxning:
I misunderstands the meaning for `last` is according to `XREAD` command and args instead of parameters.

andyxning:
@itamarhaber

itamarhaber:
@andyxning should we keep this open?

nermiller:
@andyxning just following up on this PR.

docs/data-types/streams-tutorial.md
```diff
@@ -169,7 +169,7 @@ The command that provides the ability to listen for new messages arriving into a
...
 
-In the above command we wrote `STREAMS mystream 0` so we want all the messages in the Stream `mystream` having an ID greater than `0-0`. As you can see in the example above, the command returns the key name, because actually it is possible to call this command with more than one key to read from different streams at the same time. I could write, for instance: `STREAMS mystream otherstream 0 0`. Note how after the **STREAMS** option we need to provide the key names, and later the IDs. For this reason, the **STREAMS** option must always be the last one.
+In the above command we wrote `STREAMS mystream 0` so we want all the messages in the Stream `mystream` having an ID greater than `0-0`. As you can see in the example above, the command returns the key name, because actually it is possible to call this command with more than one key to read from different streams at the same time. I could write, for instance: `STREAMS mystream otherstream 0 0`. Note how after the **STREAMS** option we need to provide the key names, and later the IDs. For this reason, the **STREAMS** option must always be the first one.
```
zuiderkwast, line 172: The original text is correct, but maybe it can be made more clear, e.g.

```suggestion
In the above command we wrote `STREAMS mystream 0` so we want all the messages in the Stream `mystream` having an ID greater than `0-0`. As you can see in the example above, the command returns the key name, because actually it is possible to call this command with more than one key to read from different streams at the same time. I could write, for instance: `STREAMS mystream otherstream 0 0`. Note how after the **STREAMS** option we need to provide the key names, and later the IDs. For this reason, the **STREAMS** option must always be the last option.
Any other options must come before the **STREAMS** option.

PR #2153, opened 2022-10-07
Correct `RENAMENX` behaviour

nadeesha:
`RENAMENX` returns an error when `key` exists.


enjoy-binbin (commented):
It returns an error when `key` does not exist.
```
127.0.0.1:6379> RENAMENX key newkey
(error) ERR no such key
```

PR #2157, opened 2022-10-11
Fix radius in GEOSEARCH example

thepill:
The example shows how to get all locations within a radius of 5 km, but the text says within 1 kilometer.

This PR will update the text to be in line with the example.


enjoy-binbin (approved):
thanks

itamarhaber (approved):

PR #2159, opened 2022-10-12
Fix typo in benchmarks

einverne:


enjoy-binbin (approved):
thanks

itamarhaber (changes requested):
Thanks for catching and fixing this - please also remove the typo from the "wordlist" file, it is there by mistake I hope (/cc @slorello89).

itamarhaber (approved):
Thanks @einverne!

PR #2162, opened 2022-10-16
Update transactions.md

naccome:
Typo fix
wont -> won't

Thanks!


enjoy-binbin (approved):
thanks, this looks more correct.
but let's wait for one more person's review

itamarhaber (approved):
Good catch @naccome and thanks for the fix.

PR #2214, opened 2022-11-17
Simplify SETEX command to use SET as the comparison instead of comman…

hbina:
…d transaction

Closes: https://github.com/redis/redis-doc/issues/2212

Signed-off-by: Hanif Bin Ariffin <hanif.ariffin.4326@gmail.com>

### DIFF
--- a/commands/setex.md
+++ b/commands/setex.md
@@ -1,16 +1,11 @@
 Set `key` to hold the string `value` and set `key` to timeout after a given
 number of seconds.
-This command is equivalent to executing the following commands:
+This command is equivalent to:
 
 ```
-SET mykey value
-EXPIRE mykey seconds
+SET key value EX seconds
 ```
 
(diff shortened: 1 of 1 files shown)


enjoy-binbin (commented):
> because this operation is very common when Redis is used as a cache.

Let's keep the original description and add an example of `set ex` after this?
I personally prefer this historical change description
but let wait for other opinions

zuiderkwast (approved):
I like this. SETEX is deprecated and we should direct people to what they should use instead and that it SET with EX.

I think we should also delete the example on SETEX and instead make sure the examples on the SET command includes EX and a TTL check. I see now that it already includes an example like `SET anotherkey "will expire in a minute" EX 60` so I think we just need to add `TTL anotherkey` to the same example.

Feel free to update the examples if you want. Otherwise, we can merge this as is.

itamarhaber (approved):

PR #2219, opened 2022-11-20
Fix XREADGROUP command error

zeekling:
Fix XREADGROUP command error


enjoy-binbin (approved):
Duplicate STREAMS, good catch, thanks for the fix

itamarhaber (approved):
Thanks @zeekling - so embarrassing 🤦 - merging

PR #2220, opened 2022-11-20
CONFIG SET without a restart

andrewvasilchuk:


enjoy-binbin (approved):
LGTM, good catch, thanks for the fix

nermiller (approved):
LGTM!

PR #2224, opened 2022-11-24
Fix the URL to streams' command list page.

kmiku7:


enjoy-binbin (approved):
good catch, thanks, verified

itamarhaber (approved):
Thanks @kmiku7 - merging

PR #2227, opened 2022-11-28
Add description of WITHSCORE argument to ZRANK and ZREVRANK.

CharlesChen888:
For PR redis/redis#11235, which adds the `WITHSCORE` option to command `ZRANK` and `ZREVRANK`.
Add description of option `WITHSCORE` argument to `zrank.md` and `zrevrank.md`.
Modify `commands.json`.

### DIFF
--- a/commands/zrank.md
+++ b/commands/zrank.md
@@ -3,14 +3,21 @@ ordered from low to high.
 The rank (or index) is 0-based, which means that the member with the lowest
 score has rank `0`.
 
+The optional `WITHSCORE` argument supplements the command's reply with the score of the element returned.
+
 Use `ZREVRANK` to get the rank of an element with the scores ordered from high
 to low.
 
 @return
 
-* If `member` exists in the sorted set, @integer-reply: the rank of `member`.
-* If `member` does not exist in the sorted set or `key` does not exist,
-  @bulk-string-reply: `nil`.
+* If `member` exists in the sorted set:
+  * using `WITHSCORE`, @array-reply: an array containing the rank and score of `member`.
+  * without using `WITHSCORE`, @integer-reply: the rank of `member`.
+* If `member` does not exist in the sorted set or `key` does not exist:
+  * using `WITHSCORE`, @array-reply: `nil`.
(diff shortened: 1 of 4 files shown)


oranagra (approved):
LGTM.

technically, json file changes aren't needed (the file is auto generated).
merging them may cause some issues (although more likely just be overwritten).

@itamarhaber correct me if i'm wrong, but merging it will immediately appear in redis.io, so not sure what's the proffered approach these days.

itamarhaber (approved):

itamarhaber:
Also LGTM.

> technically, json file changes aren't needed (the file is auto generated).

They are only needed if we want the new syntax to be shown. I assume that @CharlesChen888 generated commands.json using the standard script (https://github.com/redis/redis/blob/unstable/utils/generate-commands-json.py) and that this is just the diff.

> correct me if i'm wrong

You're not wrong or put differently, you're totally correct. I don't see a big issue with that, but this is apt to anger some users who won't notice the version note (and the breaking example until we release 7.2-rc1 as well).

CharlesChen888:
> technically, json file changes aren't needed (the file is auto generated).

Oh... I didn't realize this.

commands/zrank.md
```diff
@@ -3,14 +3,21 @@ ordered from low to high.
...
-* If `member` does not exist in the sorted set or `key` does not exist,
-  @bulk-string-reply: `nil`.
+* If `member` exists in the sorted set:
+  * using `WITHSCORE`, @array-reply: an array containing the rank and score of `member`.
+  * without using `WITHSCORE`, @integer-reply: the rank of `member`.
+* If `member` does not exist in the sorted set or `key` does not exist:
+  * using `WITHSCORE`, @array-reply: `nil`.
+  * without using `WITHSCORE`, @bulk-string-reply: `nil`.
+  
+Note that in RESP3 null and nullarray are the same, but in RESP2 they are not.
```
enjoy-binbin, line 20: https://github.com/redis/redis-doc/blob/master/wordlist
i guess we need to add nullarry to the wordlist

zuiderkwast (approved):

PR #2229, opened 2022-11-30
Update docs of consumer seen-time and active-time

guybe7:
See https://github.com/redis/redis/pull/11099


commands/xinfo-stream.md
```diff
@@ -107,12 +107,13 @@ OK
...
-              5) "pel-count"
-              6) (integer) 1
-              7) "pending"
-              8) 1) 1) "1638125133432-0"
-                    2) (integer) 1638125153423
-                    3) (integer) 1
->
+              4) (integer) 1638125133422
+              5) "active-time"
+              6) (integer) 1638125133432
```
oranagra, line 112: so many of the consumer related fields are undocumented in this sub-command?
where do users get the info about them?

commands/xinfo-consumers.md
```diff
@@ -4,7 +4,8 @@ The following information is provided for each consumer in the group:
 
 * **name**: the consumer's name
 * **pending**: the number of pending messages for the client, which are messages that were delivered but are yet to be acknowledged
-* **idle**: the number of milliseconds that have passed since the consumer last interacted with the server
+* **idle**: the number of milliseconds that have passed since the consumer last interacted with the server (Example: `XREADGROUP`)
+* **inactive**: the number of milliseconds that have passed since the consumer last successfully interacted with the server (Example: `XREADGROUP` that actually read some entries into the PEL)
```
oranagra, line 8: i'm not sure if `PEL` is a term we can use here is it extensively used in the rest of the streams documentation?

itamarhaber: It is, unfortunately/thanks god it is used throughout the docs.

@guy where does `XACK` and `X[AUTO]CLAIM` fit? Please add to the "examples"

commands/xinfo-stream.md
itamarhaber: I believe these are currently simply not documented.

itamarhaber (approved):

commands/xinfo-stream.md
guybe7: XINFO STREAM FULL is not documented at all

guybe7: https://github.com/redis/redis-doc/issues/2230

commands/xinfo-consumers.md
guybe7: added

for the record, XACK is out of the game because , oddly, it's not consumer-specific

oranagra:
not a comment about being ready or not, but maybe we should hold it back until closer to 7.2 release?
just to avoid confusing people?

p.s. i don't see any mention of a history (when the new field was added, and another was effectively changed) in either the command json files in the redis repo or the docs here... i suppose we should add it (maybe in both?)

guybe7:
@oranagra you're right - https://github.com/redis/redis/issues/11754

PR #2233, opened 2022-12-02
docs: correct minimum possible id

andrewvasilchuk:
Everywhere across the docs, as well in the Redis University course RU202, it's stated that the minimal possible id is `0-1`, but not `0-0`

### DIFF
--- a/commands/xrange.md
+++ b/commands/xrange.md
@@ -41,9 +41,8 @@ will just return every entry in the stream:
 ... other entries here ...
 ```
 
-The `-` ID is effectively just exactly as specifying `0-1`, while
-`+` is equivalent to `18446744073709551615-18446744073709551615`, however
-they are nicer to type.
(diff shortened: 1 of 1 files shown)


itamarhaber:
Hello @andrewvasilchuk 

All over the docs and in RU202 it is said (sometimes even by me ;)) that while '0-1' is the minimal valid entry ID, '0-0' is lower than it (albeit an invalid ID for an entry). Does that make sense?

andrewvasilchuk:
Hello @itamarhaber 

It makes sense. What confuses me is this statement on this page.

> The - and + special IDs mean respectively the minimum ID possible and the maximum ID possible inside a stream

So `-` should be equal to `0-1`, since `0-0` is not possible ID.

andrewvasilchuk:
@itamarhaber done

PR #2237, opened 2022-12-07
fix: setbit should return 0 on an empty instance

liuwenzhuang:
setbit return the original bit value stored at offset, so `setbit key 10 1` should return 0.

### DIFF
--- a/docs/manual/pubsub.md
+++ b/docs/manual/pubsub.md
@@ -77,12 +77,18 @@ environment (test, staging, production...).
 
 ```
 SUBSCRIBE first second
-1) "subscribe"
-2) "first"
-3) (integer) 1
-1) "subscribe"
-2) "second"
-3) (integer) 2
+*3
+$9
+subscribe
+$5
+first
+:1
+*3
+$9
+subscribe
+$6
+second
+:2
 ```
 
 At this point, from another client we issue a `PUBLISH` operation
@@ -95,9 +101,13 @@ against the channel named `second`:
 This is what the first client receives:
 
 ```
-1) "message"
-2) "second"
-3) "Hello"
+*3
+$7
+message
+$6
+second
+$5
+Hello
 ```
 
 Now the client unsubscribes itself from all the channels using the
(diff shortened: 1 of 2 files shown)


docs/manual/pubsub.md
```diff
@@ -77,18 +77,12 @@ environment (test, staging, production...).
...
-subscribe
-$6
-second
-:2
+1) "subscribe"
+2) "first"
+3) (integer) 1
+1) "subscribe"
+2) "second"
+3) (integer) 2
```
itamarhaber, line 85: Please revert - this is the wire protocol format, not redis-cli.

docs/manual/pubsub.md
```diff
@@ -101,32 +95,22 @@ against the channel named `second`:
...
-*3
-$7
-message
-$6
-second
-$5
-Hello
+1) "message"
+2) "second"
+3) "Hello"
```
itamarhaber, line 100: Same comment

docs/manual/pubsub.md
```diff
@@ -101,32 +95,22 @@ against the channel named `second`:
...
-unsubscribe
-$5
-first
-:0
+1) "unsubscribe"
+2) "second"
+3) (integer) 1
+1) "unsubscribe"
+2) "first"
+3) (integer) 0
```
itamarhaber, line 113: Same comment

itamarhaber (changes requested):
Hello @liuwenzhuang - thanks for catching that erroneous example.

Please revert all the changes you've made to the examples of Pub/Sub's wire protocol.

zuiderkwast:
> setbit return the original bit value stored at offset, so `setbit key 10 1` should return 0.

We don't know what was stored in the key before the example command was called. It may be 1.

For the other changes, they are just wrong (as @itamarhaber pointed out) so I think this PR can be closed.

itamarhaber:
I still think the example is misleading, in the sense that on an empty instance, it should return 0.

zuiderkwast:
Fair enoguh.

liuwenzhuang:
> Hello @liuwenzhuang - thanks for catching that erroneous example.
> 
> Please revert all the changes you've made to the examples of Pub/Sub's wire protocol.

Done.

itamarhaber (approved):

PR #2259, opened 2022-12-15
Fix SPUBLISH reply description

leibale:
@itamarhaber

### DIFF
--- a/commands/spublish.md
+++ b/commands/spublish.md
@@ -8,7 +8,8 @@ For more information about sharded pubsub, see [Sharded Pubsub](/topics/pubsub#s
 
 @return
 
-@integer-reply: the number of clients that received the message. Note that in a Redis Cluster, only clients that are connected to the same node as the publishing client are included in the count.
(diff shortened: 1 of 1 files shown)


commands/spublish.md
```diff
@@ -8,7 +8,7 @@ For more information about sharded pubsub, see [Sharded Pubsub](/topics/pubsub#s
 
 @return
 
-@integer-reply: the number of clients that received the message.
+@integer-reply: the number of clients that received the message. Note that in a Redis Cluster, only clients that are connected to the same node as the publishing client are included in the count.
```
itamarhaber, line 11:
```suggestion
@integer-reply: the number of clients that received the message.
Note that in a Redis Cluster, only clients that are connected to the same node as the publishing client are included in the count.
```

leibale: It's just a copy from PUBLISH, should I change it there too?

itamarhaber (approved):

PR #2271, opened 2022-12-29
Added rustis client library

mcatanzariti:


zuiderkwast (approved):
LGTM.

Just curios: What's wrong with the other Rust clients? Are they not asynchronous?

itamarhaber (approved):

mcatanzariti:
> LGTM.
> 
> Just curios: What's wrong with the other Rust clients? Are they not asynchronous?

Yes they are !

To be totally honest, my first requirement was to write a notification service based on pub/sub.

I tried a few libraries:
- `redis-rs` is the most popular one but pub/sub is still synchronous
- `redis-async-rs` is two low-level, buggy and does not seem active anymore, even if their pub/sub implementation is async
- `fred` is asynchronous & active but their pub/sub implementation is not flexible enough for my requirements.
Consequently I decided to write my own implementation of Redis for Rust with the first objective to learn Rust, that I discovered a few month ago, and also to be full async, fully compliant with Redis 7.x & Redis-Stack and to offer a very flexible Pub/Sub implementation.

Finally I reached a point where I thought it could be interesting to share it in the official documentation.

PR #2281, opened 2023-01-07
Fix small typo in tutorial.md

alexsyeo:
Should either read "elements in a sorted set" or "elements in sorted sets".


itamarhaber (approved):
Thank you @alexsyeo - merging

PR #2289, opened 2023-01-18
Clarified that LINSERT will return 0 if the key does not exist

### LINKED ISSUE #2288: LINSERT returns undocumented 0 when the key does not exist
`LINSERT` return documentation says:

```
Return
Integer reply: the length of the list after the insert operation, or -1 when the value pivot was not found.
```

It turns out this command will return 0 (the length of the list) when the key does not exist - and insert operation was not done.

The return value should be clarified, so that application writers know to look for <= 0 for pivot not found, instead of just -1.

Other commands may have similar issues, as key existence seems to take precedence over the operation's specific meaning of return value.

enjoy-binbin:
Fixes #2288

### DIFF
--- a/commands/linsert.md
+++ b/commands/linsert.md
@@ -8,8 +8,7 @@ An error is returned when `key` exists but does not hold a list value.
 
 @return
 
-@integer-reply: the length of the list after the insert operation, or `-1` when
-the value `pivot` was not found, or `0` if the `key` does not exist.
+@integer-reply: the list length after a successful insert operation, `0` if the `key` doesn't exist, and `-1` when the `pivot` wasn't found.
 
 @examples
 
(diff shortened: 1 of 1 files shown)


commands/linsert.md
```diff
@@ -9,7 +9,7 @@ An error is returned when `key` exists but does not hold a list value.
 @return
 
 @integer-reply: the length of the list after the insert operation, or `-1` when
-the value `pivot` was not found.
+the value `pivot` was not found, or `0` if the `key` does not exist.
```
itamarhaber, line 12:
```suggestion
@integer-reply: the list length after a successful insert operation, `0` if the `key` doesn't exist, and `-1` when
the `pivot` wasn't found.
```

enjoy-binbin: thanks, it does look better

PR #2299, opened 2023-01-25
Add phpCacheAdmin

RobiNN1:
A universal web dashboard with support for various cache systems with nicer ui. Also it supports both phpredis and predis clients.


itamarhaber (approved):
Good job @RobiNN1 - merging.

PR #2302, opened 2023-01-28
Fix the link for the word sets

impocode:


enjoy-binbin (approved):
good catch, thanks for the fix

itamarhaber (approved):

PR #2305, opened 2023-01-28
Update MEMORY USAGE doc to be more specific about SAMPLES

thundergolfer:
Adds an extra statement to make clearer that N samples are averaged and then the result is multiplied by nested key's cardinality. The docstring for the function has this extra specificity, so this change just carries it over into the docs: https://github1s.com/redis/redis/blob/HEAD/src/object.c#L992.


itamarhaber (approved):
Thanks @thundergolfer - merging.

PR #2307, opened 2023-01-30
Remove extra spaces

impocode:


itamarhaber (approved):
Hello @impocode - thanks for removing the spaces

enjoy-binbin (approved):

zuiderkwast (commented):
This change doesn't affect the rendering of the page at all.

We usually avoid this kind of changes, since it affects the git history in an unnecessary way. It makes it harder to see e.g. who wrote these lines and when, using git blame.

enjoy-binbin:
yean, indeed. on the other side, my opinion on the documentation is, as standardized (md format) and consistent as possible? actually i am ok with both.

Not sure how git recognizes spaces, at least in my IDE (eg. clion), it won't break the git blame much. like i think it is using git blame with -w option by default.  see https://git-scm.com/docs/git-blame#Documentation/git-blame.txt--w

zuiderkwast:
I'm OK with both too. Since Itamar has already approved it, let's merge it.

You're right it's possible to use -w with git blame and git diff, even in github's diff view.

PR #2310, opened 2023-01-31
Update distributed-locks.md to add Rust distributed lock library

hexcowboy:


hexcowboy:
> Hello @hexcowboy - it appears your repo is a fork, why not reference the upstream?

Upstream does not have async or extend functionality, but I could also add upstream to the list as a separate project if you think it's a good idea.

zuiderkwast:
@hexcowboy I see that your fork is actually a fork of a fork. :-)

Since you have renamed it and created a separate crate, I think it's fine to merge this.

IMO we don't need to add links to the other forks now, especially if they are inactive for years. It is possible to find them from your repo.

PR #2311, opened 2023-02-01
Update benchmarks.md

jerviscui:
Fix image link address and makrdown format


enjoy-binbin (approved):
this look right, @banker can you take a look?

itamarhaber:
Thanks @jerviscui - I refactored it so it would work. Merging.

PR #2314, opened 2023-02-02
Update client-side-caching.md

Virusuki:
I think it's easier to learn client-side caching if this document mentions the cache aside pattern strategy.

### DIFF
--- a/docs/manual/client-side-caching.md
+++ b/docs/manual/client-side-caching.md
@@ -8,7 +8,8 @@ aliases:
     - /topics/client-side-caching
 ---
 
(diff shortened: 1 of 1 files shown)


itamarhaber:
Hello @Virusuki 

Thanks for the suggestion, but I think it makes it less clear. Cache-aside is an access pattern, whereas client-side caching is the notion of caching in the client's (application) context.

Virusuki:
As you mentioned, the expression "client-side caching in the client's (application) context" seems appropriate.
I will to participate persistingly in the contribution of redis. :)

PR #2341, opened 2023-03-02
Update install-redis-on-linux.md Added instalisation for Arch Linux

norim-code:


itamarhaber:
Hello @norim-code, and thank you for taking the time to propose this merge.

While the instructions appear to be sane (I haven't personally tested them), I'm afraid we can't accept the change.
The reason is that the Redis project doesn't provide maintain the Arch Linux package, so it isn't in scope.
I apologize for the effort you've put into this, but I hope you understand.

PR #2344, opened 2023-03-06
Add new request policy "any_shard"

uglide:
Modules like RediSearch and RedisTimeSeries need additional request policy to inform clients that commands can be sent to any master shard, not just to the first one. It's essential for CPU-bound commands like FT.SEARCH.

### DIFF
--- a/docs/reference/command-tips.md
+++ b/docs/reference/command-tips.md
@@ -49,7 +49,10 @@ In cases where the client should adopt a behavior different than the default, th
 * **all_shards:** the client should execute the command on all master shards (e.g., the `DBSIZE` command).
   This tip is in-use by commands that don't accept key name arguments.
   The command operates atomically per shard.
-- **multi_shard:** the client should execute the command on several shards.
+- **any_shard:** the client should execute the command on any master shard.
+  This tip is can be used by commands that should be executed on different master shards to spread the load. 
+  Examples of such commands include `FT.SEARCH`, `FT.AGGREGATE` and `TS.MGET`.
+* **multi_shard:** the client should execute the command on several shards.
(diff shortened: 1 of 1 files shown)


zuiderkwast:
Related: PUBLISH and SUBSCRIBE should also be any_shard or maybe any_node.

uglide:
@itamarhaber In the original text, we use "shard" as a synonym for "master":
> **all_shards:** the client should execute the command on all **master** shards

That's why I used `any_shard` to make it consistent.

zuiderkwast:
> @itamarhaber In the original text, we use "shard" as a synonym for "master":
> 
> > **all_shards:** the client should execute the command on all **master** shards
> 
> That's why I used `any_shard` to make it consistent.

Well, that's confusing.

For sharded pubsub, we use "shard" for all the nodes in a shard:

> The cluster makes sure the published shard messages are forwarded to all nodes in the shard, so clients can subscribe to a shard channel by connecting to either the master responsible for the slot, or to any of its replicas.

oranagra:
any_shard is the default (for commands that don't take key arguments)

here's what the docs says (a few lines above the ones modified):

> The default behavior a client should implement for commands without the _request_policy_ tip is as follows:
> 1. The command doesn't accept key name arguments: the client can execute the command on an arbitrary shard.

am i missing something? (i only skimmed thought the conversation)

uglide:
@oranagra clients now interpret it as "execute all the time on the same/first shard." In most cases, it should be some round-robin to distribute the load. Do you think there is a better way to address it?

oranagra:
@uglide do the clients do that because that was their behavior before hints were available? or because they misread the docs?
do you see any reason not to fix the clients?

PR #2350, opened 2023-03-09
Use correct link to the RESP3 specification

bjosv:
Removes the last reference to the old RESP3 specification site.


itamarhaber (approved):
Thanks @bjosv 💯

PR #2352, opened 2023-03-13
More inclusivity in prior knowledge assertions -- using passive tone is less intimidating!

agent-ai-consulting:
The original author writes usefully yet elitistly about decision to use PHP: "I used PHP for the example since it can be read by everybody. The same (or better) results can be obtained using Ruby, Python, Erlang, and so on." I am suggesting a rephrasing of the first sentence so that everyone feels more welcome to this tutorial, even those of us who may not be reading PHP well. --> New sentence is: "I used PHP for the example because of its universal readability."


itamarhaber:
Hello @agent-ai-consulting - are you an AI?

itamarhaber (approved):

PR #2363, opened 2023-03-22
update module api reference for 7.2-rc1

oranagra:


oranagra:
@zuiderkwast the new module APIs show unreleased because the script doesn't recognize the `7.2-rc1` tag.
maybe we wanna improve..

PR #2381, opened 2023-04-03
Update the description of the `count` parameter in the slowlog-get co…

judeng:
fix the default return messages and add the sepcial count value '-1'


itamarhaber (approved):
Thank you @judeng - merging.

PR #2383, opened 2023-04-13
removal of $ from $ brew -version command

Dhiraj5789:
On the Install Redis on macOs page the command to check brew version contains $ in its line. Since rest of the commands given in the document start directly from brew and not $ brew as $ is already present in the terminal it will be easier to just keep brew --version so that the user can directly copy that command in their terminal, else they'll have to remove $ from the command line manually. This seems much easier from a user's perspective to copy directly and paste the command without failure.


itamarhaber (approved):
Thanks @Dhiraj5789 - merging

PR #2395, opened 2023-04-22
Update zrange.md

Andrei-Banu:
The last 2 examples have `(empty array)` as response in the cli. Updated the cli command to make sure there is data in the zset before querying it


enjoy-binbin (approved):
i checked, this look right, thanks.

itamarhaber (approved):
Thanks @Andrei-Banu - much appreciated.

The root cause for this issue is that we've replaced the interactive examples' backend. The previous implementation treated all examples in a single page as a single session, whereas the current one treats every example - regardless of the page - as a session. Until we fix the backend (created an issue to track at https://github.com/redis/redis-clinterwebz/issues/21), this is a good workaround so merging.

PR #2400, opened 2023-04-24
Added text to indicate that we need to run make command before running any scripts

Himachallad:
Handled issue: https://github.com/redis/redis-doc/issues/2399


itamarhaber:
https://github.com/redis/redis-doc/issues/2399#issuecomment-1521831258

PR #2403, opened 2023-05-04
fix broken hash in link to "read-only scripts" heading

fabiosantoscode:
I noticed the link to [this heading](https://redis.io/docs/manual/programmability/#read-only-scripts) was broken, so I fixed it.


itamarhaber (approved):
Thanks @fabiosantoscode - merging!

PR #2425, opened 2023-05-31
Update text leaderboard

trungdlp-wolffun:
https://github.com/redis/redis-doc/blob/master/wordlist#L216


itamarhaber (approved):
Thanks @trungdlp-wolffun - merging.

PR #2426, opened 2023-06-05
Add cURL & GPG to minimal linux install requirements

camdenmoors:


itamarhaber (approved):
Thanks @camdenmoors - merging

PR #2428, opened 2023-06-06
Clarify WAITAOF documentation

slavak:
Description was worded in a way that might have made it unclear the command would only wait for write commands _by the same client_ that executed it.


itamarhaber (approved):
TY @slavak - merging

Issue #2427, opened 2023-06-05
Scaling with Redis Cluster


jiangyunpeng:
hello 

I wanted to talk about the 《Remove a node 》sections [in this chapter](https://redis.io/docs/management/scaling/#remove-a-node) , recently i prepared to remove some failed node in cluster ,when i used redis-cli  --cluster del-node 10.20.32.2:6383 44a0355dff74211554c4df6a7daaca5d2cde1227 i got error: " Could not connect to Redis at failed-node-ip:failed-node-port: Connection refused ". 

The document did not note how to delete failed node . actually it should use redis-cli  --cluster call 10.20.32.3:6380 cluster forget 808bd1a6208b833f005305113fe5d9d5102ede96   i thought cluster forget command must to be applied on every node ( if i was wrong welcome to correct me)

i've spend a few time to find the correct way to remove failed node , so i hope redis-doc will add it , thank you

jiangyunpeng:
hello @itamarhaber  there is my PR https://github.com/redis/redis-doc/pull/2434   thanks

zuiderkwast:
Closed by #2434

PR #2434, opened 2023-06-08
Explain how to remove failed cluster node (#2427)

jiangyunpeng:
`redis-cli --cluster del-node` doesn't work since it tries to connect to all nodes. `--cluster call CLUSTER FORGET` works though.


zuiderkwast (commented):
Thanks. This helps but it looks like a workaround. It would be better if `del-node` can delete a failed node.

Why does `del-node` need to connect to all nodes in the cluster?

zuiderkwast (approved):

PR #2476, opened 2023-07-07
Update strings.md

sav-norem:
update strings datatype page to include hugo short codes for tabbed examples, removed "Let's play a bit with the string type, using redis-cli (all the examples will be performed via redis-cli in this tutorial)." as these examples will be in all the client languages

### DIFF
--- a/docs/data-types/strings.md
+++ b/docs/data-types/strings.md
@@ -15,13 +15,12 @@ Since Redis keys are strings, when we use the string type as a value too,
 we are mapping a string to another string. The string data type is useful
 for a number of use cases, like caching HTML fragments or pages.
 
-Let's play a bit with the string type, using `redis-cli` (all the examples
-will be performed via `redis-cli` in this tutorial).
-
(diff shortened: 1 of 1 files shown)


sav-norem:
@elena-kolevska here's the string datatype page update - assuming David Dougherty should be involved but don't see his github as an account I can tag. Also a lot of the spell checks fail as the hugo short code names are things like setnx_xx

elena-kolevska:
Thanks Savannah. You can add the IDs to the [wordlist](https://github.com/redis/redis-doc/blob/master/wordlist) file in order to force the checker to ignore them.

This is David"s hangle: @dwdougherty .

dwdougherty (commented):
These look okay to me.

dmaier-redislabs (approved):

dmaier-redislabs:
This led to a failed build. We need to investigate it. @dwdougherty will fix it and send a new pull request for the fix.

PR #2482, opened 2023-07-14
Update doc of command CLUSTER NODES.

CharlesChen888:
In redis/redis#12411 the format of the return of `CLUSTER NODES` is modified. And we need to correct the doc here.

I deleted the description about auxiliary fields in this page, but I don't find it anywhere else. I think we still need to doc it somewhere, maybe in a page specifically for the nodes.conf file.

### DIFF
--- a/commands/cluster-nodes.md
+++ b/commands/cluster-nodes.md
@@ -17,9 +17,7 @@ It is also used by `redis-cli` in order to manage a cluster.
 ## Serialization format
 
 The output of the command is just a space-separated CSV string, where
-each line represents a node in the cluster. The auxiliary fields 
-(including shard ID, human node name and extra port number) stored in
-nodes.conf file are currently hidden from client commands. The following
+each line represents a node in the cluster. The following
 is an example of output on Redis 7.2.0.
 
 ```
--- a/commands/cluster-nodes.md
+++ b/commands/cluster-nodes.md
@@ -42,13 +42,13 @@ The meaning of each field is the following:
 1. `id`: The node ID, a 40-character globally unique string generated when a node is created and never changed again (unless `CLUSTER RESET HARD` is used).
 2. `ip:port@cport`: The node address that clients should contact to run queries.
 3. `hostname`: A human readable string that can be configured via the `cluster-annouce-hostname` setting. The max length of the string is 256 characters, excluding the null terminator. The name can contain ASCII alphanumeric characters, '-', and '.' only.
-4. `flags`: A list of comma separated flags: `myself`, `master`, `slave`, `fail?`, `fail`, `handshake`, `noaddr`, `nofailover`, `noflags`. Flags are explained below.
-5. `master`: If the node is a replica, and the primary is known, the primary node ID, otherwise the "-" character.
-6. `ping-sent`: Unix time at which the currently active ping was sent, or zero if there are no pending pings, in milliseconds.
-7. `pong-recv`: Unix time the last pong was received, in milliseconds.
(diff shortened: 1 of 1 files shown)


commands/cluster-nodes.md
```diff
@@ -17,41 +17,38 @@ It is also used by `redis-cli` in order to manage a cluster.
...
+6. `ping-sent`: Unix time at which the currently active ping was sent, or zero if there are no pending pings, in milliseconds.
+7. `pong-recv`: Unix time the last pong was received, in milliseconds.
+8. `config-epoch`: The configuration epoch (or version) of the current node (or of the current primary if the node is a replica). Each time there is a failover, a new, unique, monotonically increasing configuration epoch is created. If multiple nodes claim to serve the same hash slots, the one with the higher configuration epoch wins.
+9. `link-state`: The state of the link used for the node-to-node cluster bus. Use this link to communicate with the node. Can be `connected` or `disconnected`.
+10. `slot`: A hash slot number or range. Starting from argument number 9, but there may be up to 16384 entries in total (limit never reached). This is the list of hash slots served by this node. If the entry is just a number, it is parsed as such.  If it is a range, it is in the form `start-end`, and means that the node is responsible for all the hash slots from `start` to `end` including the start and end values.s
```
madolson, line 51:
```suggestion
5. `flags`: A list of comma separated flags: `myself`, `master`, `slave`, `fail?`, `fail`, `handshake`, `noaddr`, `nofailover`, `noflags`. Flags are explained below.
6. `master`: If the node is a replica, and the primary is known, the primary node ID, otherwise the "-" character.
7. `ping-sent`: Unix time at which the currently active ping was sent, or zero if there are no pending pings, in milliseconds.
8. `pong-recv`: Unix time the last pong was received, in milliseconds.
9. `config-epoch`: The configuration epoch (or version) of the current node (or of the current primary if the node is a replica). Each time there is a failover, a new, unique, monotonically increasing configuration epoch is created. If multiple nodes claim to serve the same hash slots, the one with the higher configuration epoch wins.
10. `link-state`: The state of the link used for the node-to-node cluster bus. Use this link to communicate with the node. Can be `connected` or `disconnected`.
11. `slot`: A hash slot number or range. Starting from argument number 9, but there may be up to 16384 entries in total (limit never reached). This is the list of hash slots served by this node. If the entry is just a number, it is parsed as such.  If it is a range, it is in the form `start-end`, and means that the node is responsible for all the hash slots from `start` to `end` including the start and end values.
```

You accidentally added an s at the end of the line, so I think it makes more sense to leave the numbers in place since by the markdown standard they don't matter.

commands/cluster-nodes.md
```diff
@@ -17,31 +17,31 @@ It is also used by `redis-cli` in order to manage a cluster.
 ## Serialization format
 
 The output of the command is just a space-separated CSV string, where
-each line represents a node in the cluster. Starting from 7.2.0, the
-output of the command always contains a new auxiliary field called
-shard-id. The following is an example of output on Redis 7.2.0.
+each line represents a node in the cluster. The auxiliary fields 
+(including shard ID, human node name and extra port number) stored in
+nodes.conf file are currently hidden from client commands. The following
```
madolson, line 22:
```suggestion
each line represents a node in the cluster. The following
```

Our historical guidance has been the command should show the latest state, since aux fields aren't the latest we can leave them omitted.

madolson (approved):

madolson:
> I deleted the description about auxiliary fields in this page, but I don't find it anywhere else.

For the time being I don't think we need to document it.

PR #2485, opened 2023-07-15
Alternative method to install redis on windows via Scoop

RayyanNafees:
This method involves installing additional heavy tools like Docker & WSL2

Not every Windows PC can run WSL smoothly

& Docker is a pretty heavy engine and not everyone (especially the non-devops guys) want to install it

Best thing is to use is the Scoop package manager which keeps things tidy in their own SHIM environments


itamarhaber:
Hi @RayyanNafees, and thanks for the suggestion.

Scoop appears to be a neat package manager. However, because the Redis package that it installs isn't built by us (the OSS Redis project), we can't recommend/endorse it (I'm pretty sure that it isn't working properly btw).

Closing.

RayyanNafees:
it works seamless since 3 yrs

PR #2486, opened 2023-07-16
Fix typo (`@readonly` -> `@read`)

sh-cho:
It's one of ACL categories and I guess it's a typo right?


itamarhaber:
Thanks @sh-cho - a typo indeed.

PR #2493, opened 2023-07-20
Fix multi-shard explanation.

nihohit:


docs/reference/command-tips.md
```diff
@@ -50,7 +50,7 @@ In cases where the client should adopt a behavior different than the default, th
   This tip is in-use by commands that don't accept key name arguments.
   The command operates atomically per shard.
 - **multi_shard:** the client should execute the command on several shards.
-  The shards that execute the command are determined by the hash slots of its input key name arguments.
+  The client should split the inputs according to the hash slots of its input key name arguments. For example, the command `MSET {foo} baz {foo}1 baz1 bar vaz` should be split to `MSET {foo} baz {foo}1 baz1` and `MSET bar vaz`. If the keys are hashed to more than a single slot, the command must be split even if all the slots are managed by the same shard.
```
oranagra, line 53: maybe an example with DEL will be easier to follow?

oranagra (approved):
@itamarhaber PATL

docs/reference/command-tips.md
nihohit: I intentionally chose MSET, because it's slightly more complex than MGET / DEL. If you think that the phrasing and principle are clear enough that a less complex example would work, I'll switch to DEL.

oranagra: i think the principle is clear enough with DEL and MGET and it'll be easier to read.
it's right that the implementation in case of MSET is more complex, but that's something to worry about in implementation stage.

nihohit: done

oranagra (approved):

docs/reference/command-tips.md
```diff
@@ -50,7 +50,7 @@ In cases where the client should adopt a behavior different than the default, th
   This tip is in-use by commands that don't accept key name arguments.
   The command operates atomically per shard.
 - **multi_shard:** the client should execute the command on several shards.
-  The shards that execute the command are determined by the hash slots of its input key name arguments.
+  The client should split the inputs according to the hash slots of its input key name arguments. For example, the command `DEL {foo} {foo}1 bar` should be split to `DEL {foo} {foo}1` and `DEL bar`. If the keys are hashed to more than a single slot, the command must be split even if all the slots are managed by the same shard.
```
zuiderkwast, line 53: One sentence per line as per the guidelines in the README.

```suggestion
  The client should split the inputs according to the hash slots of its input key name arguments.
  For example, the command `DEL {foo} {foo}1 bar` should be split to `DEL {foo} {foo}1` and `DEL bar`.
  If the keys are hashed to more than a single slot, the command must be split even if all the slots are managed by the same shard.
```

zuiderkwast:
I merged this now. Thanks!

I have two other questions regarding the multi-shard request policy:

1. Why is there no mention that this behaviour of a client breaks the atomicity of the commands? I wasn't aware that we actually recommend this behaviour.
2. What shall the client return to the user if the command fails on one node and succeeds on the other nodes? (There is no obvious way to report a partial failure to the user.)

PR #2497, opened 2023-07-26
Fix minor errors and simplify explaination for Gopher support

hboutou:


zuiderkwast (commented):
I believe Gopher support has been removed from Redis a few years ago.

PR #2501, opened 2023-07-29
Update install-redis-on-linux adding arch linux

carvilsi:
Adding how to install in Arch Linux.


carvilsi:
Hi @itamarhaber 

Thanks for your reply.
The reason makes all the sense :smile:

PR #2503, opened 2023-07-31
Update strings.md , wrong command description

SHresTho12:
The string value is parsed into integer and incremented by one by the `INCR` command not the `INCRBY`.  `INCRBY` command increments the value with a given value not by 1 automatically. It increments the value by one if only it is explicitly told : `incrby total_crashes 1`


enjoy-binbin (approved):
LGTM, thanks for the fix

itamarhaber (approved):

itamarhaber:
Thanks @SHresTho12 - merging.

PR #2507, opened 2023-08-06
Remove unnecessary code in streams consumer implementation

hboutou:
When the loop is still going through backlog, whether $lastid is '0-0' or the id of the last processed message, both have same outcome in this context.  Also, once the consumer is caught up with backlog, the id used becomes `'>'`. So it looks like saving the id of last processed message is unnecessary.

### DIFF
--- a/docs/data-types/streams.md
+++ b/docs/data-types/streams.md
@@ -458,8 +458,6 @@ while true
 
         # Acknowledge the message as processed
         r.xack(:my_stream_key,GroupName,id)
-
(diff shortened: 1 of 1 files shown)


zuiderkwast (changes requested):
Not saving the last id makes the variable `$lastid` confusing. I think we can save it for clarity.

(If we don't want to save it, we don't need the variable at all and we need to update the comments to explain why we use 0-0 every time.)

PR #2516, opened 2023-08-14
updating streams data type page with bike story for CLI commands and …

sav-norem:
…hugo short codes for tabbed examples


sav-norem:
@dwdougherty I don't know how to fix this last spellcheck fail since it's like a repeated word so. I tried to get this wordlist right but if you could check this out, that'd be great!

sav-norem:
@dwdougherty okay streams are finally ready

dwdougherty:
Okay. I'll have a look soon.

docs/data-types/streams.md
```diff
@@ -114,26 +118,26 @@ The format of such IDs may look strange at first, and the gentle reader may wond
 
 If for some reason the user needs incremental IDs that are not related to time but are actually associated to another external system ID, as previously mentioned, the `XADD` command can take an explicit ID instead of the `*` wildcard ID that triggers auto-generation, like in the following examples:
 
-```
-> XADD somestream 0-1 field value
+{{< clients-example stream_toturial xadd_id >}}
```
dwdougherty, line 121: I imagine this mispelling will likely trip you up at some point with the other client examples. There are several instances of this that need to be corrected.

```suggestion
{{< clients-example stream_tutorial xadd_id >}}
```

docs/data-types/streams.md
```diff
@@ -149,65 +153,132 @@ Redis Streams support all three of the query modes described above via different
...
+      2) "Castilla"
+      3) "speed"
+      4) "30.2"
+      5) "position"
+      6) "1"
+      7) "location_id"
+      8) "1"
+{{< /clients-example >}}
+
+I have only a single entry in this range, however in real data sets, I could query for ranges of hours, or there could be many items in just two milliseconds, and the result returned could be huge. For this reason, `XRANGE` supports an optional **COUNT** option at the end. By specifying a count, I can just get the first *N* items. If I want more, I can get the last ID returned, increment the sequence part by one, and query again. Let's see this in the following example. Let's assume that the stream `race:france` was populated with 4 items. To start my iteration, getting 2 items per command, I start with the full range, but with a count of 2.
```
dwdougherty, line 211:
```suggestion
I have only a single entry in this range. However, in real data sets, I could query for ranges of hours, or there could be many items in just two milliseconds, and the result returned could be huge. For this reason, `XRANGE` supports an optional **COUNT** option at the end. By specifying a count, I can just get the first *N* items. If I want more, I can get the last ID returned, increment the sequence part by one, and query again. Let's see this in the following example. Let's assume that the stream `race:france` was populated with 4 items. To start my iteration, getting 2 items per command, I start with the full range, but with a count of 2.
```

docs/data-types/streams.md
```diff
@@ -149,65 +153,132 @@ Redis Streams support all three of the query modes described above via different
...
+      2) "Norem"
+      3) "speed"
+      4) "28.8"
+      5) "position"
+      6) "3"
+      7) "location_id"
+      8) "1"
+{{< /clients-example >}}
+
+In order to continue the iteration with the next two items, I have to pick the last ID returned, that is `1692632094485-0` and add the prefix `(` to it. The resulting exclusive range interval, that is `(1692632094485-0` in this case, can now be used as the new *start* argument for the next `XRANGE` call:
```
dwdougherty, line 235:
```suggestion
To continue the iteration with the next two items, I have to pick the last ID returned, that is `1692632094485-0`, and add the prefix `(` to it. The resulting exclusive range interval, that is `(1692632094485-0` in this case, can now be used as the new *start* argument for the next `XRANGE` call:
```

docs/data-types/streams.md
```diff
@@ -149,65 +153,132 @@ Redis Streams support all three of the query modes described above via different
...
+      2) "Castilla"
+      3) "speed"
+      4) "29.9"
+      5) "position"
+      6) "1"
+      7) "location_id"
+      8) "2"
+{{< /clients-example >}}
+
+Now that we've gotten 4 items out of a stream that only had 4 things in it, if we try to get more items, we'll get an empty array:
```
dwdougherty, line 259:
```suggestion
Now that we've retrieved 4 items out of a stream that only had 4 entries in it, if we try to retrieve more items, we'll get an empty array:
```

docs/data-types/streams.md
```diff
@@ -306,52 +389,46 @@ Now it's time to zoom in to see the fundamental consumer group commands. They ar
...
-
-```
-> XREADGROUP GROUP mygroup Alice COUNT 1 STREAMS mystream >
-1) 1) "mystream"
-   2) 1) 1) 1526569495631-0
-         2) 1) "message"
-            2) "apple"
-```
+We'll add riders to the race:italy stream and try reading something using the consumer group:
+Note: *here rider is the field name, and the name is the associated value, remember that stream items are small dictionaries.*
```
dwdougherty, line 413:
```suggestion
Note: *here rider is the field name, and the name is the associated value. Remember that stream items are small dictionaries.*
```

docs/data-types/streams.md
```diff
@@ -715,45 +779,47 @@ So basically Kafka partitions are more similar to using N different Redis keys,
...
 
 ```
-XADD mystream MAXLEN ~ 1000 * ... entry fields here ...
+XADD race:italy MAXLEN ~ 1000 * ... entry fields here ...
 ```
 
-The `~` argument between the **MAXLEN** option and the actual count means, I don't really need this to be exactly 1000 items. It can be 1000 or 1010 or 1030, just make sure to save at least 1000 items. With this argument, the trimming is performed only when we can remove a whole node. This makes it much more efficient, and it is usually what you want.
+The `~` argument between the **MAXLEN** option and the actual count means, I don't really need this to be exactly 1000 items. It can be 1000 or 1010 or 1030, just make sure to save at least 1000 items. With this argument, the trimming is performed only when we can remove a whole node. This makes it much more efficient, and it is usually what you want. You'll note here that the client libraries have various implementations of this; for example the Python client defaults to approximate and has to explicitly be set to a true length.
```
dwdougherty, line 808:
```suggestion
The `~` argument between the **MAXLEN** option and the actual count means, I don't really need this to be exactly 1000 items. It can be 1000 or 1010 or 1030, just make sure to save at least 1000 items. With this argument, the trimming is performed only when we can remove a whole node. This makes it much more efficient, and it is usually what you want. You'll note here that the client libraries have various implementations of this. For example, the Python client defaults to approximate and has to be explicitly set to a true length.
```

dwdougherty (changes requested):
Hi Savannah. Just a few changes.

docs/data-types/streams.md
sav-norem: lol yikes. fixed in the incoming commit

docs/data-types/streams.md
```diff
@@ -805,7 +805,7 @@ However trimming with **MAXLEN** can be expensive: streams are represented by ma
 XADD race:italy MAXLEN ~ 1000 * ... entry fields here ...
 ```
 
-The `~` argument between the **MAXLEN** option and the actual count means, I don't really need this to be exactly 1000 items. It can be 1000 or 1010 or 1030, just make sure to save at least 1000 items. With this argument, the trimming is performed only when we can remove a whole node. This makes it much more efficient, and it is usually what you want. You'll note here that the client libraries have various implementations of this; for example the Python client defaults to approximate and has to explicitly be set to a true length.
+The `~` argument between the **MAXLEN** option and the actual count means, I don't really need this to be exactly 1000 items. It can be 1000 or 1010 or 1030, just make sure to save at least 1000 items. With this argument, the trimming is performed only when we can remove a whole node. This makes it much more efficient, and it is usually what you want. You'll note here that the client libraries have various implementations of this, For example, the Python client defaults to approximate and has to be explicitly set to a true length.
```
dwdougherty, line 808: Oops... the comma before "For example..." should have been a period. Sorry. My bad.

```suggestion
The `~` argument between the **MAXLEN** option and the actual count means, I don't really need this to be exactly 1000 items. It can be 1000 or 1010 or 1030, just make sure to save at least 1000 items. With this argument, the trimming is performed only when we can remove a whole node. This makes it much more efficient, and it is usually what you want. You'll note here that the client libraries have various implementations of this. For example, the Python client defaults to approximate and has to be explicitly set to a true length.
```

dwdougherty:
These changes are go for merge. Will have to wait until @dmaier-redislabs is back from PTO, as I still don't have permissions.

dwdougherty (approved):

sav-norem:
@dwdougherty just wanted to bring this back up since it may have gotten forgotten

dwdougherty:
Thank you for the reminder, @sav-norem. I'll ask @dmaier-redislabs to merge.

dmaier-redislabs (approved):

PR #2526, opened 2023-08-27
Update RESP docs page

rajivharlalka:
- Removed an extra set of closing braces.


zuiderkwast (approved):
Thanks for spotting these and removing them. I can see them here (screenshot):

image

PR #2535, opened 2023-09-08
Put detail back to SCAN page.

### LINKED ISSUE #2529: `SCAN` command page doesn't explain what `COUNT` does or how `MATCH` works sufficiently.
The `SCAN` command page doesn't explain how the `COUNT` modifier works - necessary as this is confusing because it's not exactly a guarantee of how many elements the scan family of commands return... nor how `MATCH` works.  It also does not do a good job of explaining how the cursor works and when to know you have completed a scan of the keyspace.

simonprickett:
Restored important missing detail from the `SCAN` command page.

Closes #2529


soundslocke:
It would be great to get the `scan` documentation page back in order!

zuiderkwast (approved):
Thank you!

People don't look for the Tutorial page when they search for the documented scan guarantees.

Maybe we can delete some of this from the Keyspace Tutorial page where it was moved to before?

commands/scan.md
```diff
@@ -9,19 +9,203 @@ Since these commands allow for incremental iteration, returning only a small num
...
+
+However once the data structures are bigger and are promoted to use real hash tables, the `SCAN` family of commands will resort to the normal behavior. Note that since this special behavior of returning all the elements is true only for small aggregates, it has no effects on the command complexity or latency. However the exact limits to get converted into real hash tables are [user configurable](/topics/memory-optimization), so the maximum number of elements you can see returned in a single call depends on how big an aggregate data type could be and still use the packed representation.
+
+Also note that this behavior is specific of `SSCAN`, `HSCAN` and `ZSCAN`. `SCAN` itself never shows this behavior because the key space is always represented by hash tables.
+
+## Further reading
+
+For more information on `SCAN` please refer to the [The Redis Keyspace](/docs/manual/keyspace) tutorial.
```
zuiderkwast, line 208: So if we delete most of the same text from the keyspace tutorial (which was move there from scan command docs here: https://github.com/redis/redis-doc/commit/882c184514f88f048df3d773e300a39882773e9d#diff-6875ae7c6d2de818737cec8e3e3c0221294ab337dab94fbe7c05ad66e1f789ce) there is not much more information about scan there. Maybe instead something like this:

```suggestion
For more information about managing keys, please refer to the [The Redis Keyspace](/docs/manual/keyspace) tutorial.
```

And to avoid having duplicate information on two places, we can delete Scan basic usage, scan guarantees, etc. from this tutorial and instead link to the SCAN command from there.

PR #2540, opened 2023-09-15
Change 'this' to 'these'.

iportman:
fix a small grammatical mistake.

### DIFF
--- a/commands/incr.md
+++ b/commands/incr.md
@@ -82,7 +82,7 @@ END
 ```
 
 Basically we have a counter for every IP, for every different second.
-But this counters are always incremented setting an expire of 10 seconds so that
+But these counters are always incremented setting an expire of 10 seconds so that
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
You're right. This whole section has quite bad grammar and needs to be rewritten IMO:

> Basically we have a counter for every IP, for every different second. But this counters are always incremented setting an expire of 10 seconds so that they'll be removed by Redis automatically when the current second is a different one.
>
> Note the used of [MULTI](https://redis.io/commands/multi) and [EXEC](https://redis.io/commands/exec) in order to make sure that we'll both increment and set the expire at every API call.

Strange things here:

* "for every different second"
* "But" (better just remove this word)
* "this counters"
* "when the current second is a different one"
* "Note the used of"

I'll merge this PR now and maybe you want to improve the text more in another PR?

PR #2555, opened 2023-09-30
Fix stream name mismatch in code and the text

sJJdGG:
in the code above the suggested diff, the stream name is `race:france` which is coherent with all the other example in the doc. but it's wrongly named `mystream`

### DIFF
--- a/docs/data-types/streams.md
+++ b/docs/data-types/streams.md
@@ -326,7 +326,7 @@ Apart from the fact that `XREAD` can access multiple streams at once, and that w
 > XREAD BLOCK 0 STREAMS race:france $
 ```
 
-Note that in the example above, other than removing **COUNT**, I specified the new **BLOCK** option with a timeout of 0 milliseconds (that means to never timeout). Moreover, instead of passing a normal ID for the stream `race:france` I passed the special ID `$`. This special ID means that `XREAD` should use as last ID the maximum ID already stored in the stream `mystream`, so that we will receive only *new* messages, starting from the time we started listening. This is similar to the `tail -f` Unix command in some way.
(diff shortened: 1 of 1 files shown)


dwdougherty:
Thank you, Sajjad, for the PR.

There are two other occurrences of "mystream" that need to be changed. The first is in the same paragraph you mentioned, line 329. The other is on line 435, where "mystream" should be changed to "race:italy". Can you make these other two changes to this PR?

docs/data-types/streams.md
```diff
@@ -326,7 +326,7 @@ Apart from the fact that `XREAD` can access multiple streams at once, and that w
 > XREAD BLOCK 0 STREAMS race:france $
 ```
 
-Note that in the example above, other than removing **COUNT**, I specified the new **BLOCK** option with a timeout of 0 milliseconds (that means to never timeout). Moreover, instead of passing a normal ID for the stream `mystream` I passed the special ID `$`. This special ID means that `XREAD` should use as last ID the maximum ID already stored in the stream `mystream`, so that we will receive only *new* messages, starting from the time we started listening. This is similar to the `tail -f` Unix command in some way.
+Note that in the example above, other than removing **COUNT**, I specified the new **BLOCK** option with a timeout of 0 milliseconds (that means to never timeout). Moreover, instead of passing a normal ID for the stream `race:france` I passed the special ID `$`. This special ID means that `XREAD` should use as last ID the maximum ID already stored in the stream `mystream`, so that we will receive only *new* messages, starting from the time we started listening. This is similar to the `tail -f` Unix command in some way.
```
dwdougherty, line 329:
```suggestion
Note that in the example above, other than removing **COUNT**, I specified the new **BLOCK** option with a timeout of 0 milliseconds (that means to never timeout). Moreover, instead of passing a normal ID for the stream `race:france` I passed the special ID `$`. This special ID means that `XREAD` should use as last ID the maximum ID already stored in the stream `race:france`, so that we will receive only *new* messages, starting from the time we started listening. This is similar to the `tail -f` Unix command in some way.
```

PR #2557, opened 2023-10-03
fix swapped vars in memory-stats, and minor wording

oranagra:
the percentage is the total (current) out of peak (higher), not the other way around.


zuiderkwast (approved):
Good catch.

Most other metrics are included in "total", thus the confusion, I guess.

The formulation still looks a bit confusing though. Can we make it look more intuitive?

For example, we can explain that `peek.allocated` is the maximum observed total allocated, while `total.allocated` is the current total?

And perhaps something similar for peek.percentage.

oranagra:
not my catch :smile: 

can you suggest changes? or make another PR and close this one...

zuiderkwast:
> can you suggest changes? or make another PR and close this one...

Another time maybe. Your fix is correct so let's merge it.

PR #2569, opened 2023-10-18
Moving HMSET used by examples to HSET

chayim:
[HMSET](https://redis.io/commands/hmset/) is deprecated as of Redis 4.0.0, and [HSET ](https://redis.io/commands/hset/) is the prefered method for this interaction. This PR updates the examples where we were refering to the deprecated command.


zuiderkwast (approved):
Thanks!

PR #2587, opened 2023-10-30
Correction of map representation in RESP2

heyone-top:
In the protocol spec, the description of how maps are represented in RESP2 is incorrect.

"Maps in RESP2 are represented by arrays, in which each element is a key-value tuple. Each tuple is an array with two elements, these being the key and the value."

This is not true. A map in RESP2 is represented as a flat array containing the keys and values, not nested arrays.

### DIFF
--- a/docs/reference/protocol-spec.md
+++ b/docs/reference/protocol-spec.md
@@ -516,8 +516,9 @@ However, low-level programming languages (such as C, for example) will likely re
 
 {{% alert title="Map pattern in RESP2" color="info" %}}
 RESP2 doesn't have a map type.
-Maps in RESP2 are represented by arrays, in which each two element is a key-value tuple.
-Each tuple is an array with two elements, these being the key and the value.
+A map in RESP2 is represented by a flat array containing the keys and the values.
+The first element is a key, followed by the corresponding value, then the next key and so on, like this:
+`key1, value1, key2, value2, ...`.
 {{% /alert %}}
 
 <a name="set-reply"></a>
--- a/docs/reference/protocol-spec.md
+++ b/docs/reference/protocol-spec.md
@@ -516,7 +516,7 @@ However, low-level programming languages (such as C, for example) will likely re
 
 {{% alert title="Map pattern in RESP2" color="info" %}}
 RESP2 doesn't have a map type.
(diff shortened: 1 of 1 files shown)


heyone-top:
hgetall myhash
*8
$1
a
$1
b
$1
c
$1
d
$1
i
$1
1
$1
j
$1
2

docs/reference/protocol-spec.md
```diff
@@ -516,7 +516,7 @@ However, low-level programming languages (such as C, for example) will likely re
 
 {{% alert title="Map pattern in RESP2" color="info" %}}
 RESP2 doesn't have a map type.
-Maps in RESP2 are represented by arrays, in which each element is a key-value tuple.
+Maps in RESP2 are represented by arrays, in which each two element is a key-value tuple.
```
dwdougherty, line 519:
```suggestion
Maps in RESP2 are represented by arrays, in which each pair of elements is a key-value tuple.
```

dwdougherty (dismissed):
Just a minor change and I'll merge.

docs/reference/protocol-spec.md
zuiderkwast: This is not correct.

A map in RESP2 is a flat array with alternating key and value, like `[key1, value1, key2, value2, ...]`. At least that's what HGETALL returns.

The next line "Each tuple is an array with two elements, these being the key and the value." is also wrong.

zuiderkwast: The word "tuple" is confusing in particular. I will suggest another formulation.

docs/reference/protocol-spec.md
```diff
@@ -516,7 +516,7 @@ However, low-level programming languages (such as C, for example) will likely re
...
 Each tuple is an array with two elements, these being the key and the value.
```
zuiderkwast, line 520: Maybe this is more clear?

```suggestion
A map in RESP2 is represented by a flat array containing the keys and the values.
The first element is a key, followed by the corresponding value, then the next key and so on, like this:
`key1, value1, key2, value2, ...`.
```

zuiderkwast (approved):

PR #2592, opened 2023-11-03
fix issues 2591 https://github.com/redis/redis-doc/issues/2591

PansonPanson:


dwdougherty (approved):
LGTM

zuiderkwast (changes requested):
Is this better English? I think the old text is fine and I don't  see the benefit in changing it.

I think the documentation of this command should match DECR, INCR, INCRBY and INCRBYFLOAT. They all use a similar description.

PR #2596, opened 2023-11-07
Update ZRANK WITHSCORE reference in resp2 & resp3_replies.json

Obeyed:
The WITHSCORE option was referenced as WITHSCORES (plural)


dwdougherty (approved):
LGTM

PR #2598, opened 2023-11-09
added radish, the gleam client

massivefermion:
[Radish](https://hex.pm/packages/radish) is the client for the [Gleam programming language](https://gleam.run/). It's newly published so there is still a lot of work to do.
Just one question, what should I do to get recommended? What are the criteria?
Thanks


zuiderkwast (approved):
Thanks. Gleam is very cool. :-) Personally I use Erlang.

PR #2602, opened 2023-11-15
Lindex return code change for index is out of range #12769.

hwware:
This request is to cover return value changes implemented as part of https://github.com/redis/redis/pull/12769.

### DIFF
--- a/resp2_replies.json
+++ b/resp2_replies.json
@@ -641,7 +641,7 @@
   ],
   "LINDEX": [
     "One of the following:",
-    "* [Nil reply](/docs/reference/protocol-spec#bulk-strings): when _index_ is out of range.",
+    "* [Simple error reply](/docs/reference/protocol-spec#simple-errors): when _index_ is out of range.",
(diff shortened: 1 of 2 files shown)


dwdougherty:
Hello Wen Hui. I'm one of the Redis technical writers. In [this file on the main Redis source repo](https://github.com/redis/redis/blob/unstable/src/commands/lindex.json), there's the following definition:

 ```
       "reply_schema": {
            "oneOf": [
                {
                    "type": "null",
                    "description": "Index is out of range"
                },
                {
                    "description": "The requested element",
                    "type": "string"
                }
            ]
        },
```

If that reference is not correct, then it should also be amended.

hwware:
Thank you for reminding the details, I have updated the reply_schema in source repo as well. Please wait until https://github.com/redis/redis/pull/11924 gets merged. so that there will not be any confusions. 

Thank you.

dwdougherty:
Thanks, Wen Hui. Looks like this is going into 8.0, so I'll wait until that's out to merge this PR.

PR #2609, opened 2023-11-29
Remove the duplicated entry in sorted-sets.md

zhzy0077:


dwdougherty (approved):
Thanks for catching this. LGTM.

PR #2610, opened 2023-11-29
Update KEYS and SCAN page with new optimization about pattern with hash tag.

CharlesChen888:
As the optimization done in redis/redis#12754, we may doc this in the command page.

### DIFF
--- a/commands/keys.md
+++ b/commands/keys.md
@@ -30,6 +30,7 @@ Note that when enabling `Redis cluster-mode`, using a pattern with hash tag that
 Because any key matching the pattern will be in the same slot.
 Redis would only iterate over keys in the slot, rather than the whole database.
 For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
+To use pattern with hash tag, see https://redis.io/docs/reference/cluster-spec/#hash-tags for more information.
 
 @examples
 
--- a/commands/keys.md
+++ b/commands/keys.md
@@ -26,6 +26,8 @@ Supported glob-style patterns:
 
 Use `\` to escape special characters if you want to match them verbatim.
 
+Note that using patterns with hash tags which can imply a specific slot can save complexity, as Redis would only iterate over keys in the slot, rather than the whole database. For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
+
 @examples
 
 ```cli
--- a/commands/keys.md
+++ b/commands/keys.md
@@ -26,7 +26,10 @@ Supported glob-style patterns:
 
 Use `\` to escape special characters if you want to match them verbatim.
 
-Note that when enabling `Redis cluster-mode`, using a pattern with hash tag which can imply a specific slot can save complexity, as any key matching the pattern will be in the same slot, Redis would only iterate over keys in the slot, rather than the whole database. For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
+Note that when enabling `Redis cluster-mode`, using a pattern with hash tag that implies a specific slot can save complexity.
+Because any key matching the pattern will be in the same slot.
+Redis would only iterate over keys in the slot, rather than the whole database.
+For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
 
 @examples
 
--- a/commands/keys.md
+++ b/commands/keys.md
@@ -26,7 +26,7 @@ Supported glob-style patterns:
 
 Use `\` to escape special characters if you want to match them verbatim.
 
-Note that using patterns with hash tags which can imply a specific slot can save complexity, as Redis would only iterate over keys in the slot, rather than the whole database. For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
+Note that using a pattern with hash tag which can imply a specific slot can save complexity, as any key matching the pattern will be in the same slot, Redis would only iterate over keys in the slot, rather than the whole database. For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
 
 @examples
 
--- a/commands/keys.md
+++ b/commands/keys.md
@@ -26,7 +26,7 @@ Supported glob-style patterns:
 
 Use `\` to escape special characters if you want to match them verbatim.
 
-Note that using a pattern with hash tag which can imply a specific slot can save complexity, as any key matching the pattern will be in the same slot, Redis would only iterate over keys in the slot, rather than the whole database. For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
+Note that when enabling `Redis cluster-mode`, using a pattern with hash tag which can imply a specific slot can save complexity, as any key matching the pattern will be in the same slot, Redis would only iterate over keys in the slot, rather than the whole database. For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
 
 @examples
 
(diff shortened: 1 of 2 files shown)


commands/keys.md
```diff
@@ -26,6 +26,8 @@ Supported glob-style patterns:
 
 Use `\` to escape special characters if you want to match them verbatim.
 
+Note that using a pattern with hash tag which can imply a specific slot can save complexity, as any key matching the pattern will be in the same slot, Redis would only iterate over keys in the slot, rather than the whole database. For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
```
oranagra, line 29: we're missing a word about `cluster-enabled` (when disabled we don't have this optimization)

oranagra:
let's add the same comment about SCAN in this PR as well.

soloestoy:
Currently, there are several limitations when using hashtags in patterns. For instance, there cannot be wildcards or escape character before the hashtag, and the hashtag within curly braces cannot contain wildcards or support escape character. This is related to the implementation of the `patternHashSlot` function:

```c
/* If it can be inferred that the given glob-style pattern, as implemented in
 * stringmatchlen() in util.c, only can match keys belonging to a single slot,
 * that slot is returned. Otherwise -1 is returned. */
int patternHashSlot(char *pattern, int length) {
    int s = -1; /* index of the first '{' */

    for (int i = 0; i < length; i++) {
        if (pattern[i] == '*' || pattern[i] == '?' || pattern[i] == '[') {
            /* Wildcard or character class found. Keys can be in any slot. */
            return -1;
        } else if (pattern[i] == '\\') {
            /* Escaped character. Computing slot in this case is not
             * implemented. We would need a temp buffer. */
            return -1;
        } else if (s == -1 && pattern[i] == '{') {
            /* Opening brace '{' found. */
            s = i;
        } else if (s >= 0 && pattern[i] == '}' && i == s + 1) {
            /* Empty tag '{}' found. The whole key is hashed. Ignore braces. */
            s = -2;
        } else if (s >= 0 && pattern[i] == '}') {
            /* Non-empty tag '{...}' found. Hash what's between braces. */
            return crc16(pattern + s + 1, i - s - 1) & 0x3FFF;
        }
    }

    /* The pattern matches a single key. Hash the whole pattern. */
    return crc16(pattern, length) & 0x3FFF;
}
```

I believe we need @zuiderkwast 's help to carefully describe the rules for using hashtags in patterns.

commands/keys.md
```diff
@@ -26,6 +26,8 @@ Supported glob-style patterns:
 
 Use `\` to escape special characters if you want to match them verbatim.
 
+Note that when enabling `Redis cluster-mode`, using a pattern with hash tag which can imply a specific slot can save complexity, as any key matching the pattern will be in the same slot, Redis would only iterate over keys in the slot, rather than the whole database. For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
```
zuiderkwast, line 29: Please add a linebreak between each sentence and avoid very long lines, as described in the styling guideline for this repo (see the README). The single line break has no significance in Markdown anyway.

I think we need to mention that this is an *optimization*. (For KEYS and SCAN it is an optimization. For SORT it is more than an optimization.)

As @soloestoy pointed out, there are some specific rules where the optimization applies which he summarized very well. I just split it up into 3 bullets:

* the pattern must contain a hashtag
* there cannot be wildcards or escape characters before the hashtag
* the hashtag within curly braces cannot contain wildcards or escape characters

The example is very good to clarify these rules:

> For example, `{abc}*` can successfully recognize the hashtag and only scan the slot corresponding to `abc`. However, `*{abc}`, `{a*c}`, or `{a\*bc}` cannot recognize the hashtag.

Then I think @dwdougherty can suggest another way of writing this, perhaps using shorter sentences?

commands/scan.md
```diff
@@ -148,6 +148,7 @@ redis 127.0.0.1:6379>
 
 As you can see most of the calls returned zero elements, but the last call where a `COUNT` of 1000 was used in order to force the command to do more scanning for that iteration.
 
+Note that when enabling `Redis cluster-mode`, using a pattern with hash tag which can imply a specific slot can save complexity, as any key matching the pattern will be in the same slot, Redis would only iterate over keys in the slot, rather than the whole database. For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
```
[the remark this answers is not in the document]
dwdougherty: Yeah... I agree. I'll change it to "implies".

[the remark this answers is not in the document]
zuiderkwast: ... or something like "when a pattern can only match keys from a single slot".

I think the whole section can be written in a simpler way somehow. I can try to come up with a suggestion.

commands/keys.md
soloestoy: I think it's better to add the rules for using hashtags in patterns and the list of supported commands to the [cluster hashtag](https://github.com/redis/redis-doc/blob/master/docs/reference/cluster-spec.md#hash-tags) documentation. This way, all the commands that support using hashtags in patterns can reference this link. Can you help to do this?

zuiderkwast: Done in #2646. Please have a look.

soloestoy (approved):

soloestoy:
@oranagra @zuiderkwast please approve

commands/keys.md
```diff
@@ -26,6 +26,12 @@ Supported glob-style patterns:
 
 Use `\` to escape special characters if you want to match them verbatim.
 
+Note that when enabling `Redis cluster-mode`, using a pattern with hash tag that implies a specific slot can save complexity.
+Because any key matching the pattern will be in the same slot.
+Redis would only iterate over keys in the slot, rather than the whole database.
```
zuiderkwast, line 31: There is something strange about the formulation "can save complexity". I'm trying to make it clearer and also comparing to #2646. Is this better?

```suggestion
When using [Redis Cluster](/docs/management/scaling/), the search is optimized for patterns that imply a single slot.
If a pattern can only match keys of one slot,
Redis only iterates over keys in that slot, rather than the whole database,
when searching for keys matching the pattern.
```

commands/scan.md
```diff
@@ -148,6 +148,11 @@ redis 127.0.0.1:6379>
 
 As you can see most of the calls returned zero elements, but the last call where a `COUNT` of 1000 was used in order to force the command to do more scanning for that iteration.
 
+Note that when enabling `Redis cluster-mode`, using a pattern with hash tag that implies a specific slot can save complexity.
+Because any key matching the pattern will be in the same slot.
+Redis would only iterate over keys in the slot, rather than the whole database.
```
zuiderkwast, line 153: Same as above.

```suggestion
When using [Redis Cluster](/docs/management/scaling/), the search is optimized for patterns that imply a single slot.
If a pattern can only match keys of one slot,
Redis only iterates over keys in that slot, rather than the whole database,
when searching for keys matching the pattern.
```

commands/keys.md
```diff
@@ -26,6 +26,12 @@ Supported glob-style patterns:
...
+For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
+To use pattern with hash tag, see https://redis.io/docs/reference/cluster-spec/#hash-tags for more information.
```
zuiderkwast, line 33:
```suggestion
To use pattern with hash tag, see [Hash tags](/docs/reference/cluster-spec/#hash-tags) in the Cluster specification for more information.
```

commands/scan.md
```diff
@@ -148,6 +148,11 @@ redis 127.0.0.1:6379>
...
+For example, with the pattern `{a}h*llo`, Redis would only try to match it with the keys in slot 15495, which hash tag `{a}` implies.
+To use pattern with hash tag, see https://redis.io/docs/reference/cluster-spec/#hash-tags for more information.
```
zuiderkwast, line 155:
```suggestion
To use pattern with hash tag, see [Hash tags](/docs/reference/cluster-spec/#hash-tags) in the Cluster specification for more information.
```

zuiderkwast (commented):
Minor suggestions to formulation.

zuiderkwast (approved):

oranagra (approved):

PR #2613, opened 2023-11-30
Revise the response description of `OBJECT FREQ` command

moznion:
`OBJECT FREQ` would return the nil/null response if the target key doesn't exist.


dwdougherty (approved):
Behavior confirmed. LGTM. Thank you very much!

PR #2614, opened 2023-11-30
Remove nil/null reply description from ZADD command

moznion:
This command doesn't return nil/null reply even if the NX/XX/LT/GT options are conflict; in that case, it returns the error reply like the following:

```
zadd z XX NX 1 foo
-ERR XX and NX options at the same time are not compatible
zadd z NX GT 1 foo
-ERR GT, LT, and/or NX options at the same time are not compatible
zadd z LT GT 1 foo
-ERR GT, LT, and/or NX options at the same time are not compatible
```

### DIFF
--- a/resp2_replies.json
+++ b/resp2_replies.json
@@ -1204,7 +1204,6 @@
   ],
   "ZADD": [
     "Any of the following:",
-    "* [Nil reply](/docs/reference/protocol-spec#bulk-strings): if the operation was aborted because of a conflict with one of the _XX/NX/LT/GT_ options.",
     "* [Integer reply](/docs/reference/protocol-spec#integers): the number of new members when the _CH_ option is not used.",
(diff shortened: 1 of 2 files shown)


zuiderkwast (commented):
Interesting. Can you find the PR where this was changed in redis?

zuiderkwast:
You are right that an error is returned when the options conflict, but ZADD can still return null in some cases. See the test case "ZADD INCR LT/GT replies with nill if score not updated" in [tests/unit/type/zset.tcl](https://github.com/redis/redis/blob/unstable/tests/unit/type/zset.tcl#L253).

```
127.0.0.1:6379> zadd ztmp 28 x
(integer) 1
127.0.0.1:6379> zadd ztmp lt incr 1 x
(nil)
```

I think we should change the description to something like "... if the score was not updated when INCR was combined with LT/GT options" similar to what is used in the test case.

PR #2617, opened 2023-12-05
SET options EX, PX, EXAT, PXAT take positive integer argument

LiorKogan:
Clarify that argument is a positive integer, unlike EXPIRE which accepts zero and negative numbers too.

### DIFF
--- a/commands/set.md
+++ b/commands/set.md
@@ -6,10 +6,10 @@ Any previous time to live associated with the key is discarded on successful `SE
 
 The `SET` command supports a set of options that modify its behavior:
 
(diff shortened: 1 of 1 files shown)


oranagra (approved):
context: unlike EXPIRE, which can take negative numbers and implicitly delete the key.
for the SET command, negative TTLs are disallowed.
see this PR for some context: https://github.com/redis/redis/pull/8287

LiorKogan:
(not just negative values, but also 0)

zuiderkwast (approved):

PR #2630, opened 2023-12-22
fixed #886

### LINKED ISSUE #886: key expiry and ttl during the evaluation of lua scripts
As far as I know, these behaviors are not documented:

1. If a key with expiry (ttl) exists at the start of eval, it will not get expired during the evaluation of a script.
2. During the evaluation of a lua script, ttl of a key can go down to zero. After that, it remains zero until the script terminates.
3. To have logic based on the value of `ttl` can violate the concept of [scripts as pure functions](https://redis.io/commands/eval#scripts-as-pure-functions).

mich-elle-luna:
adds a couple of tips to EVAL doc about TTL and fixes a typo #886

### DIFF
--- a/commands/eval.md
+++ b/commands/eval.md
@@ -8,9 +8,11 @@ These names of input keys are available to the script as the [_KEYS_ global runt
 Any additional input arguments **should not** represent names of keys.
 
 **Important:**
-to ensure the correct execution of scripts, both in standalone and clustered deployments, all names of keys that a script accesses must be explicitly provided as input key arguments.
+To ensure the correct execution of scripts, both in standalone and clustered deployments, all names of keys that a script accesses must be explicitly provided as input key arguments.
 The script **should only** access keys whose names are given as input arguments.
 Scripts **should never** access keys with programmatically-generated names or based on the contents of data structures stored in the database.
+If a key with expiry (TTL) exists at the start of eval, it will not get expired during the evaluation of a script.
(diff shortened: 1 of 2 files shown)


commands/eval.md
```diff
@@ -8,9 +8,11 @@ These names of input keys are available to the script as the [_KEYS_ global runt
 Any additional input arguments **should not** represent names of keys.
 
 **Important:**
-to ensure the correct execution of scripts, both in standalone and clustered deployments, all names of keys that a script accesses must be explicitly provided as input key arguments.
+To ensure the correct execution of scripts, both in standalone and clustered deployments, all names of keys that a script accesses must be explicitly provided as input key arguments.
 The script **should only** access keys whose names are given as input arguments.
 Scripts **should never** access keys with programmatically-generated names or based on the contents of data structures stored in the database.
+If a key with expiry (TTL) exists at the start of eval, it will not get expired during the evaluation of a script.
+During the evaluation of a Lua script, TTL of a key can go down to zero. After that, it remains zero until the script terminates.
```
zuiderkwast, line 15: Are you sure? Can you give an example?

I thought that we freeze time during evaluation, so TTL will return the same value every time throughout the evaluation.

zuiderkwast: See redis/redis#10300

CC @enjoy-binbin

enjoy-binbin: yean, we will freeze the time during the evaluation, the [P]TTL should return the same value in this case in Redis 7.2

PR #2636, opened 2023-12-29
Add "Production usage" section to Jedis quick start

uglide:
cc @dwdougherty @sazzad16 @chayim

### DIFF
--- a/docs/connect/clients/java.md
+++ b/docs/connect/clients/java.md
@@ -190,17 +190,16 @@ public class Main {
 ### Production usage
 
 ### Configuring Connection pool
-As was covered in the previous section, it's recommended to use `JedisPool` or `JedisPooled`. 
-`JedisPooled` is available since Jedis 4.0.0 and provides similar capabilities as `JedisPool` with a simpler API.
-A connection pool holds a specified amount of connections. 
-Creates more connections when needed and terminates them when they are unwanted.
+As mentioned in the previous section, use `JedisPool` or `JedisPooled` to create a connection pool.
+`JedisPooled`, added in Jedis version 4.0.0, provides capabilities similar to `JedisPool` but with a more straightforward API.
+A connection pool holds a specified number of connections, creates more connections when necessary, and terminates them when they are no longer needed.
 
-Here is a simplified connection lifecycle in the pool:
+Here is a simplified connection lifecycle in a pool:
 
 1. A connection is requested from the pool.
-2. A connection is served. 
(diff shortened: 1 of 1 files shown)


docs/connect/clients/java.md
```diff
@@ -187,104 +187,126 @@ public class Main {
 }
 ```
 
-### Example: Indexing and querying JSON documents
+### Production usage
 
-Make sure that you have Redis Stack and `Jedis` installed. 
+### Configuring Connection pool
+As was covered in the previous section, it's recommended to use `JedisPool` or `JedisPooled`. 
+`JedisPooled` is available since Jedis 4.0.0 and is a wrapper around `JedisPool` that provides a simpler API.
```
sazzad16, line 194: `wrapper` is wrong term here.
It allows similar (exactly same) connection (resource) maintenance as JedisPool.
But, yes, it provides simpler API.

docs/connect/clients/java.md
```diff
@@ -187,104 +187,126 @@ public class Main {
...
+// JedisPooled does all hard work on fetching and releasing connection to the pool
+// to prevent connection starvation
+JedisPooled jedis = new JedisPooled(poolConfig, "localhost", 6379);
 ```
 
-Create an index. In this example, all JSON documents with the key prefix `user:` are indexed. For more information, see [Query syntax](/docs/interact/search-and-query/query/).
 
+### Timeout
+
+To set a timeout for a connection, use `JedisPooled` or `JedisPool` constructor with `timeout` parameter or `JedisClientConfig` with `socketTimeout` parameter:
```
sazzad16, line 247: There are several types of timeout.
- connectionTimeout
- soTimeout / socketTimeout
- infiniteSoTimeout / blockingSocketTimeout (socketTimeout but for blocking commands)

Lone timeout parameter in constructors acts value for both connectionTimeout and socketTimeout.

docs/connect/clients/java.md
uglide: Fixed. Thanks!

docs/connect/clients/java.md
uglide: Fixed

dwdougherty (approved):
LGTM

sazzad16 (approved):
It suggests Jedis `4.3.1`. Should we do an updated one?
Otherwise, LGTM.

sazzad16:
@uglide BTW, spellcheck fails.

PR #2639, opened 2024-01-05
Add UI Bakery

lugovsky:
UI Bakery allows the building of custom internal tools, admin panels, and dashboards on top of Redis and other databases. It also has an example explorer application built on top of Redis, which can be used as a starting point for a custom Redis UI app: https://uibakery.io/templates/redis-admin.

Therefore, it could be a good addition to the list of GUI tools.

### DIFF
--- a/tools/gui/github.com/uibakery/self-hosted.json
+++ b/tools/gui/github.com/uibakery/self-hosted.json
@@ -1,0 +1,8 @@
+{
+    "name": "UI Bakery",
+    "description": "A drag and drop UI builder for web apps on top of Redis and other databases.",
+    "homepage": "https://uibakery.io"
+    "twitter": [
+        "uibakery"
+    ]
+}


zuiderkwast (changes requested):
UI Bakery looks like a CMS capable of using a lot of databases. I don't think we should list all CMS-es and frameworks that can use Redis. The Tools page are for "Tools for managing and deploying Redis". So I would say no, but maybe someone else here has a different opinion.

Regarding the redis-admin thing, https://uibakery.io/templates/redis-admin, it looks like a better candidate, but it's part of a hosted service without source code available? I suppose it can only be used to manage the Redis instance that comes with the UI bakery subscription? So far, all tools listed on redis.io are open source tools (or at least source available) that can be used for managing any redis instance.

lugovsky:
Hello @zuiderkwast,

Thank you for your comment. However, I still believe UI Bakery can be a valuable addition to the list of Redis tools.

UI Bakery is not specifically a CMS. It's a low-code development platform that enables the creation of custom web applications through a drag-and-drop interface, utilizing data from various databases, including Redis. This approach eliminates the need to learn UI frameworks or set up deployment pipelines, as the platform manages these aspects. It enables faster web app development compared to the traditional coding approach. UI Bakery does not include Redis as part of its service or installation. Instead, developers can connect to their own Redis instances.

UI Bakery offers two usage options: as a SaaS or as a self-hosted installation via Docker. Although it is not open source, both options provide a free plan.

The template you mentioned is an example application we built to demonstrate the platform's capabilities using sample Redis data. This template serves as a starting point for building a custom Redis Admin web application with UI Bakery. Users can connect their own Redis database and modify the UI and business logic according to their needs.

zuiderkwast:
So it's a tool that can create almost any kind of web application. The Redis website lists "Tools for managing and deploying Redis". I don't think it fits under this category.

Does anyone have a different optinion? @itamarhaber @enjoy-binbin

lugovsky:
> So it's a tool that can create almost any kind of web application. The Redis website lists "Tools for managing and deploying Redis". I don't think it fits under this category.

It's a tool designed to build any kind of web application on top of databases, including Redis. For instance, one can build a custom UI to manage their Redis database records, as shown on this link: [https://uibakery.io/templates/redis-admin](https://uibakery.io/templates/redis-admin). More complex use cases might include UIs with custom business logic, allowing non-technical users to manage the Redis DB, which is hardly achievable with tools currently listed in the docs.

Therefore, I don't see why it can't fall under the category of "Tools for Managing Redis." It seems incorrect to disregard this tool simply because it's not Redis-specific and supports other databases.

enjoy-binbin:
> So it's a tool that can create almost any kind of web application. The Redis website lists "Tools for managing and deploying Redis". I don't think it fits under this category.
Does anyone have a different optinion? @itamarhaber @enjoy-binbin

i don't have much knowledge in this area to respond...
let’s wait for @itamarhaber 's response

lugovsky:
Hello @itamarhaber. Would you be able to review this?

PR #2642, opened 2024-01-10
Update FUNCTION LIST example to mention output has flags

enjoy-binbin:


zuiderkwast (approved):
Thanks.

Were the flags added to Redis after the docs were written?

enjoy-binbin:
yes, they were already added. see https://redis.io/docs/interact/programmability/lua-api/#script_flags

PR #2643, opened 2024-01-11
Add notes of `:0@0` addresses in `CLUSTER NODES`

bjosv:


commands/cluster-nodes.md
```diff
@@ -38,7 +38,7 @@ Each line is composed of the following fields:
 The meaning of each field is the following:
 
 1. `id`: The node ID, a 40-character globally unique string generated when a node is created and never changed again (unless `CLUSTER RESET HARD` is used).
-2. `ip:port@cport`: The node address that clients should contact to run queries.
+2. `ip:port@cport`: The node address that clients should contact to run queries, along with the used cluster bus port. `:0@0` can be expected when the address is no longer known for this node ID, hence flagged with `noaddr`.
```
zuiderkwast, line 41:
```suggestion
2. `ip:port@cport`: The node address that clients should contact to run queries, along with the used cluster bus port.
   `:0@0` can be expected when the address is no longer known for this node ID, hence flagged with `noaddr`.
```

PR #2676, opened 2024-03-03
Fix the rendering of the RESP documentation for the INFO command

qwertzguy:
On https://redis.io/commands/info/ at the bottom, the RESP documentation renders like this:
image
Notice the
>in the form of : where

and
>separated map like =.

and
>terminated by .


zuiderkwast (approved):
Thanks!

PR #2683, opened 2024-03-11
Correct RESP3 return type Set for SDIFF, SINTER, SUNION, SMEMBERS, SPOP

zuiderkwast:
The return type of these commands is listed as array, but in RESP3 they return a set reply.

I suspect the reason for this originates from the fact that the reply schema is a JSON schema, and JSON doesn't have sets, so the schemas don't distinguish between arrays and sets.

@guybe7

### DIFF
--- a/resp3_replies.json
+++ b/resp3_replies.json
@@ -956,7 +956,7 @@
     "[Bulk string reply](/docs/reference/protocol-spec#bulk-strings): the SHA1 digest of the script added into the script cache."
   ],
   "SDIFF": [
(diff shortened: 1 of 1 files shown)


Yury-Fridlyand:
Voting for merging this:
```
$ redis-cli -3
127.0.0.1:6379> sinter s1
1~ "a"
127.0.0.1:6379> sinter s1
1~ "a"
127.0.0.1:6379> sinter s1 s2
(empty set)
127.0.0.1:6379> smembers s1
1~ "a"
127.0.0.1:6379>
```

PR #2687, opened 2024-03-13
Add description for XREAD new '+' special ID for last stream entry (#…

ronen-kalish:
Adding description of the '+' new special stream ID for the XREAD command.

### DIFF
--- a/commands/xread.md
+++ b/commands/xread.md
@@ -183,6 +183,26 @@ Once we get some replies, the next call will be something like:
 
 And so forth.
 
+## The special `+` ID
+
+It is sometimes desirable to read the last entry in a stream. For a single
+stream this can be easily achieved using the `XREVRANGE` command, like so:
+
+```
+> XREVRANGE stream + - COUNT 1
+```
+But if there are many streams involved, this approach is quickly becoming
+too slow, having to issue a command per stream.
+Instead, starting from redis 7.6, the `+` sign can be used as a special ID,
+requesting the last available entry in a stream. For example:
+
+```
+> XREAD STREAM streamA streamB streamC streamD + + + +
+```
+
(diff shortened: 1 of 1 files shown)


commands/xread.md
```diff
@@ -183,6 +183,26 @@ Once we get some replies, the next call will be something like:
...
+```
+> XREVRANGE stream + - COUNT 1
+```
+But if there are many streams involved, this approach is quickly becoming
+too slow, having to issue a command per stream.
+Instead, starting from redis 7.6, the `+` sign can be used as a special ID,
+requesting the last available entry in a stream. For example:
+
+```
+> XREAD STREAM streamA streamB streamC streamD + + + +
```
oranagra, line 200: do we want to mention that the pattern is usually that you start iterating on a stream with `+` and then switch to `$` on the next call?

ronen-kalish: I don't know if this is the common use-case, maybe sometimes users just need to peep at the last element and don't continue from there. But I can add a mention if you think it's required.

oranagra: @majklik, @guybe7 please advise.

majklik: One nice use case is the mentioned peek fro last value in multiple streams. In this case is perfect one time shoot with: XREAD STREAM streamA streamB streamC streamD + + + +
But for reliable reading of the content of streams, when I need to start from the last available value (which is my case), after the initial "XREAD STREAM streamA streamB streamC streamD + + + +" it is convenient to continue with the obtained IDs from the streams in the style: "XREAD STREAM streamA streamB streamC streamD idA idB idC idD". If the initial reading shows that one of the streams is empty, I will use 0 as the id for subsequent readings.

PR #2700, opened 2024-03-26
DOC-3547: gather Java clients under Java 'folder'

dwdougherty:
Created new Java "folder" under Connect > Clients, and moved the Jedis and Lettuce guides under it.

@mich-elle-luna Igor would like this merged today at some point, so please review and approve.

java-reorg


mich-elle-luna (approved):
thank you!

PR #2720, opened 2024-12-07
Update PHP Redlock impl. doc

mvorisek:
Fix the PHP Redlock impl. link and move it as the 1st mentioned PHP impl. as it is the most popular and the only library which supports both https://github.com/phpredis/phpredis and https://github.com/predis/predis clients.


mvorisek:
closing as this repo is obsolete

